PR #40, opened 2016-01-09
Change `let module` to `mod` keyword. Change `mod` to `%`.

### LINKED ISSUE #26: Change `let module` to `mod` keyword.
Change `let module` to `mod` keyword.

Rust uses the word `mod` to declare a module. This really doesn't matter at all to me, but if others think it's an improvement, it's a trivial change to make.

Here's an example: https://doc.rust-lang.org/reference.html#modules

yunxing:
Fix #26

### DIFF
--- a/formatTest/formatOutput.re
+++ b/formatTest/formatOutput.re
@@ -830,7 +830,7 @@ type intListTranformer = list int => list int;
 
 type x = list (int, string);
 
-let module HoldsAType = {
+mod HoldsAType = {
   type hasPrime 'a 'b 'c =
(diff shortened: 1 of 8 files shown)


jordwalke:
Now that we have proper identifier swapping systems in place, I'd like to see us bring back `%` as the infix mod operator. We could always just have a prelude that defines the function `%`, but then it is not as seamless of a conversion back/forth between ocaml/reason and would require runtime support. We can swap the operators for `(mod)` and `%` perhaps.

PR #243, opened 2016-03-30
Add support for line comments of the form `//`

dxu:
This adds support for line comments of the form `// comment`.
1. It updates the lexer to recognize `//`, by maintaining a `single_line_comment` `bool` indicating whether we're currently parsing a single line comment. this differentiation prevents cases like `// */ let x = 10` from executing, so that the `*/` does not get parsed as an end of comment. It also maintains a separate `line_comment_start_loc` to keep track of the line comment locations to separate the logic for checking for unmatched comments, so that newlines and nested `*/` comment endings can be differentiated. It also updates the pretty printer to handle these cases accordingly.
2. There is an edge case with regards to interleaved comments having to maintain their "multiline" status.

For example, in the `wrappingTest.re` cases in facebook/Reason,

```
/*A*/
let named
    /* a::a */
    a::a
    /* b::b */
    b::b =>
  /* a + b */
  a + b;

/*B*/
let namedAlias
    /* a::aa */
    a::aa
    /* b::bb */
    b::bb =>
  /* aa + bb */
  aa + bb;
```

must be compiled into 

```
//A
let named /* a::a */ a::a /* b::b */ b::b => /* a + b */ a + b;

//B
let namedAlias
    /* a::aa */
    a::aa
    /* b::bb */
    b::bb => /* aa + bb */ aa + bb;
```

as opposed to 

```
//A
let named // a::a  a::a // b::b  b::b => // a + b  a + b;

//B
let namedAlias
    // a::aa 
    a::aa
    // b::bb 
    b::bb => // aa + bb  aa + bb;
```

to fix this, I check if the listConfig every defines its `break` attribute to be `Never` or `IfNeed`, it will force a multiline comment. 
1. As per the discussion on facebook/ReasonSyntax#41, there is also the edge case of nested inline comments within a line comment. To handle this, there is a new error within the lexer, `Unterminated_nested_comments` that get thrown when a nested comment is detected to not have been matched.

There are a few open questions that aren't answered by this pull request, both of which can be found in greater detail in the discussion on facebook/reasonsyntax#41:
1. How should we handle line breaking? Should it be handled through easy_format (which is causing the current edge cases)? 
2. Converting from ML <-> reason, you could potentially run into problems with nested comments - for example, if you had strings of either `/*` or `(*` converting from one to the other, it could potentially cause issues.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ formatTest/customMLFiles/*.ml
 formatTest/customMLFormatOutput.re
 .jenga
 src/reason_parser.messages.bak
+*~
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 # Portions Copyright (c) 2015-present, Facebook, Inc. All rights reserved.
 
-SHELL=/bin/bash -o pipefail
+SHELL=bash -o pipefail
 
 default: build test
 
--- a/Makefile
+++ b/Makefile
@@ -33,12 +33,12 @@ clean:
 
 # Compile error messages into ml file, checks if the error messages are complete and not redundent
 compile_error: update_error
-	menhir --strict --unused-tokens src/reason_parser.mly --compile-errors src/reason_parser.messages > src/reason_parser_message.ml
+	menhir --explain --strict --unused-tokens src/reason_parser.mly --compile-errors src/reason_parser.messages > src/reason_parser_message.ml
 
 # Update error messages based on new grammar
 update_error:
 	@ cp -f src/reason_parser.messages src/reason_parser.messages.bak
-	@ if ! menhir --strict --unused-tokens src/reason_parser.mly --update-errors src/reason_parser.messages.bak | sed -e 's/[[:space:]]*$$//g' > src/reason_parser.messages ; then \
+	@ if ! menhir --explain --strict --unused-tokens src/reason_parser.mly --update-errors src/reason_parser.messages.bak | sed -e 's/[[:space:]]*$$//g' > src/reason_parser.messages ; then \
 		cp src/reason_parser.messages.bak src/reason_parser.messages ; \
 		exit 1 ; \
 	fi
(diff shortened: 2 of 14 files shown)


bslatkin:
Did this ever land? [This page](http://facebook.github.io/reason/mlCompared.html#reason-and-ocaml-comments) says that Reason supports line comments in the form `// blah blah`, but in practice I can't get them to work. `/* blah blah */` comments work fine.

dxu:
This unfortunately was never landed, and I think not fully reviewed. There was another issue regarding the comment interleaving logic introduced by another commit that is referenced above, which was causing issues at the time (#302). Due to some other responsibilities that cropped up in my personal life, I haven't gotten a chance to look at or address either in a while, so I'm not whether or not it's been fixed.

Also, this hasn't been rebased (as is apparent from the garbled history), but this PR also became bugged when the repo changed its name from facebook/Reason to facebook/reason (it caused my previous repo, dxu/Reason, to no longer be recognized as a fork), so there's a slightly more up-to-date fork at https://github.com/dxu/reason/tree/comments. It still isn't rebased with the most recent commits, but 

Sorry for the delay! @yunxing, let me know if you wanted to take this over, otherwise I might not be able to get to this until a few weeks from now. 

This pull request should also be closed and replaced with one from a repository that is detected as a fork.

yunxing:
@dxu Thanks for the update. I will take over this diff as soon as I get a cycle next Thursday.

jaredly:
bump! This would be super nice

yunxing:
We have the necessary support in our comment interleaving logic now. Should be pretty easy, gimme some more days.

tekknolagi:
@yunxing any update on this?

chenglou:
Nooooooooo

tekknolagi:
Does this work now?

chenglou:
Nope

jordwalke:
cc @let-def :D

jordwalke:
:D 
https://github.com/facebook/reason/pull/1817

PR #366, opened 2016-04-21
[Syntax] Convert poly variant from using ` to #

chenglou:
This is a stepping stone to turning interpolated string from {|bla|} to
`bla`. Which is more familiar to JS devs and frees the weird syntax
error of switch bla {|lol => 1}

Test plan: `make test`

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/mlVariants.re
+++ b/formatTest/typeCheckedTests/expected_output/mlVariants.re
@@ -1,13 +1,13 @@
 /* Copyright (c) 2015-present, Facebook, Inc. All rights reserved. */
 type polyVariantsInMl = [
-  | `IntTuple of (int, int)
-  | `StillAnIntTuple of (int, int)
+  | #IntTuple of (int, int)
+  | #StillAnIntTuple of (int, int)
 ];
 
-let intTuple = `IntTuple (1, 2);
+let intTuple = #IntTuple (1, 2);
 
-let stillAnIntTuple = `StillAnIntTuple (4, 5);
+let stillAnIntTuple = #StillAnIntTuple (4, 5);
 
 let sumThem =
-  fun | `IntTuple (x, y) => x + y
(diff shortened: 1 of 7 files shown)


SanderSpies:
Has there been any consensus yet on how the `#` will be used? @chenglou asked me to look into this, but I want to make sure it's still on the table to be implemented.

Also if we do want to have this, where can I best start with writing tests?

hhugo:
How does the following convert to Reason ?
```
type t = [ `a | `b ]
type u = [ t | `c ]
let f = function
  | #t -> 1
  | `c -> 2
```

SanderSpies:
On my machine it transforms to:
```
type t = [ | `a | `b];
type u = [ t | `c];

let f =
  fun
  | #t => 1
  | `c => 2;
```

I don't think we have tests for this yet - at least a quick search for `#` within formatTests didn't show this scenario.

hhugo:
@SanderSpies, this doesn't reflect this very PR (which is marked as merged).

SanderSpies:
@hhugo It was reverted.

hhugo:
got it.

braibant:
Regarding `switch bla {|lol => 1}`, how do you plan to support the identifiers part of quoted/interpolated strings in reason? http://caml.inria.fr/pub/docs/manual-ocaml/extn.html#sec249

This makes it possible to have an arbitrary number of string litteral delimiters (which is nice when nesting those interpolated strings), and there was plans to use those identifiers in PPX rewriters IIRC.

chenglou:
JS has the syntax bla\`stuff`. Maybe we can use that

braibant:
Can you give an example of how this works?

braibant:
Note that `int_of_string"1"` works in OCaml, and I do not see how this would go with the backtick proposal.

hcarty:
How would
<code>
{foo|foo`|foo}
</code>
translate, as a concrete example?

PR #383, opened 2016-04-25
Add enough OPAM metadata to build mechanically

avsm:
This introduces an OPAM repository under `repo/` which can be used
by automated tooling to get all the pins required.

To see it in action, you can install [dopam](https://github.com/avsm/dopam)
on a Mac without any development tools.  Reason can be built by running:

```
dopam init reason
make
```

A number of other distributions and compiler versions can also be
experimented with by passing options to `dopam init` or doing:

```
dopam bulk-compilers
dopam bulk-distros
```

`dopam` works by initialising a Docker container per combination
of OS/distro, and running OPAM inside the container.  The source
tree is mapped in from the Mac host, so normal editors can be used.

If this PR is useful, I will also update the `README` and documentation to provide the Docker option as a quick install mechanism.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ src_test/_tags
 .DS_Store
 formatTest/customMLFiles/*.ml
 formatTest/customMLFormatOutput.re
+.dopam/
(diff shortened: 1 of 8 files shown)


avsm:
This is happily comprehensively unnecessary now, so closing :-)

jordwalke:
Yes, thank you to everyone who has helped us port this to a modern setup consisting of opam and even dune builds!

PR #472, opened 2016-05-17
fix a broken link in projects page

yutongp:
I linked it to github readme, let me know if you want it link to [index.html#reason-install](http://facebook.github.io/reason/index.html#reason-install) instead


yunxing:
Good catch =)

PR #476, opened 2016-05-18
Add eval `opam config env`

SanderSpies:
I followed the installation instructions, but the rebuild command didn't work - but after running eval `opam config env` everything worked.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@ opam update
 opam pin add -y merlin https://github.com/the-lambda-church/merlin.git#reason-0.0.1
 opam pin add -y merlin_extend https://github.com/let-def/merlin-extend.git#reason-0.0.1
 opam pin add -y reason https://github.com/facebook/reason.git#0.0.6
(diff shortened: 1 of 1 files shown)


yunxing:
Good catch. One does need to `eval opam config env` to make new opam configs take effects in the current shell. We should definitely put this instruction in the doc.

chenglou:
The instruction to add `opam config env` is a few line above (in case you didn't see it)
(Hey again!)

yunxing:
@chenglou I think we need to be explicit that people have to type it in the current shell.

yunxing:
@SanderSpies Could you move the line right below `opam switch 4.02.3` ? If someone was using another switch before he switched to `4.02.3`, the package pinned below will be pinned to the original switch instead of `4.02.3` unless he types `eval opam config env`.

SanderSpies:
@yunxing Will do.

@chenglou (Hey!) Right, my mistake indeed.

PR #481, opened 2016-05-18
Ocp indent all the .ml files

SanderSpies:
Although not perfect, it ensures that the indentation is the same everywhere.

(What triggered this was:  https://github.com/facebook/reason/blob/master/src/reason_pprint_ast.ml#L3756)

Perhaps an idea to put this into the "Reason developer" workflow?


SanderSpies:
(btw. I fully understand if this PR is too big)

jordwalke:
We have some people with important outstanding PRs (such as 4.03 support). How would this effect their ability to git rebase?

yunxing:
This is definitely helpful. 

Maybe it is easier to do it incrementally ?

kayceesrk:
@jordwalke was going to redo the 4.03 support PR since we now use menhir so this particular commit will not affect it. But the PR is big, and might be worth doing incrementally.

jordwalke:
Agree, @kayceesrk

SanderSpies:
Alright, I'll be picking this up incrementally :+1:.

PR #484, opened 2016-05-18
Fix records examples

borellvi:
The examples here aren't actually using Records.

### DIFF
--- a/index.html
+++ b/index.html
@@ -324,7 +324,7 @@ values they describe.
 
 ```reason
 type person = {age: int, name: string};
-let me = {age: 30, name: "Jordan"};
+let me: person = {age: 30, name: "Jordan"};
 print_string me.name;
 ```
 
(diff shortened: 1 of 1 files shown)


facebook-github-bot-4:
Thank you for signing our Contributor License Agreement. We can now accept your code for this (and any) Facebook open source project. Thanks!

kayceesrk:
Thanks. But why do you say?

> The examples here aren't actually using Records.

For example,

```
$ rtop
Reason # type person = {age: int, name: string};
type person = { age : int, name : bytes, }                                                                                                                                                                         Reason # let me = {age: 30, name: "Jordan"};
let me : person = {age : 30, name : "Jordan"}                                                                                                                                                                      Reason # 
```

You don't need explicit type annotations in Reason. Type inference figures out the types for you.

borellvi:
I see, maybe it would be good to specify that in Reason section then. Thanks for that, I will close the PR.

PR #490, opened 2016-05-19
Quote opam URLs to avoid "zsh: no matches found" errors.

benjamn:
Not sure if this error is `zsh`-only, but the quoted URLs should work in all shells.


yunxing:
Looks good to me, thanks!

jordwalke:
Thanks @benjamn !

Issue #488, opened 2016-05-18
Syntax for pattern matching on lists


jayelm:
Right now, both of these work:

```Reason
let head = fun lst => switch lst {
  | [] => failwith "Empty list"
  | [hd, ...tl] => hd
};

let head = fun lst => switch lst {
  | [] => failwith "Empty list"
  | hd::tl => hd
};
```

Is this by design? Since `item1 :: item2 :: rest` has been replaced by `[item1, item2, ...rest]`, it seems like sticking to the latter syntax while pattern matching is the more consistent option.

jordwalke:
Good call. We must have merely not deleted this parsing rule yet. Feel free to send a PR to remove it.

modlfo:
Hi!
Among all the syntax changes, this is the only one I did not like. What seems odd for me is that the "..." is used as well in records "{...a,  }". My suggestion is to use the "..." only in one of the contexts. 

It would be interesting to know the motivations behind this syntax change.

yunxing:
@modlfo Good question! 

When I first saw the syntax, I feel it is pretty intuitive. I think `...` as "expand" operator.

In the case of `[hd, ...tl]`, I interpret it as "expand list `tl`, add `hd` in front of it"

In the case of `{...a, }`, I interpret it as "expand record `a`, modify certain fields of it"

jordwalke:
Hi @modlfo: It's good to hear you found the rest of the changes helpful/preferable. This is the syntax used by JavaScript for Array and object(record) "spreading". That syntactic pattern `...` represents "that other thing, but with these things added".
The reasoning was that it doesn't matter too much what they look like as long as they have no syntactic pitfalls and were somewhat easy to learn.
I don't really love them, personally (I just find three dots `...` a little annoying to type), but I find like that more people can instantly recognize them in my code even without having to learn a new language.

I like this better fwiw, do you agree?

```
let r = {myRecord | nextField: "hi"};
let a = [head | tl ];
```

modlfo:
Hi @yunxing that's exactly the case (`[...tl, hd]`) I was thinking that it can be confusing with `...`

Since I'm more familiar with OCaml and F# I have a strong preference for the cons operator `::` to create and pattern-match lists.  `h :: t` is more compact than `[ h, ... t ]`.

[...]

PR #515, opened 2016-05-22
Give error when using :: notation as a pattern.

SanderSpies:
With this change the following error: 

```
File "Hello.re", line 8, characters 6-8:
:: is not supported in Reason, please use [hd, ...tl] instead
```

will be given when

```
| hd :: tl -> 
```

is used.

### DIFF
--- a/formatTest/unit_tests/input/variants.re
+++ b/formatTest/unit_tests/input/variants.re
@@ -276,7 +276,7 @@ let res = switch (ylw, prp) {
 
 let rec atLeastOneFlushableChildAndNoWipNoPending composition atPriority => switch composition {
   | [] => false
-  | hd::tl =>
+  | [hd, ...tl] =>
       switch hd {
         | OpaqueGraph {lifecycle: Reconciled (_, [])} =>
             atLeastOneFlushableChildAndNoWipNoPending tl atPriority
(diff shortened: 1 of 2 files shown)


SanderSpies:
Related to #488

yunxing:
My concern with this diff is that it is going to confuse existing ocaml programmers as they are used to `::`, and the error message looks confusing (although this is the best we can give with our current tooling).

If we want to merge this, we should at least provide some better error messages so they understand what the alternative is. 

I'm going push a fix to #518. With that fix, a new api will be provided for you so you can put more details in the error message.

hhugo:
note that `::` can be user defined in OCaml 4.03 (https://github.com/ocaml/ocaml/pull/234/files).

jordwalke:
@hhugo I believe that is seamlessly supported in this case because the Reason Syntax simply treats the text representation as a "view" of the AST, and we parse *into* the tree that creates Cons's out of `::`. So if you were opening a `Prelude` that redefined `::`, then the list expressions and pattern matching syntax of `Reason` is and will remain fully compatible with that custom definition (modulo any bugs we're not aware of).

yunxing:
@SanderSpies Looks good, thanks!

PR #519, opened 2016-05-23
Minor cleanup

yunxing:
It is confusing to have both `Reason_util` and `Reason_utils`. This patch will change `Reason_utils` to `Syntax_util` as it has more syntax related utilities.


jordwalke:
Great! Merge when you are ready.

Issue #507, opened 2016-05-20
Can't pattern-match a list as [a, ...as]


yawaramin:
On compiling the following:

```
let xs = [1, 2, 3];
let [a, ...as] = xs;
```

We get the compile error:

```
File "Hello.re", line 2, characters 11-13:
Error: 745: <SYNTAX ERROR>
```

If we change the pattern to `[b, ...bs]` the compile succeeds.

yawaramin:
Whoops. `as` is a keyword.

yunxing:
Good catch (I was confused also). 

I think in this case we should change the error message of error state 745. Reopening it.

chenglou:
@yunxing ping me on your diff if you make one. I wanna see menhir error warning in action!

yunxing:
@chenglou Sure. I'm planning to write a detailed doc about how to contribute to menhir error warning database in the weekend so other people can contribute to it leisurely.

PR #520, opened 2016-05-23
Add error message for error state 745

yunxing:
#507

Error message for invalid state after `...`

@chenglou @yawaramin

### DIFF
--- a/src/reason_parser.messages
+++ b/src/reason_parser.messages
@@ -398,7 +398,7 @@ parse_pattern: LBRACKET UNDERSCORE COMMA DOTDOTDOT WITH
 ## pattern_comma_list COMMA DOTDOTDOT
 ##
 
-<SYNTAX ERROR>
+Expecting a valid list identifier.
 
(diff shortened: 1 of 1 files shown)


chenglou:
Woah, that's it? Nice...
Is it possible to warn more precisely? E.g. "instead you've put `as`"

yunxing:
I was thinking if we should do that or integrate it with Better_error_messages? (since you will see the original input anyway with Better_error_messages).

chenglou:
I'd rather defer as much as possible to the compiler/parser, since not everyone will use BetterErrors. Actually, the majority won't be using it at all; hopefully they'll be staying within the editor and not needing to see any error after compilation. The earlier the warning the better.

That being said, there _is_ a  category of error messages that should definitely stay in BetterErrors-land. E.g. the ones that concerns the ecosystem, best practices, and all other moving domains:
screenshot 2016-05-22 23 13 00

yunxing:
@chenglou Good point.

Issue #406, opened 2016-05-05
merlin-destruct generates ocaml even in reason files


jberdine:
It seems that merlin-destruct does not convert the ocaml code it inserts into reason.

PR #524, opened 2016-05-23
[Installation] Pin merlin and merlin-extend to master since reason branch is bgone

clentfort:
The `reason-0.0.1`-branches are gone from both `merlin` and `merlin-extend`, we should no longer point to those.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,8 +24,8 @@ opam init
 opam switch 4.02.3
 eval `opam config env`
 opam update
-opam pin add -y merlin 'https://github.com/the-lambda-church/merlin.git#reason-0.0.1'
-opam pin add -y merlin_extend 'https://github.com/let-def/merlin-extend.git#reason-0.0.1'
+opam pin add -y merlin 'https://github.com/the-lambda-church/merlin.git'
+opam pin add -y merlin_extend 'https://github.com/let-def/merlin-extend.git'
 opam pin add -y reason 'https://github.com/facebook/reason.git#0.0.6'
 
 ```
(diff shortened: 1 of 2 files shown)


yunxing:
Thanks! My understanding is that merlin's master gets updated pretty frequently. Should use a known reliable hashtag here instead?

cristianoc:
Note that Reason does not build with the latest merlin-extend:
```
File "src/ocamlmerlin_reason.ml", line 44, characters 36-49:
Error: Signature mismatch:
       ...
       The value `pretty_print' is required but not provided
```
It should be just a matter of implementing `pretty_print` which has been added to the signature.

jordwalke:
The `pretty_print` issue is a Good First Task: https://github.com/facebook/reason/issues/406

@clentfort I see those tags for `reason-0.0.1` here. Did you say they were gone?
 https://github.com/the-lambda-church/merlin/tags

clentfort:
Interesting, could swear they weren't there yesterday! Well then it might be better to continue linking to those tags. Leading to another question: Wouldn't it make more sense to pin `merlin` as `merlin-reason` or something to prevent conflicts with other installations of merlin that are not optimized for Reason? I don't know anything about the OCaml eco-system but couldn't lead this to conflicts?

let-def:
Once [this](https://github.com/let-def/Reason/blob/master/src/ocamlmerlin_reason.ml#L47) is fixed, there will be no need for an optimized `merlin-reason`.
At that point, I will release merlin 2.5.1 and merlin-extend on opam, and the next version of Reason should work without merlin-specific pins.

Issue #510, opened 2016-05-21
rtop throws "Error: Reference to undefined global `Reason_util'"


Hobinx:
I followed the install instruction on OS X, and the installation test worked fine, but starting `rtop` throws:

```
Exception: "Assert_failure src/reason_util.ml:17:4".
File "/Users/###/.opam/4.02.3/bin/rtop_init.ml", line 1:
Error: Reference to undefined global `Reason_util'
```

yunxing:
Hi @Hobinx, thanks for reporting!

Could you try to set the environment variable "OCAMLRUNPARAM" to "b" and try again?

kayceesrk:
@Hobinx Do you happen to have a `~/.ocamlinit` file? Mine had `#camlp4o;;`, which caused this error. Reason is not compatible with camlp4, and commenting this out makes the error go away for me.

Hobinx:
@kayceesrk Yeah, I have a `~/.ocamlinit` file, and I comment out  `#camlp4o`, but same error again when I started the `rtop`.

Hobinx:
@yunxing I set the environment variable "OCAMLRUNPARAM" to "b" and run `rtop`, the same error message prints out. Looks like `rtop` is not loaded, the toplevel is just `utop`, because when I type `[1, 2, 3]` in it, it prints out the type `(int * int  * int) list`, rather than normal list in Reason's basic syntax.

trevorriles:
@Hobinx I deleted the `~/.ocamlinit` file, and re-ran `opam init` and ``eval `opam config env``` and I could run `rtop`

yunxing:
@trevorriles @Hobinx hmm looks like the issue is in ~/.ocamlinit file. Could you please share yours so I can reproduce it on my end?

Hobinx:
@yunxing 
```
(* Added by OPAM. *)
let () =
  try Topdirs.dir_directory (Sys.getenv "OCAML_TOPLEVEL_PATH")
  with Not_found -> ()
;;
#use "topfind";;
#thread;;


#camlp4o;; 
#require "core.top";;
#require "core.syntax";;
```

Hobinx:
@trevorriles Yeah, you are right. I deleted the `.ocamlinit` file, and re-ran `opam init` and ``eval `opam config env` ``,  and problem solved.

jordwalke:
Should we have a `.reasoninit` instead?

yunxing:
[...]

PR #527, opened 2016-05-24
Use .reasoninit instead of .ocamlinit

yunxing:
By changing the loading order in rtop_init, I believe rtop can use
whatever directives that are valid in .ocamlinit (including #camlp4o).

The only requirement is that the loaded file should be in reason syntax. Thus
we have to create a new file.

#510

### DIFF
--- a/src/rtop.sh
+++ b/src/rtop.sh
@@ -20,4 +20,13 @@ if [[ $@ =~ "stdin" ]]; then
    export stdin=1
 fi
 
+if [ ! -f $HOME/.reasoninit ]; then
+    # TODO: ideally we should generate this file by refmtting
+    # .ocamlinit. But refmt doesn't support toplevel formatting yet
+    echo "/* Added by rtop */
+try Topdirs.dir_directory (Sys.getenv \"OCAML_TOPLEVEL_PATH\")
+with Not_found => ()
+" > $HOME/.reasoninit
+fi
+
 utop -init $DIR/rtop_init.ml -I $HOME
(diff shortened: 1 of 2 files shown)


kayceesrk:
This is a good idea. For the very least, it remove the mysterious `Reference to undefined global Reason_util'`. By making it an explicit process to create `.reasoninit`, we can avoid such unpleasant bugs.

yunxing:
@kayceesrk Yes. I think in the future we can also mirror `.ocamlinit` to `.reasoninit` if we add toplevel_directives support to Reason. 

As long as `.reasoninit` is loaded after `#require "reason";;`, users should be able to use `#camlp4o` without any issue. But this requires `.reasoninit` to be written in `reason` syntax.

jordwalke:
I just repinned master and it looks like this fails rtop for me because it installed a dummy `.reasoninit` file which is written in OCaml syntax, and rtop fails with a parsing error.

yunxing:
@jordwalke Could you post the reasoninit file that got installed?

jordwalke:
Here's what was added:
```ocaml
/* Added by rtop */
try Topdirs.dir_directory (Sys.getenv "OCAML_TOPLEVEL_PATH")
with Not_found => ()


```

I gave the correct version above.

jordwalke:
That has a Reason comment, but OCaml `try with`

yunxing:
Working on a fix. I didn't detect this because my .reasoninit was empty.

yunxing:
hmm, I just figured out an issue with this approach: Reason syntax can change over time, .reasoninit should be auto updated as well.

jordwalke:
And when you have multiple versions installed in various projects, which should be used to parse it? I believe that `.reasoninit` should be local to the sandbox, which are per project. Then there should also be a global sandbox, and that global sandbox can look for its init file in the home directory.

PR #532, opened 2016-05-25
Add test for end of line comments on module include lines

jordwalke:


jordwalke:
Merging because low risk, just a test case.

PR #534, opened 2016-05-25
Catch exception `Not_found` from error message database

yunxing:
- Catch exception `Not_found` from error message database and provide a nicer error
- Fix some typos in existing error messages

#553


yunxing:
Merging as this is low risk.

PR #539, opened 2016-05-25
Fix old syntax `functor`, should be `fun` instead.

IwanKaramazow:
Functor doesn't compile, it should be fun instead. 
Replaced the faulty entries in the docs with syntax that compiles, i.e. functor becomes fun.

### DIFF
--- a/modules.html
+++ b/modules.html
@@ -277,7 +277,7 @@ keyword designates functor in this case.
 module type HasT = {
   type t;
 };
-let module Pairify = func (X: HasT) => {
+let module Pairify = fun (X: HasT) => {
    type p = (X.t, X.t);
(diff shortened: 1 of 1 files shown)


modules.html
```diff
@@ -277,7 +277,7 @@
 module type HasT = {
   type t;
 };
-let module Pairify = functor (X: HasT) => {
+let module Pairify = func (X: HasT) => {
```
chenglou: `fun`?

IwanKaramazow: oops, typo. should be `fun` indeed

chenglou:
Will merge after the `func` fix.

IwanKaramazow:
Fixed :)

chenglou:
Thanks!

Issue #419, opened 2016-05-10
avoid dependencies for the core of reason


bobzhang:
the core part of reason is the parser, which currently consists of these modules: `reason_config`, `reason_utils`, `reason_lexer`, `reason_parser`.

In module `reason_utils`, it used a function `Str.global_replace` which can be easily replaced by a little function. (It introduced dependencies on `str` library, which  is not supported by js backend, also `str` itself relies on c stubs, better to avoid it)

about menhir, according to the manual

>  Can I ship a generated parser while avoiding a dependency on MenhirLib? Yes. One option is to use the code-based back-end (that is, to not use --table). In this case, the generated parser is self-contained. Another option is to use the table-based back-end (that is, use --table) and include a copy of the files menhirLib.{ml,mli}togetherwiththegeneratedparser.Thecommandmenhir --suggest-menhirLibwill tell you where to find these source files.

So, if we use `--table`, we just need check in generated files and two other files `menhirLib.{ml,mli}`

bobzhang:
so for real use (not playground), there is no need to get rid of dependencies on str, the only thing is to check in menhir generated code and (menhirLib.ml(i)), and you are happy to keep the core dependency minimal
see https://github.com/bloomberg/bucklescript/issues/396

yunxing:
@bobzhang Thanks. Now I understand the motivations behind removing the dependencies. I will start working on this soon.

yunxing:
@bobzhang In order to support bucklescript, do you also want us to provide build procedures based on the generated Menhir file, or are you going to integrate it into bucklescript ?

bobzhang:
@yunxing thanks for having a look, I have a pack tool which will bundle all files from reason into a single ocaml file, so there is no build procedure required, we only need the core of reason does not introduce 3rd party dependencies, in that case, I would also need include the source of the third party library.

Btw, `str` is fine, since it is in the core distribution of ocaml

yunxing:
@bobzhang nice, good to know. I will write a replacement for Str.

bobzhang:
@yunxing @jordwalke  is it possible to check in generated ml file from menhir ?  
We do have menhir in production, however, menhir required by reason is the latest one. I see https://github.com/jordwalke/npm-menhir, so I might give it a try if the decision is not checking in generated files, the main issue with npm-menhir is that it takes too much time to dowload and compile and I see no obvious disadvantages if we check in generated files, it would save reason's user's time.

yunxing:
[...]

PR #543, opened 2016-05-26
Remove Str in syntax_util

yunxing:
Remove `Str` in syntax_util by providing a custom string replace function.

#419

### DIFF
--- a/src/syntax_util.ml
+++ b/src/syntax_util.ml
@@ -3,15 +3,19 @@ open Asttypes
 open Parsetree
 open Longident
 
+external unsafe_get : string -> int -> char = "%string_unsafe_get"
+
 (** [is_prefixed prefix i str] checks if prefix is the prefix of str
   * starting from position i
   *)
-let is_prefixed prefix i str =
+let is_prefixed prefix str i =
   let len = String.length prefix in
-  if i + len > String.length str then
-    1
-  else
-    String.compare (String.sub str i len) prefix
+  if i + len > String.length str then false else
+  let rec loop j =
+    if j > len then true else
+      if unsafe_get prefix j <> unsafe_get str (i + j) then false else loop (j + 1)
+    in
+  loop 0
 
 
 let rec replace_string_ old_str new_str i str buffer =
@@ -19,7 +23,7 @@ let rec replace_string_ old_str new_str i str buffer =
     ()
   else
     (* found match *)
-    if 0 = is_prefixed old_str i str then
+    if is_prefixed old_str str i then
       (* split string *)
       let old_str_len = String.length old_str in
       Buffer.add_string buffer new_str;
--- a/src/syntax_util.ml
+++ b/src/syntax_util.ml
@@ -3,7 +3,39 @@ open Asttypes
 open Parsetree
 open Longident
 
-let a = Str.global_replace
+(** [is_prefixed prefix i str] checks if prefix is the prefix of str
+  * starting from position i
+  *)
+let is_prefixed prefix i str =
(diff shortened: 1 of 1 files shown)


jordwalke:
This is great, thanks @yunxing!

src/syntax_util.ml
```diff
@@ -3,7 +3,32 @@ open Asttypes
...
+(** [is_prefixed prefix str] checks if prefix is the prefix of str *)
+let is_prefixed prefix str =
+  let len = String.length prefix in
+  if len > String.length str then
+    1
+  else
+    String.compare (String.sub str 0 len) prefix
+
+(** [replace_string old_str new_str str] replaces old_str to new_str in str *)
+let rec replace_string old_str new_str str =
```
cristianoc: How long is `str` in practice here?
When `old_str` is not contained in `str`, for example, there will be a linear number of copies of (prefixes of) `str`, all in memory (not tail recursive), for quadratic time+space.

Module `Buffer` from the standard library allows to do concatenation in amortized linear time. Also, one could keep an index for the visited segment of `str`, instead of making a copy.

yunxing:
@cristianoc You are right. I didn't pay much attention to the efficiency of the algorithm since the str here is usually identifiers and the length shouldn't be too long.

Using `Buffer` and index makes sense. I will try them.

bobzhang:
I think efficiency does not matter too much here

yunxing:
Updated with some performance improvements as suggested by @cristianoc

cristianoc:
@yunxing great, merge away!

src/syntax_util.ml
```diff
@@ -3,7 +3,39 @@ open Asttypes
...
-let a = Str.global_replace
+(** [is_prefixed prefix i str] checks if prefix is the prefix of str
+  * starting from position i
+  *)
+let is_prefixed prefix i str =
+  let len = String.length prefix in
+  if i + len > String.length str then
+    1
+  else
+    String.compare (String.sub str i len) prefix
```
hhugo: There is no need to allocate a string here. Especially because `is_prefixed` is called at every step in `replace_string`.

yunxing: Any hints for how to not allocate a substring here ?

hhugo: https://github.com/dbuenzli/astring/blob/master/src/astring_string.ml#L80

yunxing: Thanks! That's helpful.

src/syntax_util.ml
```diff
@@ -3,23 +3,27 @@ open Asttypes
...
-let is_prefixed prefix i str =
+let is_prefixed prefix str i =
   let len = String.length prefix in
-  if i + len > String.length str then
-    1
-  else
-    String.compare (String.sub str i len) prefix
+  if i + len > String.length str then false else
+  let rec loop j =
+    if j > len then true else
```
cristianoc: `j >= len`

src/syntax_util.ml
```diff
@@ -3,23 +3,27 @@ open Asttypes
...
+      if unsafe_get prefix j <> unsafe_get str (i + j) then false else loop (j + 1)
```
cristianoc: `String.unsafe_get` will do, no need for the `external` declaration.

src/syntax_util.ml
yunxing: oops

src/syntax_util.ml
yunxing: Good point!

PR #548, opened 2016-05-27
Fixes in `mlCompared`

raphael-proust:
There are some syntax errors in the OCaml examples.

### DIFF
--- a/mlCompared.html
+++ b/mlCompared.html
@@ -364,8 +364,8 @@ there are separate syntaxes for tuple types `(x * y)` and tuple values values
 >     <td>
 >       <pre>
 > type r =
->   {x: int, y: int};
-> let myRec: r = {x = 0, y = 10};</pre>
+>   {x: int; y: int};
+> let myRec: r = {x = 0; y = 10};</pre>
 >
 >     </td>
 >     <td>
@@ -628,7 +628,7 @@ type and returns a new type.
 let x: int list = [2]
 type listOfListOfInts =
   int list list
-type ('a, 'b) tup = ('a, 'b)
+type ('a, 'b) tup = ('a \* 'b)
 type pairs = (int, int) tup list
 let tuples: pairs = [(2, 3)]</pre>
     </td>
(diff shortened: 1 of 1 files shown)


facebook-github-bot-7:

mlCompared.html
```diff
@@ -628,7 +628,7 @@
 let x: int list = [2]
 type listOfListOfInts =
   int list list
-type ('a, 'b) tup = ('a, 'b)
+type ('a, 'b) tup = ('a \* 'b)
```
jordwalke: I think these parenthesis should be omitted (though it won't matter in this case, but would in the case of variant constructors IIRC).

jordwalke:
Thank you! Please see the github bot above.

While you're here, I noticed you [forked the OCaml pprint library](https://github.com/raphael-proust/ocaml-pprint): Can you say how that compares to OCaml's Format library and if we should be using it for pretty printing the Reason syntax?

raphael-proust:
Re: `pprint` library.

`PPrint` makes more composable printers at the cost of some verbosity. In my personal opinion this verbosity is actually better than the tersness of OCaml's Format module: things are more explicit, things have names (instead of weird combinations of `%`, `@`, `[]` and such).

Another alternative is `fmt` which builds combinators on top of the standard library's `Format`. The documentation is quite extensive. It also offers (ANSI) terminal colour handling.

raphael-proust:
Re: CLA.

I don't have time right now to read the legalese and figure out what it means in conjunction with local laws and previous agreements. If you want, you can get some already CLA-checked experienced OCamler to proof-read. They will probably come to a similar patch on their own. (Note that there are other things to fix: some examples are non-idiomatic, some examples present only one of the possible alternative.)

mlCompared.html
raphael-proust: It would only matter in the variant constructor of a type declaration indeed.

yunxing:
Thanks for the contribution. The diff looks good. Just confirmed with our Open Source team that token changes are ok without signing a CLA.

Issue #505, opened 2016-05-20
rtop finishes parsing when not terminated with a `;` with no execution


simonjbeaumont:
```
Reason # let f = fun n => n;
let f : 'a => 'a = <fun>
Reason # f 1
Reason # f
Reason # f 1;
- : int = 1
```

chenglou:
cc @jordwalke @yunxing this is a bit troublesome because it swallows errors.
<img width="571" alt="screenshot 2016-05-26 20 55 39" src="https://cloud.githubusercontent.com/assets/1909539/15597372/3953fb2e-2384-11e6-9c0b-8ee91d313944.png">

jordwalke:
I don't think this was always the case.

yunxing:
@simonjbeaumont Sorry just saw this. I think this could be related to some of the stuff that I added to support rtop -stdin and use file. 

Let me see what's a good solution here.

yunxing:
This should have been fixed in trunk. Closing it.

PR #549, opened 2016-05-27
Fix rtop -stdin

yunxing:
- Completely throw away the hack that we put into to suppoert `rtop -stdin`.
Now use `refmt | utop -stdin` instead.

- Remove a hack in reason_parser.mly which causes
"rtop finishes parsing when not terminated with a `;` with no execution" (https://github.com/facebook/reason/issues/505). This hack was there to support `rtop -stdin`

### DIFF
--- a/src/reason_parser.mly
+++ b/src/reason_parser.mly
@@ -1129,7 +1129,6 @@ interface:
 toplevel_phrase: _toplevel_phrase {apply_mapper_chain_to_toplevel_phrase $1 default_mapper_chain}
 _toplevel_phrase:
     structure_item SEMI                 { Ptop_def [$1]}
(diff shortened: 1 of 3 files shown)


jordwalke:
There are issues with the OCaml source printer and there will likely remain issues with it until it is replaced with something like `refmt` for Reason syntax.
It seems this causes us to rely on it for accuracy? I'd prefer this diff to what currently exists (incorrect parsing on newlines in `rtop`), but I think there's got to e a better way, right?

yunxing:
Back online. Yes it will be depending on refmt for accuracy. Based on current use cases (mostly do a quick test for simple statement) I think it is fine for now.

Issue #556, opened 2016-05-28
How to use dependencies like Core with Reason & rebuild?


davidpelaez:
I had never taken the time to learn OCaml but Reason does indeed feel more accessible to me. Thank you for contributing to make OCaml easier for people like me! I'm using the weekend to get started and trying to follow Real World OCaml examples with Reason syntax. I've been unable to `open Core.Std`. Here's the question I wrote in StackOverflow:

I'm new to OCaml and I'm trying to try (:P) Facebook Reason syntax. I cannot find a way to make this compile because if cannot find the Core module (already installed with opam).

I'm trying to compile a sample program from Real World OCaml

```
open Core.Std;

let rec read_and_accumulate accum => {
  let line = In_channel.input_line In_channel.stdin;
  switch line {
    | None => accum
    | Some x => read_and_accumulate (accum +. Float.of_string x)
  }
};

let () = printf "Total: %F\n" (read_and_accumulate 0.);
```

This is the command I'm using for compilation: `rebuild accum.native.`

When I have this in _tags (from the instructions in https://janestreet.github.io/installation.html)

```
true: package(core,ppx_jane)
true: thread,debug
```

My error changes but I still don't know what to do:

```
File "_tags", line 1, characters 6-28:
Warning: tag "package" does not expect a parameter, but is used with parameter "core,ppx_jane"
File "_tags", line 1, characters 6-28:
Warning: the tag "package(core,ppx_jane)" is not used in any flag or dependency declaration, so it will have no effect; it may be a typo. Otherwise you can use `mark_tag_used` in your myocamlbuild.ml to disable this warning.
+ /Users/David/.opam/4.02.3/bin/ocamldep.opt -modules -pp refmt -impl accum2.re | tee accum2.re.depends accum2.ml.depends
accum2.re: Core Float In_channel
+ /Users/David/.opam/4.02.3/bin/ocamlc.opt -c -g -thread -pp '-g -thread' -pp refmt -o accum2.cmo -intf-suffix .rei -impl accum2.re
File "accum2.re", line 1, characters 5-13:
Error: Unbound module Core
Command exited with code 2.
Compilation unsuccessful after building 2 targets (0 cached) in 00:00:00.
```

What do I have to do to use Core with Reason?

Adopting the syntax is pretty easy and I have only been reading for a couple hours, but there's zero docs about how to use Reason for non OCaml users.

[...]

PR #558, opened 2016-06-01
[rebuild] Correctly handle `useocamlfind`

yunxing:
#556 

Checks if `use-ocamlfind` is presented. If it is, sets OCAMLFIND_COMMANDS so that ocamlfind knows how to use `reopt` to parse reason files.


yunxing:
Merging for now. Feel free to reopen if any issue is found.

Issue #491, opened 2016-05-19
"of" word is not needed.


jordwalke:
```
type something = 
   | Constructor of int int
   | Another of (list int);
```

Multiple people have suggested removing the need for the `of` keyword. Sounds great to me, if possible and doesn't introduce grammar conflicts.

@kayceesrk 

Tagging this as "good first task" for newcomers in the community to give it a shot. It should be relatively straightforward.

SanderSpies:
Should exception also change? 

For example: 

`exception Key_not_found of string`

would become 

`exception Key_not_found string`

yunxing:
I think we should also change exception for consistency.

kayceesrk:
Exceptions should also change. `exception Key_not_found of string` is really syntactic sugar for extending the inbuilt extensible variant type `exn` and is equivalent to `type exn += Key_no_found of string`.

SanderSpies:
Alright, I want to take a go at this issue.

SanderSpies:
So I've been breaking my head on this issue - as this is really unfamiliar territory for me. I made a change to `reason_parser.mly` and I'm running into a reduce/reduce issue. 

This is the reason_parser.conflicts that I'm getting: 

```
** Conflict (reduce/reduce) in state 96.
** Token involved: LPAREN
** This state is reached from type_kind after reading:

EQUAL UIDENT 

** The derivations that appear below have the following common factor:
** (The question mark symbol (?) represents the spot where the derivations begin to differ.)

type_kind 
(?)

** In state 96, looking ahead at LPAREN, reducing production
** constr_ident -> UIDENT 
** is permitted because of the following sub-derivation:

EQUAL constructor_declarations 
      constructor_declaration_no_leading_bar 
      constr_ident generalized_constructor_arguments attributes // lookahead token appears because generalized_constructor_arguments can begin with LPAREN
      UIDENT . 

** In state 96, looking ahead at LPAREN, reducing production
** mod_ext_longident -> UIDENT 
** is permitted because of the following sub-derivation:

EQUAL core_type 
      mark_position_typ(_core_type) 
      _core_type 
      core_type2 
      mark_position_typ(_core_type2) 
      _core_type2 
      non_arrowed_core_type 
      non_arrowed_non_simple_core_type 
      mark_position_typ(_non_arrowed_non_simple_core_type) 
      _non_arrowed_non_simple_core_type 
      type_longident non_arrowed_simple_core_type_list 
      mod_ext_longident DOT LIDENT 
[...]

PR #563, opened 2016-06-03
#491 - Remove "of" word.

SanderSpies:
Changes: 
- remove of from lexer
- remove of from messages file
- remove of from tests
- remove of from printer
- remove of from parser generator

Most important part is the change within reason_parser.mly.

Thanks @yunxing for the tips.

### DIFF
--- a/src/reason_parser.mly
+++ b/src/reason_parser.mly
@@ -3552,13 +3552,7 @@ constructor_declarations:
 ;
 
 constructor_declaration_no_leading_bar:
-  | as_loc(UIDENT) generalized_constructor_arguments attributes
-      {
-       let args,res = $2 in
-       let loc = mklocation $symbolstartpos $endpos in
-       Type.constructor $1 ~args ?res ~loc ~attrs:$3
-      }
-  | as_loc(constr_ident2) generalized_constructor_arguments attributes
+  | as_loc(constr_ident) generalized_constructor_arguments attributes
      {
         let args,res = $2 in
         let loc = mklocation $symbolstartpos $endpos in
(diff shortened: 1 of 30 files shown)


src/reason_parser.mly
```diff
@@ -3553,7 +3552,7 @@ constructor_declarations:
 ;
 
 constructor_declaration_no_leading_bar:
-  | as_loc(constr_ident) generalized_constructor_arguments attributes
+  | as_loc(UIDENT) generalized_constructor_arguments attributes
```
yunxing: I think you should also add a production here for `constr_ident` without `UIDENT`.

src/reason_parser.mly
```diff
@@ -4130,8 +4129,15 @@ mod_longident:
...
     UIDENT                                      { Lident $1 }
   | mod_ext_longident DOT UIDENT                { Ldot($1, $3) }
-  | mod_ext_longident LPAREN mod_ext_longident RPAREN { lapply $1 $3 }
+  | mod_ext2 { $1 }
+;
+
+mod_ext2:
+   | mod_ext_longident DOT UIDENT LPAREN mod_ext_longident RPAREN { lapply ( Ldot($1, $3)) $5 }
+   | mod_ext2 LPAREN mod_ext_longident RPAREN { lapply $1 $3 }
+   | UIDENT LPAREN mod_ext_longident RPAREN { lapply (Lident($1)) $3 }
```
yunxing: This is smart!

src/reason_parser.mly
SanderSpies: I just did.

yunxing: Actually, instead of splitting `constr_ident` into two parts (UIDENT + non-UIDENT), there is another trick that I just used recently, which is to use `%inline`on constr_ident like:

```
%inline constr_ident:
    UIDENT                                      { $1 }
  | LPAREN RPAREN                               { "()" }
  | COLONCOLON                                  { "::" }
  | FALSE                                       { "false" }
  | TRUE                                        { "true" }
;
```

This trick basically expands into what you are manually writing now. For more information, take a look at:
[menhir's manual](http://gallium.inria.fr/~fpottier/menhir/manual.pdf), chapter 5.3

SanderSpies: Interesting, thanks.

yunxing:
Thanks so much!

Issue #375, opened 2016-04-24
Do not force break for "small" if statements.


jordwalke:
In cases like the following, forcing a break is really ugly.
```rust
let res =
   myFunction
     arg
     (if (x) {expr} else {expr2});
```

So let's change the `if` formatting to not force breaks if:
- There is only one sequence item inside of the `{}`
- There is no "else if".

This is a "great first task" for someone looking to help out.

SanderSpies:
I'm "claiming" this issue.

jordwalke:
It is actually quite challenging believe it or not. I have tried twice and failed.

SanderSpies:
Ah, guess @cristianoc is doing this one! Then I pick up something else.

cristianoc:
Sorry @SanderSpies I was trying to assign the task to you, then got my name by mistake.

SanderSpies:
@cristianoc No worries - thanks for assigning.

jordwalke:
You could take a look and document why it's challenging.

SanderSpies:
@jordwalke Seems you already did that: https://github.com/mjambon/easy-format/issues/12. 

Is there btw. already a function to count sequences or more ideal a sequence "threshold" function?

jordwalke:
I don't believe such a thing exists, but the challenge was mostly in the easy-format issue. Good find.

SanderSpies:
Close please (fixed already(

yunxing:
Closing this. 

@SanderSpies I made a small change on top of your diff so it would still break if the if statement can't fit into a single line.

PR #569, opened 2016-06-09
Print if on a single line.

SanderSpies:
Related to #375

Not finished yet. Opening this PR for discussion.

It's not clear to me yet how long constants or idents (e.g "this is a a very long constant string") should be handled.

Also, I think constants and idents are the only ones that I need to check for. Please correct me if I'm wrong in my assumptions.

Other recommendations are of course also welcome.


formatTest/unit_tests/expected_output/if.re
```diff
@@ -84,13 +70,7 @@ if printIfFirstArgGreater {
...
 };
 
-if (
-  if x {
-    true
-  } else {
-    false
-  }
-) {
+if (if x {true} else {false}) {
```
kassens: I find the original easier to read, even if it uses up more lines.

src/reason_parser_message.ml
```diff
@@ -10,29 +10,29 @@ let message =
         "<SYNTAX ERROR>\n"
```
kassens: Just sent #570 to get these changes out. I ran into this as well.

formatTest/unit_tests/expected_output/if.re
SanderSpies: I'm pretty neutral on this, besides that an if in an if test is a bit confusing.

jberdine:
I like the short form.  I think that when evaluating choices like this, it is important to consider what the alternatives look like in the context of, perhaps a lot of, other code.  Once surrounded by other code, the value the short forms give by allowing more code to be seen at once is significant.

src/reason_parser_message.ml
yunxing: This will gone after rebasing against master.

SanderSpies:
Closed because #621

PR #570, opened 2016-06-09
Update for reason parser state changes

kassens:
This diff is the result of just running `make test` on my machine. It looks like
some state machine values changed.

I have no idea what I'm doing here. I assume this has nothing to do with my setup and someone just forgot to update these files.


SanderSpies:
My bad probably, however I don't think this is important right now as it will change anyway when something in the parser changes.

SanderSpies:
OTOH this is probably nicer in the git history.

kassens:
Seems low risk, so I'm just gonna merge this.

PR #571, opened 2016-06-09
Wrap type definitions on one line per variant

### LINKED ISSUE #546: [refmt] Add newline before each type alternative
I think the formatter should add newlines before the `|` in type definitions.

```rust
/* input and desired(?) output */
type binTree 'x =
  | Node of (binTree 'x) (binTree 'x)
  | Leaf of 'x;

/* current refmt output */
type binTree 'x = | Node of (binTree 'x) (binTree 'x) | Leaf of 'x;
```

kassens:
Closes #546.

### DIFF
--- a/src/reason_pprint_ast.ml
+++ b/src/reason_pprint_ast.ml
@@ -2152,7 +2152,7 @@ class printer  ()= object(self:'self)
         ]
       (* EQUAL private_flag BAR constructor_declarations {(Ptype_variant _, $2, None)} *)
       | (Ptype_variant lst, scope, None) ->  [
-          privatize scope [makeList ~break:IfNeed ~postSpace:true ~inline:(true, true) (List.map self#type_variant_leaf lst)]
+          privatize scope [makeList ~break:Always_rec ~postSpace:true ~inline:(true, true) (List.map self#type_variant_leaf lst)]
         ]
       (* EQUAL DOTDOT {(Ptype_open, Public, None)} *)
       | (Ptype_open, Public, None) -> [
--- a/formatTest/typeCheckedTests/expected_output/arityConversion.re
+++ b/formatTest/typeCheckedTests/expected_output/arityConversion.re
@@ -11,7 +11,9 @@ let b =
   MultiArgumentsConstructor 1 2 [@implicit_arity];
 
 let module Test = {
-  type a = | And (int, int) | Or (int, int);
+  type a =
+    | And (int, int)
+    | Or (int, int);
 };
 
 Test.And (1, 2);
--- a/formatTest/typeCheckedTests/expected_output/attributes.re
+++ b/formatTest/typeCheckedTests/expected_output/attributes.re
@@ -36,7 +36,8 @@ type attributedIntsInTuple = (
 )
 [@@onTopLevelTypeDef];
 
-type myDataType 'x 'y = | MyDataType 'x 'y;
+type myDataType 'x 'y =
+  | MyDataType 'x 'y;
 
 type myType =
   (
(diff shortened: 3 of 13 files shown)


src/reason_pprint_ast.ml
```diff
@@ -2152,7 +2152,7 @@ class printer  ()= object(self:'self)
         ]
       (* EQUAL private_flag BAR constructor_declarations {(Ptype_variant _, $2, None)} *)
       | (Ptype_variant lst, scope, None) ->  [
-          privatize scope [makeList ~break:IfNeed ~postSpace:true ~inline:(true, true) (List.map self#type_variant_leaf lst)]
+          privatize scope [makeList ~break:Always_rec ~postSpace:true ~inline:(true, true) (List.map self#type_variant_leaf lst)]
```
kassens: This seems to do the right thing, but I'm not sure if `Always` should be used here instead?

If I use `Always`, it doesn't insert a newline after the `=`, so maybe it's missing a forced newline there and drop the `_rec`?

SanderSpies: This seems right.

See also: https://github.com/mjambon/easy-format/blob/6d16b6a9c71fa98386ba785de159dc14bc385ad0/easy_format.mli#L48

kassens:
This more or less moves into the opposite direction to #569.

This PR is more liberal with newlines whereas the other compacts the code more. Opinions?

jberdine:
I still think that if the type definition is short enough to fit on one line and not self-evident enough, add comments, then it won't fit on one line anymore.  If the type definition is short and clear, it's easier to read the code as a whole if the definition is short, since it allows more context to be seen at once.

jordwalke:
Let's make this be configurable then. Next, we can focus on having per project formatting settings since this kind of thing is going to come up often.

AppShipIt:
Configuration feels like a slippery slope (see all the ESLint rules)?

jordwalke:
@AppShipIt How so? Surely, not everyone would agree on a single width to wrap source code at, right?

SanderSpies:
@jordwalke I would prefer us to make the decision as it would prevent countless of useless coding style discussions in the future.

AppShipIt:
> Surely, not everyone would agree on a single width to wrap source code at, right?

Right but as a former UX lead I look for the XY problem. Why do people wrap at different widths. One reason is readability similar to constraining website copy to 960 pixels. Another is display device. I code on my laptop with split editors and occasionally present with larger font for readability at a distance. If only those two use cases I'd suggest configurability is the wrong solution. Just constrain to content width. Except the final factor is consistency. A configurable width enables a team (using a shared editor config) to discourage complex expressions.

On the other hand this issue feels more semantic. Perhaps keeping simple statements on a single line will discourage nesting complex. But you could theoretically configure everything. What don't you configure? How do you name properties? How do you migrate their names as convention necessitates? That's the slippery slope I'm alluding to. Hash these TIMTOWTDI once so teams never have to.

yunxing:
Getting back to this. I personally think we should be more opinionated about configs so Reason code across projects can be more consistent and predictable. If people have really strong and different opinions about one specific behavior, we should then make it a configurable variable (like print-width). 

In this specific case, I think breaking them into multiple lines makes it much easier to read for me. If you consider the following test case:

image

vs 

image

With the new change, I can instantly tell the type has three different constructors. 

Let's get this merged. We can continue our discussion in the issue.

yunxing:
Closed via : https://github.com/facebook/reason/commit/0479a2b7e64716f2dd69ec46d8343e4139c5865b (fixed a test case on top of this).

jberdine:
This unnecessary expansion of type definitions makes them more heavyweight, which discourages them, and is exactly the wrong direction.

SanderSpies:
Yeah, this change is indeed in the wrong direction if you look at: https://github.com/facebook/reason/pull/571/files#diff-9055ce1ca76486a2757c3aa92264647cL42. My preference would go to only breaking typing definitions if the constructors contain arguments.

PR #577, opened 2016-06-12
Fix comments inside OR pattern on case list

### LINKED ISSUE #559: poor formatting of comments between switch cases
Consider:

```
let _ =
  switch x {
  | None
  | Some _ => ()
  };
```

If a comment is added between the cases:

```
let _ =
  switch x {
  | None
  /* between */
  | Some _ => ()
  };
```

then refmt is very heavy-handed:

```
let _ =
  switch x {
  | None
  | /* between */
    Some _ =>
    ()
  };
```

It would be good if the `|` and the pattern `Some _` weren't split up, and if the line breaking didn't change.

tadeuzagallo:
Fixes #559 

I also had to update a couple other tests, where comments on or patterns would be pushed to the next line, but I think it has the correct behaviour now.

### DIFF
--- a/formatTest/unit_tests/expected_output/wrappingTest.re
+++ b/formatTest/unit_tests/expected_output/wrappingTest.re
@@ -2216,8 +2216,8 @@ let blah =
 let blahCurriedX x =>
   fun
   /* Comment before first bar */
-  /* Comment between first bar and OR pattern */
-  | Red x
+  | /* Comment between first bar and OR pattern */
+    Red x
   | Black x
   | Green x => 1
   /* Comment before second bar */
--- a/formatTest/unit_tests/expected_output/wrappingTest.re
+++ b/formatTest/unit_tests/expected_output/wrappingTest.re
@@ -2216,7 +2216,8 @@ let blah =
 let blahCurriedX x =>
   fun
   /* Comment before first bar */
-  | /* Comment between first bar and OR pattern */ Red x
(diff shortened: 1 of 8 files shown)


SanderSpies:
You are on a roll! Nice.

formatTest/unit_tests/expected_output/wrappingTest.re
```diff
@@ -2216,7 +2216,8 @@ let blah =
 let blahCurriedX x =>
   fun
   /* Comment before first bar */
-  | /* Comment between first bar and OR pattern */ Red x
+  /* Comment between first bar and OR pattern */
+  | Red x
```
tadeuzagallo: This is probably the most questionable change, but should be easy to fix if the original is preferred.

SanderSpies: I would prefer this to be fixed, as the input is:
```
  /* Comment before first bar */
  | /* Comment between first bar and OR pattern */
    (Red x | Black x | Green x) => 1
```

tadeuzagallo: Now it should look like

```swift
/* Comment before first bar */
| /* Comment between first bar and OR pattern */
  Red x
| Black x
| Green x => 1
```

Is it ok?

yunxing:
Should be fixed along with the new comments interleaving code.

PR #579, opened 2016-06-13
[Formatter] Fix (?) leading/trailing newline

chenglou:
4d9adf2e47a0bf121f294293fea25441f22f7c79 introduced a leading new line to the formatter, which is slightly annoying upon formatting a file.
I'm a bit afraid of touching that diff's comment logic, so I used the catch-all solution of trimming the final input (then appending a newline).
This also has the effect of adding a newline to some existing tests which don't end with one. I think this is desired? If not, I can replace `trim` with a `trim_leading` or something.

@jordwalke @yunxing


chenglou:
@yunxing @jordwalke merging this because I'm OCD-ing over the missing/extra lines at places. Retrospective comments welcome!

Issue #568, opened 2016-06-08
"switch x" vs "match x with"


spencerwi:
What's the motivating factor for changing `match x with` to `switch x`? I understand the reason for the braces, but it seems like changing to `switch` just makes Reason less compatible with OCaml, while also causing confusion with `switch-case` and losing some clear readability.

From a readability standpoint, I read the following:

```
match x with {
    Some y => y
    None => 0
}
```

as "match x with Some y and return y, or with None and return 0", where 

```
switch x {
    Some y => y
    None => 0
}
```

maybe reads as "switch x to Some y then y, or None then 0", which doesn't really convey the same thing to new developers. A move to `switch x` over `match x with` loses clear readability for the sake of keystrokes (incidentally, the opposite choice from the one taken with the braces/no-braces tradeoff).

Calling this `switch x` also confuses it with other languages' switch-case constructs, which _are_ less powerful than pattern-matching, which may lead users to believe that Reason's pattern-matching is similarly limited.

jordwalke:
The later example is missing the "case bars". If you put those in the last example, and read them aloud as "case Some x", I wonder if that changes your mental flow when reading it.

> but it seems like changing to switch just makes Reason less compatible with OCaml

Reason is fully compatible with OCaml regardless of the particular syntax chosen for `switch`. I think you know this already, but I'm just clarifying for any readers out there who weren't aware.

> Calling this switch x also confuses it with other languages' switch-case constructs, which are less powerful than pattern-matching

This is now no longer true, with the introduction of Swift.

spencerwi:
I just wrote my sample based on the current Reason documentation (http://facebook.github.io/reason/mlCompared.html#reason-and-ocaml-pattern-matching).

By "less compatible", I mean it's one more thing that I have to change if I'm migrating a codebase from OCaml to Reason. It's one more obstacle between having the same code "just work" for both OCaml and Reason, which, granted, it's a different syntax, but usually that syntax should have a good -- pardon the pun -- reason for being different.

[...]

Issue #557, opened 2016-05-30
Announcing reason on the ocaml mailing list


braibant:
Any reason, `reason` was not announced on the ocaml mailing list at this point? It seems to me that those changes might make it even harder for the existing ocaml community to understand and read code written with the `reason` syntax. It could a nice occasion be nice to benefit from an healthy discussion with the rest of the OCaml community (before iterating even more through syntax changes). 

(For instance, for me, #550 seems not that intuitive.)

jordwalke:
No reason except lack of time, and wanting to have a better, easier to try out experience before broadcasting more widely (I'd like a js_of_ocaml top level to try immediately for example).
We didn't really expect Reason to get as much interest as it has already (1200+ stars in just a few days without much open source communication about it).

w.r.t. intuitiveness: While Reason *does* fix many of the pitfalls that experienced OCamlers report, the main goal was to fix pitfalls and friction that people totally *new* to the language typically experience. The choices we make to appeal to these different audiences will vary dramatically.
It's entirely possible that experienced OCamlers (who care primarily about collaborating with other OCamlers), won't find anything valuable in the new syntax, or even the upcoming build/namespace/share workflow. I think that's totally okay and those same people will likely enjoy the network effects of having more people care about their same underlying technology (at least I hope they would).
There are some tools in Reason that provide a nice small, very incremental path forward for the existing OCaml concrete syntax. Once you get all the syntax automatically formatted via Reason, you can *rapidly* evolve the syntax to be whatever you want it to be including a syntax that is mostly the same as the current OCaml syntax. It's not easy to accomplish such an automated formatter, but Reason has done so, or at least seeks to. Some have proposed that we can eventually use these same tools to help evolve the current OCaml syntax and I think that sounds great too.

braibant:
I understand that there has been a lot of activity around `reason`, which probably comes from the fact that it was broadcasted quite widely on hacker news.  

I think that `reason` is a very nice, if not straightforward, idea at it's core: that is, it's possible to change the syntax of an existing language without changing its type-system, its semantics nor its runtime (which are for me the parts that define a language). With that idea, and the nice reformating tools, it's possible to have a tight update loop experimenting with the syntax; and I look forward to see a version of the reformating tools geared toward the existing OCaml syntax. (To be fair, I also look forward to see new work going on in other parts of the reason toolchain and ecosystem, and I am very excited about the idea of the `OCaml/reason` community growing and exchanging.)  

[...]

PR #580, opened 2016-06-13
Alternative syntax for method calls - changes foo#bar to foo..bar.

### LINKED ISSUE #523: Alternative syntax for method calls.
It would be fantastic to be able to do 

```
foo.bar 1 2
```

instead of

```
foo#bar 1 2
```

Primarily, it would make the language more approachable for people coming from javascript, python, etc.

SanderSpies:
Fixes #523

Also related to #366


jordwalke:
I'd like to hear what others think of this.

tadeuzagallo:
I'm new here, and to OCaml, but this seems very counter intuitive to me. I'd expect it to be a range, but not a method call. I think I might have seen someone suggesting `::`, imho that'd be a better option.

SanderSpies:
@tadeuzagallo Agree on that it looks like a range, however using `::` makes it look like a list.

tadeuzagallo:
@SanderSpies Oh, didn't think about it. I guess I wouldn't use `::` for lists either, but that's probably out of the scope of this diff.

yunxing:
@SanderSpies Thanks for implementing this. `foo..bar` actually looks confusing to me (more like a range as suppose to method selection). Maybe `->` looks better?

SanderSpies:
@yunxing looks good to me.

@tadeuzagallo what do you think?

braibant:
I am following activity on `reason` and see many proposals for syntax changes that are "local" (in the sense that they consider one syntactic construction at a time). Yet, those changes have ramifications on other changes (e.g., this one on #581, #523). 

`reason` has some traction, and it would be nice not to confuse people more by making changes to the syntax that turn out to be blocking for something else. Would it make sense to discuss changes to the syntax more globally, through appropriate channels (for instance, using a single issue, rather than many different ones)? 

Also, as I pointed out in #557 (and was pointed out independently in #568), it would be nice if the changes in the reason syntax did not make the code not-understandable by people accustomed to the OCaml syntax (or less compatible in a sense).

jordwalke:
One thought: if `::` looks like a list, then named arguments already have that problem! However, `::` only resembles a list to people who use OCaml daily.

@braibant : Many of these proposed changes are just pull requests, many of them don't arise from any official discussions - they're just people getting their ideas out there in the best possible way - *code*. We welcome pull requests, even if they're silly or just a means of learning. When syntactic changes like this are very local, and have ramifications to the wider syntactic real estate, that will come up in discussion on the diff and it's a really great way for people to learn. So I'd say we can take individual, smaller PRs with a grain of salt and people who send them understand this as well. Then, in addition to that, as you suggested, we should have larger, more holistic conversations for big topics such as "named arguments" (you can see we have existing threads for those which cover a ton of edge cases and considers how it effects the entire grammar).

I am curious about one comment you made - that making the syntax immediately recognizable to OCaml programmers should be a goal. I agree it would be nice, but surely that comes at the cost of making it more difficult to recognize for non-OCaml programmers. If we have to choose one of those audiences to cater to, which one do you suggest and why? A few things to consider:

1. There are 99 non-OCaml programmers for every 1 OCaml programmer.
2. OCaml programmers eat new programming languages for breakfast and can learn them very quickly - many OCaml programmers use OCaml to implement new languages.
3. OCaml programmers already have a syntax that they love, correct?

bsansouci:
What about `->`? It might look like a pointer dereferencing for low level folks, but at least `myObj->method1->method2` won't look too out of place. OCaml folks might hate it though because it looks like a type declaration :/

jordwalke:
Thanks for the ideas, @bsansouci!

Another extreme idea I had thrown around earlier was that records would look more like variants, in that they have all leading capital letters. It makes sense when pattern matching or creating variants with records inside - capitals become the visual language for "pattern matchable thing":

```rust
let somePoint = Some {X: 0, Y: 100};
let res = switch {
  | None => 0
  | Some p => p.X + p.Y
};

let myObject = {
   myMethod: fun a b => a + b,
   private myPrivateMethod: fun a b => a + b
};
let result = myObject.myMethod 10 20;
let result = myObject.myPrivateMethod 10 20;
```

I decided against it because records are faster than objects (a better default) and it makes punning weird (`X` is punned to `x`). Instead, I leaned more towards coming up with a proposal for minimally extensible records that would retain performance, but satisfy many of the use cases for objects, and then I'd be less concerned with how ugly objects were.

Just my current thoughts, that I'm totally flexible on.

PR #581, opened 2016-06-13
Change #pattern to *pattern

### LINKED ISSUE #578: Change # (polymorphic variant abbreviation)  to `*` pattern match.
Related to #366 

Within OCaml it's possible to use polymorphic variant abbreviation patterns (match any of this polymorphic type) which use `#` as a marker. If we would change the current ` to # it would therefore become ambiguous (https://github.com/facebook/reason/pull/366#issuecomment-225347382).

To work around this ambiguity I think we should change the pattern match `#polymorphic-type` to `*polymorphic-type` or something similar.

SanderSpies:
Fixes #578.

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/mlVariants.re
+++ b/formatTest/typeCheckedTests/expected_output/mlVariants.re
@@ -22,5 +22,5 @@ type u2 = [ t2 | `c];
 
 let listPatternWithHash =
   fun
-  | #t2 => 1
+  | *t2 => 1
   | `c => 2;
--- a/formatTest/typeCheckedTests/expected_output/mlVariants.re
+++ b/formatTest/typeCheckedTests/expected_output/mlVariants.re
@@ -15,3 +15,12 @@ let sumThem =
   | `StillAnIntTuple (a, b) => a + b;
 
 type nonrec t = | A int | B bool;
+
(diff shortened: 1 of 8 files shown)


jordwalke:
Wow, I had no idea we could even use `#polymorphic-type` in a pattern match!
I see the motivation here. We should come up with a consistent language for representing row-polymorphic types and then try to use that in this case. It's likely not going to be the `*`. Perhaps your goal was merely to get it out of the syntactic real estate quickly so that we could use `#` for polymorphic variants? The syntax for polymorphic pattern type match likely should match the syntax for class types which also use the `#`. If we change one we should probably change the other.

SanderSpies:
Yes, the idea is to make it possible to use `#` for polymorphic variants. This PR was mostly to start the discussion on how to solve this. I see class types as something different though and from my understanding it's not related. Actually for this change I think we should focus on polymorphic variants instead. In that sense it should be somewhat related to the `#` that is being proposed, or we need to choose a different syntax.

jordwalke:
> see class types as something different though and from my understanding it's not related.

You may wish to read up on the object system and class types. Polymoprhic variants and objects are actually very much related and it was a very enlightening learning experience to discover the connection.

PR #585, opened 2016-06-14
Add travis-ci config

tadeuzagallo:
Testing the configuration

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,16 @@
+language: objective-c
+osx_image: xcode7.3
+before_install:
+  - brew update
+  - brew install opam
+  - opam init
+  - opam switch 4.02.3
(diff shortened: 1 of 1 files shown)


tadeuzagallo:
It works!

/cc @jordwalke

yunxing:
Looks good, thanks!

yunxing:
@tadeuzagallo hmm... how does this work? Do you need admin access to add authorize travis to access this repo?

tadeuzagallo:
You need admin access to enable Travis for the repo. After that it should be fine. Is that what you mean?

yunxing:
@tadeuzagallo Right. Was just wondering where in fb's internal system can one gets admin access to the repo. But we can discuss that offline.

PR #596, opened 2016-06-19
Remove 'of' usage inside atom-reason.

SanderSpies:
The `npm install` step for atom-reason doesn't work at the moment, because of the `of` keyword usage. This PR removes the 'of' keyword.


yunxing:
Looks good to me. Thanks!

PR #598, opened 2016-06-20
[atom-reason] Add 'with-doc' 

SanderSpies:
Shows ocamldoc when using autocomplete within Nuclide. I'm not sure yet about the "Full description" part - but in general I think this is an improvement. 

(see also: https://github.com/the-lambda-church/merlin/issues/437#issuecomment-147189213)


SanderSpies:
Example: screen shot 2016-06-20 at 08 42 07

SanderSpies:
The current UI isn't optimal as we can't provide HTML for the description. Luckily there is a PR already to solve this: https://github.com/atom/autocomplete-plus/pull/733.

chenglou:
Cool! How's the perf of this?

SanderSpies:
I cite: 

```
The "with doc" variant also try to lookup OCamldoc, which is slightly more time consuming.
```

(https://github.com/the-lambda-church/merlin/blob/master/doc/dev/PROTOCOL.md).

I did some simple testing within the Reason codebase but did not notice any slow down.

chenglou:
Alright this is great, thanks. How come the artifacts aren't checked in here?

SanderSpies:
Which artifacts? 

Also note that #596 should be merged first.

chenglou:
The ones you get from running `npm start`. You did post the result so I presume you did run that? Do you have a local git ignore that excluded these artifacts? Also, last time you said the generated artifact looked really different from the current ones?

SanderSpies:
No, I just excluded them - I was under the impression that they would be generated anyway. 

My latest push includes the artifacts - and you can also see the naming issue (https://github.com/SanderSpies/reason/blob/efaca49eb5a7fb7f3f269328faf7c676f298dacf/editorSupport/atom-reason/output_byte_debug_js/app.js) that I mentioned before on IRC.

chenglou:
KP: this stretches the autocomplete window. I'm personally fine with this. If someone ever raises such issue we'll add the appropriate css overrides.

Sorry for delaying this so long. Merging!

PR #604, opened 2016-06-27
4.03.0

SanderSpies:
Work in progress - so do not merge obviously.

This builds upon the work already done by @kayceesrk.


SanderSpies:
The last formatTest issue that I'm facing is that `let _` isn't properly removed in certain situations when converting OCaml code to Reason code.

For example https://github.com/facebook/reason/blob/master/formatTest/typeCheckedTests/input/mlSyntax.ml#L48 should result into https://github.com/facebook/reason/blob/master/formatTest/typeCheckedTests/expected_output/mlSyntax.re#L63

SanderSpies:
So everything works (famous last words) - I still have to write some extra formatTests though for the new 4.03 features.

SanderSpies:
FormatTests to write for 4.03
---
- [x] PR#5528: inline records for constructor arguments
- [x] GPR#234: allow "[]" as a user-defined constructor.
- [ ] PR#6374: allow "_ t" as a short-hand for "(_, _, ..) t" for n-ary type
- [ ] PR#6806: Syntax shortcut for putting a type annotation on a record field
- [ ] PR#6806: Allow type annotations before the "->" in "fun <args> -> <expr>"
- [ ] GPR#26: support for "(type a b)" as syntactic sugar for "(type a) (type b)"
- [ ] GPR#42: short functor type syntax: "S -> T" for "functor (_ : S) -> T"
- [ ] GPR#88: allow field punning in object copying expressions

Taken from: http://caml.inria.fr/pub/distrib/ocaml-4.03/notes/Changes.

Please let me know if I miss something here.

jordwalke:
Regarding `let _ = expression` being converted into `expression;`. I would say that it seems *preferable* to preserve the `let _ =`. Or at least I'd say it doesn't matter much. I know the internal representation of top level imperative commands changed some time around `4.02.3-4.03`. What you're seeing might be that change. I'd say, all else being equal, preserving more is desirable and we should be thoughtful about the things which we don't preserve.

hhugo:
see `PR#6865: remove special case for "let _ = expr" structure items` https://github.com/ocaml/ocaml/commit/01bf671f33d7b20e1357f71ef064aaddba642992

yunxing:
@SanderSpies Are you still adding tests to this diff?

jordwalke:
So the plan for this is to either make `Reason` 4.03+ or two use the `4.03` AST, but then create a thin adapter that uses the `4.03` AST in the parser (for our own sanity, only having one canonical AST representation), but then converts it to the `4.02` representation for older compiler versions.
@cristianoc How is Infer support for 4.03 looking?

bobzhang:
FYI, we decide to target 4.04 (skip 4.03) for BuckleScript, some patches needed for Buckle are landed. 4.04 beta will be coming soon.

SanderSpies:
@yunxing not actively at the moment - I'm waiting on input here. 

I know the wish to convert to 4.02 but haven't put any effort in that yet. I believe others in OCaml ecosystem want to do something similar.

jordwalke:
Bob, so we will not be able to use BuckleScript with Reason unless we also target 4.04?

bobzhang:
It will take a while for us to upgrade for 4.04, so we will stay at 4.02.3 for at least a couple of months, but I think we will skip 4.03

bobzhang:
btw, there will be no bug fix release for 4.03

jordwalke:
I think everyone should be targeting the 4.04 AST then (not sure how many changes there were since 4.03). Then we can build support for previous versions on an as-needed basis.

cristianoc:
@jordwalke Infer could switch to 4.03 whenever there is support, or 4.04.
There are plans to switch to a newer version of Reason, as soon as comment formatting is stable. I originally thought about converting to 4.03 at the same time, but there's no strong motivation for doing that.

jordwalke:
Okay, so there's no issue with Infer's use of `Obj.magic` anymore?

jordwalke:
That's good to hear. I think we can get this diff in if that's the case!

cristianoc:
@jordwalke the issue with `Obj.magic` was actually in `joyson`, a depenency. There was a pull request to fix it some time ago, I should check but I guess it will have been pushed by now.

jordwalke:
Sander, do we need to do anything for PR 6374? Do we at least need to test that we can convert from ML to Reason?

SanderSpies:
Yes, that would be a wise idea - I'll continue this PR once JSX is merged.

SanderSpies:
Closing this in favour of the "OCaml lost in spacetime" edition.

PR #607, opened 2016-06-30
List `opam update` before `opam switch` in README

IvanVergiliev:
If you have an existing opam installation, 4.02.3 may not be an existing compiler. Running `opam update` before trying to switch to it fixes the problem.


yunxing:
Good point. Thanks for trying this out!

PR #610, opened 2016-07-04
Better error message for "invalid module item"

yunxing:
Instead of module item, this patch changes the term to "statement".

Also, add a special message for the case when we are sure that the semicolon is missing.

(Took me a while to merge the error messages for all the duplicated states. We should really have a way to automate it).


chenglou:
Furiously merging this

PR #614, opened 2016-07-05
Expose case_list in the printer.

### LINKED ISSUE #406: merlin-destruct generates ocaml even in reason files

yunxing:
This should help fix #406 

cc @let-def


yunxing:
Merging this. Should be harmless.

PR #615, opened 2016-07-05
[Prepare Repo for being an npm package]

jordwalke:
Summary:

- utop and many of its dependencies are not yet converted to an npm
package, so we make a `make` target without `rtop`/`utop` support.

- We need to therefore not export the `Reason_utop` module as part of
  the `Reason` library because that module won't even be built if
  building without `utop` support (such as when building npm from
  `Reason`.) No one is likely depending on that module being packaged in
  the Reason library - it's mainly just for building the `rtop` binary.

Is this the best way to do this? No, the best way would be to get utop + rtop installed as a result of `npm install reason`, but it will take some more time to port all of utop to a set of `npm` packages.

Reviewers: @chenglou @bsansouci @SanderSpies @yunxing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,12 @@ SHELL=bash -o pipefail
 
 default: build test
 
+build_without_utop: compile_error
+	cp pkg/META.in pkg/META
+	ocaml pkg/build.ml native=true native-dynlink=true utop=false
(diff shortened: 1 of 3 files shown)


jordwalke:
Tests pass so merging this very simple diff.

PR #618, opened 2016-07-05
Add .gitattributes

yunxing:
This is needed for #604 and would be needed in other places as well.


yunxing:
Merging this as it's harmless.

Issue #611, opened 2016-07-04
Lots of noise in .messages


SanderSpies:
This is bad for reviewing. 

For my local reviewing I'll probably use a search and replace regexp that removes all the state numbers with an empty string. This way I'll be able to actually see what changed, perhaps we could add something like this to the repo?

yunxing:
Hmm, the states are there for debugging purpose so I'm not sure if we can just remove them. 

Two things I think we can do are:
- Treat .messages as a binary file in git, so it won't show up in git show / diff.
- Have a separate projected file that is generated by filtering out all the state numbers in the original .messages file as you mentioned. So people can just review that file instead.

SanderSpies:
Yeah, the second item is exactly what we need.

PR #619, opened 2016-07-05
Introduce a simplified version of error messages database

yunxing:
Introduce another file, which is the simplified version of .messages that won't create too much noise for people to review when the grammar changes. 

This fixes the second part of #611.

@SanderSpies


SanderSpies:
Looks great!

SanderSpies:
Thanks!

PR #620, opened 2016-07-05
Support M.{} syntax

### LINKED ISSUE #132: M.{} is not parseable
`M.{}` is the syntax sugar way to write {let open M; {} }. 

Currently it is not parseable and throws a syntax error, we have to print it as `M.({})` instead.

yunxing:
This work is built on top of @dxu's work to fix #132


dxu:
lgtm. should it pretty print `M.{}` by default?

yunxing:
@dxu Good point. Let me change.

PR #621, opened 2016-07-06
Single line if statement.

SanderSpies:
This was supposed to be #569

But then my IDE decided to create the rebase I did not want... and I'm too lazy to research and correct it. 

Apologies for the inconvenience.

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/attributes.re
+++ b/formatTest/typeCheckedTests/expected_output/attributes.re
@@ -109,22 +109,10 @@ let add a => a [@onRet];
 
 let add = (fun a => a) [@onEntireFunction];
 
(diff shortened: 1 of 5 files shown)


yunxing:
@SanderSpies Thanks for working on this.

Diff looks good to me. Just have one question: Is the single line still going to break if it is too long?

SanderSpies:
It does break, but doesn't look pretty:
```
:118: if breakIfTooLong {
:119:   if somethingElse
:120:     {"some very very long text for testing"} else
:121:     {"another very very long text for testing"}
```

yunxing:
@SanderSpies Let's merge this. I think it the long single line case would be rare.

PR #622, opened 2016-07-06
Add support for `begin` and `end` as variable names

### LINKED ISSUE #603: Can't have a variable named `end`
Couldn't find anything in the docs about this.

image

yunxing:
Use the operator swapping infrastructure to provide support for `begin` and `end`.

The challenge here is make sure the programs written in Reason with `begin` or `end`
as variable names are still representable in ocaml.

The approach is:
- During parse time, whenever we see a `begin/end` in reason, we change it to "_begin_ / _end_" in the
  AST.

- During print time, whenever we see a `_begin_/_end_` in AST, we change it to "begin / end" in the
  reason.

In this way, reason programs with "begin/end" will be formatted as "_begin_/_end_" in ocaml.

The downside with this approach is that, in some rare cases, if one writes both "begin/end" and "_being_/_end_" in reason, both will be formatted to "begin/end" (since during printing time, we format "_begin_/_end_" to "begin/end") and the compilation will likely to fail.

This fixes #603

cc @jordwalke

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,0 +1,8 @@
+{
+  "name": "reason",
+  "dependencies": {
+    "@opam-alpha/camlp4": {
+      "version": "4.4.27"
+    }
+  }
+}


jordwalke:
I think the formatting of your comment is messed up a little bit.

jordwalke:
Making a small table to describe how identifiers are expressed:

<table>
<tr>
  <td>Identifier</td>
  <td>Expressed In Reason</td>
  <td>Expressed In OCaml</td>
</tr>
  <td>_begin_</td>
  <td>begin</td>
  <td>_begin_</td>
<tr>
  <td>_end_</td>
  <td>end</td>
  <td>_end_</td>
</tr>
</table>

cristianoc:
Technically, we don't have to give up on having the same AST-representable language, and a bijection between OCaml and Reason identifiers. E.g.
<table>
<tr>
  <td>Reason</td>
  <td>OCaml/AST</td>
</tr>
  <td>begin</td>
  <td>_begin</td>
<tr>
  <td>_begin</td>
  <td>__begin</td>
</tr>
<tr>
  <td>__begin</td>
  <td>___begin</td>
</tr>
<tr>
  <td>___begin</td>
  <td>____begin</td>
</tr>
<tr>
  <td>...</td>
  <td>...</td>
</tr>
</table>

yunxing:
@cristianoc Brilliant idea. Will change.

cristianoc:
The same issue happens in the opposite direction: OCaml programs that use `switch` as an identifier, while it is a keyword in Reason.

PR #623, opened 2016-07-06
Add syntax sugar for let z = M.(a, b)

yunxing:
@chenglou

### DIFF
--- a/formatTest/unit_tests/expected_output/modules.re
+++ b/formatTest/unit_tests/expected_output/modules.re
@@ -504,6 +504,9 @@ let module N = {
     34;
     35
   };
+  let z = M.(34, 35);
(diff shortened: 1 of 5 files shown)


chenglou:
Since this is constrained to module open, any chance to have some punning here? Not in this diff, but just as a general idea: `let (a, b) = MyModule`. Not too sure it's a good idea though...

Will leave @jordwalke to check this.

yunxing:
@chenglou hmm, the punning looks not very intuitive as we are eliding stuff that are too faraway from each other.

chenglou:
Yeah forget about that for now.

jordwalke:
Nice catch. I didn't know we forgot to support this.

PR #629, opened 2016-07-08
Add convenient symlinks to binaries

jordwalke:
Summary: This helps when we're not using an opam installation. The opam
.install file is what creates all the nice names such as `refmt`, but
since we won't always be using `opam` to bootstrap reason on `npm`, it's
nice to have some convenient renamed binary links.

Test Plan:

Reviewers: yunxing

CC:

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,17 @@
     "easy-format": "https://github.com/npm-ml/easy-format.git#npm-1.2.0",
     "merlin-extend": "https://github.com/npm-ml/merlin-extend.git#npm-0.0.1"
   },
+  "exportedEnvVars": {
+    "PATH": {
+      "val": "./_build/bin/",
+      "resolveAsRelativePath": true,
+      "global": true,
+      "globalCollisionBehavior": "joinPath"
+    }
+  },
   "scripts": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -21,7 +21,17 @@
...
+    "PATH": {
+      "val": "./_build/bin/",
+      "resolveAsRelativePath": true,
+      "global": true,
+      "globalCollisionBehavior": "joinPath"
+    }
+  },
   "scripts": {
+    "whereisocamlc": ". dependencyEnv && which ocamlc",
+    "whereisreason": ". dependencyEnv && which refmt",
```
jordwalke: This is a good example of how `dependencyEnv` works. Once my package (or my dependencies) specify `exportedEnvVars`, then `. dependencyEnv` will configure the env variables.

yunxing: When there is a collision on "PATH", does it join the local path to the beginning or to the end? If it is to the end, ocamlc and refmt will still use the globally installed one.

jordwalke: I believe it joins to the beginning.

yunxing:
Cool, looks good to me.

PR #643, opened 2016-07-15
New Comment Interleaving Logic

yunxing:
I'm working on a new comment interleaving logic that helps us:
1) Remove some of our previous location hacks in the parser, which break Merlin's assumption. 
2) Format comments in some cases correctly. For example:
This:
image 
will be formatted to 
image 
instead of 
image

Also this:
image
will be formatted to 
image

instead of 
image


jordwalke:
Excited about the improvements here, to the code and quality of output.

SanderSpies:
Also excited over here. Happy to see comments handling being improved!

SanderSpies:
Why closed?

yunxing:
mis-pressed :)

yunxing:
Some more changes (improvements?):
image
will be formatted as 
image
instead of 
image

yunxing:
Updated all the test cases. 

There are some regressions in aesthetics some are easy to fix while some are not. But for majority of changes it is an improvement.

formatTest/typeCheckedTests/expected_output/attributes.re
```diff
@@ -7,11 +8,11 @@
  * This has a nice side effect when printing the terms:
  * If a node has attributes attached to it,
  */
+
```
jordwalke: Was this intentionally changed? I had originally thought that all subsequent comments should have no spaces between them. Are you thinking that multiple subsequent single `*` comments should have no new lines but `**` comments should always have separating newlines?

formatTest/typeCheckedTests/expected_output/attributes.re
```diff
@@ -306,7 +309,6 @@ let myFun
 
 let myFun
     (X (hello [@onHello]) | Y (hello [@onHello])) => hello;
-
 /* Another bug: Cannot have an attribute on or pattern
-   let myFun = fun ((X hello | Y hello) [@onOrPattern]) => hello;
-   */
+let myFun = fun ((X hello | Y hello) [@onOrPattern]) => hello;
```
jordwalke: This looks like a regression, no?

formatTest/typeCheckedTests/expected_output/basics.re
```diff
@@ -28,7 +27,7 @@ let reasonDoubleBarNested =
   | _ => false;
 
 /* Liberal use of the Any pattern being compatible with multiple
-   arguments  */
+  arguments  */
```
jordwalke: And I think this looks like a regression, too, right?

formatTest/typeCheckedTests/expected_output/comments.re
```diff
@@ -21,43 +34,44 @@
...
+  */
 
+/**
+  **
+  */
 let testingNotQuiteEndOfLineComments = [
-  "Item 1", /* Comment For First Item */
-  "Item 2", /* Comment For Second Item */
-  "Item 3", /* Comment For Third Item */
+  "Item 1" /* Comment For First Item */,
```
jordwalke: This looks like a nice improvement.

formatTest/typeCheckedTests/expected_output/attributes.re
yunxing: Yeah, I was thinking the following use case:
```
/** The following family of functions does bar */

/** func1 does blah */
let func1 =  ...

/** func2 does blah */
let func2 = ...

/** func 3 does blah */
let func3 = ...

/** The following family of functions do foo */

/** func1 does blah */
let func1 =  ...

/** func2 does blah */
let func2 = ...

/** func 3 does blah */
let func3 = ...
```

Without the space between two doc comments, it would be formatted as : 
```
/** The following family of functions do bar */
/** func1 does blah */
let func1 =  ...

/** func2 does blah */
let func2 = ...

/** func 3 does blah */
let func3 = ...

/** The following family of functions does foo */
/** func1 does blah */
let func1 =  ...

/** func2 does blah */
let func2 = ...

/** func 3 does blah */
let func3 = ...
```

formatTest/typeCheckedTests/expected_output/attributes.re
yunxing: Interesting.

The input to this is:
image

I don't have any special logic yet to reformat multiline comments. I originally thought it's good to maintain the original format so it doesn't mess up with people drawing ascii arts in comments.

formatTest/typeCheckedTests/expected_output/attributes.re
jordwalke: Maybe we can allow subsequent single start comments to not have spaces.

```
/* single */
/* single */

/** double star comments always have a space above them */
/* single */
code...

/* single */
code...
```

formatTest/typeCheckedTests/expected_output/attributes.re
jordwalke: I had to put a lot of effort specifically into *not* maintaining original formatting. The reason is that when you reindent a large file, the original number of leading spaces needs to change to reflect the indentation.

formatTest/typeCheckedTests/expected_output/attributes.re
yunxing: yeah that's the current behavior I think.

formatTest/typeCheckedTests/expected_output/attributes.re
yunxing: Good point. Will change.

formatTest/typeCheckedTests/expected_output/comments.re
```diff
@@ -21,43 +34,44 @@
...
 
-/* Comment after list bracket */
 /* This time no space between bracket and comment */
-let testingEndOfLineComments = [];
+let testingEndOfLineComments =
+  [] /* Comment after list bracket */;
 
-/* Comment after list bracket */
-type t = (int, int); /* End of line on t */
+type t = (int, int /* End of line on t */);
```
yunxing: This is a regression which I can't get right. The input is 
image
and the algorithm is confused about where it should attach the comment to.

jordwalke: You should see what happens when in Reason you format:

```
type t = (int, int);  /* End of line on t */
```

In that case, is it preserved? There are more places for a comment to exist in Reason, and as long as each of them is preserved in Reason, it's pretty acceptable for converting from OCaml to behave this way.

However, the old code gave priority to the comment being on the entire type definition, which is why it was correctly formatted the way you might expect. The heuristic to preserving this is pretty simple. If the "attachment location" of the comment is an entire top level definition (includes the entire definition) and the comment is on the end of a line, then we render it as an end of line comment on the entire definition.

formatTest/typeCheckedTests/expected_output/comments.re
```diff
@@ -104,15 +114,15 @@ let result =
 
 type pointWithManyKindsOfComments = {
   /* Line before x */
-  x: string, /* x field */
+  x: string /* x field */,
```
jordwalke: Isn't this a regression?

formatTest/typeCheckedTests/expected_output/comments.re
```diff
@@ -121,10 +131,13 @@ type typeParamPointWithComments 'a = {
 let name_equal x y => x == y;
 
 let equal i1 i2 =>
-  i1.contents === i2.contents && true /* most unlikely first */;
+  i1.contents === i2.contents &&
+    true /* most unlikely first */;
```
jordwalke: I think this is a regression too. Because it is an end of line comment, and the "attachment location" totally includes an expression `x == y && z`, then the comment should be considered to be on that entire expression, while also preserving the fact that it's rendered on the end of the line.

When this happens, what we could do is maintain the end of line-ness, and then when rendering prevent *all* line breaks inside of the region that the end of line comment annotates. Even if we resize the text to be super skinny (`50` character width for example), the end of line comment will still appear on the exact same portion of the AST.

That would help us achieve the goal that we could reformat to a thin width, then resize to a wide width, and have all the comments in exactly the same place that they were originally.

yunxing:
@jordwalke 
The algorithm I have works like this:

1. It expands the comment's location 
2. It tries to find a "perfect" attachment of a comment to a AST node -- the comment's start location equals to the AST node's end location. If there are multiple AST nodes that can be perfectly attached to, it finds the deepest one.
3. If it fails to find a perfect attachment place, fall back to find a "good enough" attachment node. Instead of finding the deepest one, it finds the one as shallow in the tree as possible.

In Reason, EOL comment works fine because in the case of 
`type a = (int, int); /* comment */`
we have a semicolon and ")" to prevent the comment being "perfectly" attached to `int` and it will fall back to 3, which attach the comment to the entire type statement.

In ML, as you can see, this approach doesn't work quiet well because in the case of 
`type a = int * int /* comment */`, if we expand the start location of the comment, it will be "perfectly" attached to "int".

After tried different solutions to resolve this regression, I believe the best one is do not expand the start location of the comment in ML (it is unnecessary most of times anyway). We still need to do expand the end location of the comment though as people usually put comment on the right of an identifier, but that won't cause the issue.

I've updated the diff with the new test cases. The result looks pretty good and we can get goods from both worlds.

jordwalke:
> In Reason, EOL comment works fine because in the case of  `type a = (int, int); /* comment */` we have a semicolon and...

Will this change when/if we remove semicolons from Reason?

cristianoc:
@yunxing I'm very excited about this direction.

A couple of questions:

1. Once the attachment has been decided, would it be possible in principle to add this attachment information to the AST?
2. If there is ambiguity and the heuristics gives the result that the programmer had not planned, is there an escape route to change the attachment? Now, or in future.
3. Does the design have the property that the attachment is preserved by printing following by parsing, or is it something that the parser+pretty printed needs to carefully preserve?

yunxing:
@jordwalke 
Not in this case. 

If we remove semicolon, in the case of `type a = (int, int) /* comment */ `, the comment will be attached to the ast node `(int, int)` and will thus be printed as `type a = (int, int) /* comment */ `. 

However we will face a similar problem if we decide to add the semicolon back later and are converting the code from the version without semicolon to the version with semicolon: since the comment is attached to `(int, int)`, it would be formatted as `type a = (int, int) /* comment */ ;`

@cristianoc 

> Once the attachment has been decided, would it be possible in principle to add this attachment information to the AST?

I made a mistake in my previous comment -- The current algorithm is attaching comments to PrintTree node, instead of AST. 

I think it is possible to store the comment in the AST as attributes. The challenge would be how to store the relative position of the comment.
```
let a = {
   /* all the following comments will be attached to the node 1 */
   /* above */ 
  /* left*/ 1 /* right */
  /* below */
};
```

> If there is ambiguity and the heuristics gives the result that the programmer had not planned, is there an escape route to change the attachment? Now, or in future.

One general philosophy I have is to preserve the intention of the programmer as much as possible (more finer grained comments interleaving) to minimize the ambiguity of Reason -> Reason formatting. 

There could be a lot of ambiguity when converting ML to Reason but once converted to reason, we will give the programmer the options to arbitrarily resolve the ambiguity. For example, in the case of formatting `type a = int * int /* comment*/` to `type a = (int, int); /* comment*/`. If the programmer really meant to attach a comment after the last `int`, he would have to manually modify the code to `type a = (int, int /* comment */);`

> Does the design have the property that the attachment is preserved by printing following by parsing, or is it something that the parser+pretty printed needs to carefully preserve?

The attachment is preserved during printing time, no assumption for the parser.

yunxing:
Discussed offline with @jordwalke. There are is an interesting challenges we are facing:

The user may increase the print-width from a small value to a larger value, if we don't preserver end-of-line-ness, the result may look confusing:

image

In this example, it maybe **confusing** which value (2, or 3) the comment is decorating. 

The solution we had previously to force end-of-line comment to break so it always stay end-of-line no matter how big we set the print-width to: 
image. 


Another solution that I'm experimenting with is to always add a pair of parens to wrap the comment and the node. It will look like this when the width is small:
image
and look like this when we increase the width
image

bsansouci:
I like the paren idea. Especially if you can make Reason parse
```rust
let a = 
  f
    1
    2 /* we need a 2 here */
    3
```
and add the parens when formatting.

jordwalke:
Could you make it so that the parenthesis are only rendered when the end of line comment was rendered on the same line (no longer at the end of a line)?

I imagine that's hard to do with Easy_format right now (you can't predict whether something will end up breaking).

It seems the easier approach is to find the end of line comments, and then ensure that the contents of that line *never* break regardless of width. We can accomplish that with some kind of post-processing.

yunxing:
@jordwalke yeah, I think we can post-process the tree so content of eol never break regardless of width. But how does that help the issue here?

jordwalke:
Forcing the containing list to break solves the "growing" format width problem (so does mandating parens).

Forcing the item being end-of-line-commented on solves the shrinking problem in the event that there is no closing paren (or "simple"ifying end token).

```
type something =
   typeOne * typeTwo  (* End comment *)
```

In that example we want the end of line comment to be on the tuple type.
But if the format width shrinks to become small enough, we need to preserve the fact that the end of line comment applies to the entire tuple type. We cannot simply stick the end of line comment on the last item in the tuple, because it now appears to be attached to the `typeTwo`.


```
type something =
   typeOne *
   typeTwo  (* This isn't what we want *)
```


But if we have a formatting algorithm that post processes the layout tree, so that end of line comments make all commented on content never break upon formatting, then we've succeeded in preserving the end-of-line intent across various formatting widths.



```
type something =
   typeOne *  typeTwo  (* This isn't what we want to enforce *)
```

jordwalke:
@yunxing and I talked again today.

We seek to implement a convention for inferring the "attachment" of a comment, which is the AST node a comment applies to. This convention that we implement allows the user to convey an *intent* about which node a comment applies to. Here are some desirable features of the reformatter.

- You may reformat to any width, and every comment should still attach to what it previously attached to.
    ```
    forall comments c, widths w :
        attachment(c, source) === attachment(c, refmt(source, w))
    ```
- Temporarily formatting to smaller or larger widths, before finally formatting to the original width should have no effect on any comment position whatsoever.
```
forall widths w != currentWidth:
    refmt(source, currentWidth) === refmt(refmt(source, w), currentWidth)
```
This second invariant means that shrinking your width, and then expanding it back out should fully restore your file to its original content (likewise for growing and then shrinking it back).

- End of line comments should have a well defined attachment and that attachment should be preserved according to the first invariant, but also its *end-of-lineness* should also be preserved. Once on an end of line - always on an end of line.


### Some challenges

Here's some ways that an implementation might fail to meet the constraints above:

- Inline comments `let x = fun a /*a*/ b /*b */ => a` pose a challenge if an implementation decides to format them as end of line comments.
```
let x = fun a /*a*/ b /*b */ => a;
/* Shrinking to a smaller width gives: */
let x =
  fun
     a /*a*/
     b /*b */
    => a
```
Because of the third invariant above, these comments are now stuck as end of line comments - expanding the formatting to a wider width will not render them inline again. Therefore, the *second* invariant is now violated.

- Another issue with a potential implementation, is that the way it defines "attachment", and the way it reformats, can end up violating the first constraint - that reformatting should never change the "intent" of a comment - what it attaches to. If our implementation decides to attach end-of-line comments to any node that is fully contained on that same line - to its left, then line breaks can change the meaning.

```
type t =
   int * (int * int)  (* end of line *)
/* What if it was formatted to */
type t =
   int *
     (int * int)  (* end of line *)
```
In this example, because of how we defined attachment/intent for end-of-line comments, we've violated the first invariant. Reformatting has changed what a comment attaches to. We either need to fix the formatting algorithm to not violate the constraint, or we need to redefine what "attachment" means for end of line comments (so that it doesn't violate the invariants, and yet it still useful).

yunxing:
@jordwalke Thanks for the summary. 

We have two very vague ideas that we discussed offline to handle this two issues. 

For the first issue, we would need to classify all comments into three categories:
  - end of line comments
  - inline comments
  - single line comments
We think the key to resolve the conflicts between the second and the third category is to keep every comments' category unchanged after formatting. That is, a single line comment will remain as a single line comment, an inline comment will still be an inline comment after formatting:
   ```
   forall comment in comments:
      source.comment.category == refmt(source).comment.category
   ```
However, this variant could be extremely hard to maintain and has to be relaxed, as we may have code like this:
```
{
  x /* inline comment */;
};
```
when we format the code, according to the syntax, the trailing semicolon will be ignored:
```
{
  x /* inline comment */ <- this now becomes an end-of-line comment from inline comment
};
```

For the second issue, I think the right direction would be attaching the EOL comment to the AST node that's on the same line. A trick here would be to to post-process the tree so that everything that is "being attached" by a end of line comment would stay as a single line (Never_break).

As an example, if a user write something like this:
```
let a = { a; b; }
```
It will be formatted as the following code since we always break on statements
```
{
  a; 
  b
}
```

However, if a user attach a EOL comment to the same statement
```
{ a; b; } /* eol */
```
It would be formatted unchanged to preserve the "attachment" of the comment.
```
{ a; b;}  /* eol */
```

yunxing:
Rebased with master.

formatTest/typeCheckedTests/expected_output/arityConversion.re
```diff
@@ -7,7 +7,8 @@ type bcd =
 let a = TupleConstructor (1, 2);
 
 let b =
-  MultiArgumentsConstructor 1 2 [@implicit_arity];
```
jordwalke: This seemed to have nothing to do with comment interleaving. Do you know what happened? (Not a blocker)

formatTest/typeCheckedTests/expected_output/attributes.re
```diff
@@ -109,7 +111,8 @@ let add a => a [@onRet];
 
 let add = (fun a => a) [@onEntireFunction];
 
-let res = if true {false} else {false [@onFalse]};
```
jordwalke: Another case seemingly unrelated to comment interleaving.

formatTest/typeCheckedTests/expected_output/comments.re
```diff
@@ -21,38 +34,38 @@
 /* comment ***/
 /* comment ****/
 /* comment *****/
+
```
jordwalke: Yeah, this looks better more correct (I thought were were already doing this but apparently not).

formatTest/typeCheckedTests/expected_output/knownMlIssues.re
```diff
@@ -12,12 +11,11 @@ let blahCurriedX x =>
   fun
   | Red 10
   | Black 20
-  | Green 10 => 1
-  /* After or pattern green */
+  | Green 10 =>
+    1 /* After or pattern green */
```
jordwalke: Shouldn't this remain as an end of line comment on the entire case? This is not formatted as an end of line comment on the `1`, which changes the meaning. (The code before did not preserve correctly either, but I was under the impression that your latest approach preserves end of line comments, and their meaning).

formatTest/typeCheckedTests/expected_output/oo.re
```diff
@@ -30,28 +29,27 @@ let tmp = {
 /**
  * Comment on stackWithAttributes.
  */
```
jordwalke: The input was:

```
/**
 * Comment on stackWithAttributes.
 */
class virtual stackWithAttributes 'a init =>
/* Before class */
{
```

Is the change intensional? In the new version, it looks like the comment is on the `as`, but in the original, it looked like it was on the entire class expression `{}`.

formatTest/typeCheckedTests/expected_output/reasonComments.re
```diff
@@ -1,29 +1,39 @@
 3; /* - */
 
-3;
+3; /*-*/
```
jordwalke: Is this PR updated with the latest change that includes the semicolon in the location of the top level binding/eval? If so, I would have expected the comment to be preserved to the left of the semi.

formatTest/typeCheckedTests/expected_output/reasonComments.re
```diff
@@ -61,36 +73,31 @@ let testingEndOfLineComments = [
...
-]
-/* Comment after rbracket */;
+] /* Comment after rbracket */;
 
 /* The space between ; and comment shoudn't matter */
 let testPlacementOfTrailingComment = [
   "Item 0" /* */
   /* Comment after last item in list. */
-];
+]; /* Comment after semi */
```
jordwalke: It's tough for me to recall the approach/algorithm for how this is preserved as an end of line comment, in such a way that the meaning is clear and preserved across reformats.
I know we discussed and you might not have implemented a solution, but documenting it would be a good idea.

```
let x =
   somethingHere (
   ); /* Comment */
```

What is the comment on, the let binding or expression? How is that meaning preserved when there is enough space to fit it on one line:

```
let x = somethingHere ();   /* Comment */
```

Now it is entirely on the let binding, even if we said the original meaning was that the comment applied to the function application only.

formatTest/unit_tests/expected_output/modules.re
```diff
@@ -194,13 +196,9 @@ module type ASig = {let a: int;};
 
 module type BSig = {let b: int;};
 
-let module AMod = {
-  let a = 10;
-};
+let module AMod = {let a = 10;};
```
jordwalke: I thought we force broke these? Did something change?

formatTest/typeCheckedTests/expected_output/arityConversion.re
yunxing: Yeah, the original code was at width 50 but didn't break, it is a bug so I fixed it.

formatTest/typeCheckedTests/expected_output/oo.re
yunxing: This is intentional. 

I just feel the position of "{" doesn't look good in this case:
```
class virtual stackWithAttributes 'a init =>
/* Before class */
{
```

formatTest/typeCheckedTests/expected_output/knownMlIssues.re
yunxing: It should. We don't have the loc information for the case so the algorithm doesn't know where to attach to. 

The algorithm does try to attach the comment to the entire case group (from |RED to 1), but figures out that the entire group is not on the same line. 
```
  | Red 10
  | Black 20
  | Green 10 -> 1 (* After or pattern green *)
```

formatTest/typeCheckedTests/expected_output/reasonComments.re
yunxing: Good point. Probably a bug on my end. Will fix.

yunxing: I think the root cause of this is we don't have a location information for the atom "3". I will address this in another diff.

formatTest/unit_tests/expected_output/wrappingTest.re
```diff
@@ -2297,8 +2294,10 @@ let anotherRecord = {
 };
 
 type polymorphicCommentedType
-  /* Commenting the first type variable */'a
-  /* Commenting the second type variable */'b =
+  /* Commenting the first type variable */
```
jordwalke: Looks like an improvement - single line comments are preserved correctly.

formatTest/unit_tests/expected_output/wrappingTest.re
```diff
@@ -2351,26 +2351,23 @@ module type ASig = {let a: int;};
...
 module type BSig = {let b: int;};
 
-let module AMod = {
-  let a = 10;
+let module AMod = {let a = 10;};
+
+let module BMod = {let b = 10;};
+
+let module CurriedSugar
+           /* Commenting before First curried functor arg */
```
jordwalke: These look like fixes! Nice.

formatTest/unit_tests/expected_output/wrappingTest.re
```diff
@@ -2466,169 +2444,136 @@ let produceRecord /*commentBeforeArg*/ x => {
...
+let myPolyFuncCommentAfterColon: 'a .'a => 'a =
   /*afterColon */
-  'a => 'a
- =
   fun o => o;
 
 let myPolyFuncCommentBeforeArrow:
   'a .
-  'a => /*beforeArrow */  'a
+  'a /*beforeArrow */ => 'a
```
jordwalke: This is really that "perfect attachment" logic (Step 2) in the algorithm working its magic.

src/reason_parser.mly
```diff
@@ -3079,31 +3079,22 @@ _or_pattern:
...
 
-
-/**
- * Makes the bar part of the pattern location. This makes it much easier to
- * interleave comments. Normally in switch/fun| each case doesn't have a
- * position, but the left end of the pattern and right end of the expression is
- * used as the location. We need to make the left location of the pattern match
- * what end of line comments appear to be attaching to.
- */
-bar_located_pattern: BAR pattern  {
```
jordwalke: I'm curious how you were able to preserve end of line comments on entire cases when you removed the hack. Come to think of it, I recall some examples where the eol comment wasn't preserved correctly on entire match cases. What can we do to fix it?

formatTest/typeCheckedTests/expected_output/knownMlIssues.re
jordwalke: Ah, I see that makes sense. I see the challenge in this case - no pun intended!

formatTest/typeCheckedTests/expected_output/oo.re
jordwalke: I agree it doesn't look as good, but if we don't like the look, then we should have written the single line comment inside the braces. (I understand that's not an option for converting from ML).

Did you have to include special logic to introduce the new behavior? I think the old behavior was better because you could choose which of the two styles you wanted, and then the formatter would preserve it.

formatTest/typeCheckedTests/expected_output/reasonComments.re
jordwalke: Ah, sounds good - surprised we don't have loc info for the atom yet. Follow up for this sounds like the best approach.

src/reason_pprint_ast.ml
```diff
@@ -1296,303 +1335,488 @@ let convertIsListyToIsSequencey isListyImpl =
...
-  extractComments comments (
-    fun comAttachFirst comAttachLast comPhysFirst comPhysLast ->
-      let entirelyPhysicallyAbove = comPhysFirst < firstChar && comPhysLast < firstChar in
-      entirelyPhysicallyAbove
-  )
+let beforeLoc loc1 loc2 =
+  loc1.loc_end.Lexing.pos_cnum <= loc2.loc_start.Lexing.pos_cnum
 
+let rec attachEOLComment layout txt =
+  WithComment (txt, layout)
```
jordwalke: I'm trying to find all the uses of `WithComment`, and I'm only able to find ones that model end of line comments. Should `WithComment` be called `WithEndOfLineComment`?

src/reason_pprint_ast.ml
```diff
@@ -1066,6 +1101,15 @@ let easyLabelForEolComment labelTerm term =
   } in
   Easy_format.Label ((labelTerm, settings), term)
 
+let autoBreakLabel labelTerm term =
+  let settings = {
+    label_break = `Never;
```
jordwalke: auto break label, has a setting of `Never` break?

src/reason_parser.mly
yunxing: Yes that's true. 

For single "|" case, we already have a SourceMap that includes the bar, so we are able to attach to the entire match case.

src/reason_pprint_ast.ml
yunxing: Good find

src/reason_pprint_ast.ml
yunxing: Good point

formatTest/typeCheckedTests/expected_output/reasonComments.re
yunxing: All comments after semicolon are perfectly attached to the entire let statement (Step 2) and it naturally becomes end of line comment. 

I like the idea of formatting it to the top of statement and I think it is making more sense the more I think about it:
```
  /* Comment after semi */
 let testPlacementOfTrailingComment = [
   "Item 0" /* */
   /* Comment after last item in list. */
];
```

formatTest/typeCheckedTests/expected_output/reasonComments.re
yunxing: Yeah. We don't have loc info for a lot of atoms as far as I can remember.

formatTest/typeCheckedTests/expected_output/arityConversion.re
jordwalke: Great.

formatTest/typeCheckedTests/expected_output/oo.re
jordwalke: Is it difficult to add back the ability to have either form?

yunxing: Just saw this comment. Yes I had to add special logic for the new behavior. I can revert it and it will also make the code simpler.

formatTest/typeCheckedTests/expected_output/reasonComments.re
jordwalke: I would suggest doing that (formatting to the top of the entire statement). It's useful to have a distinction between perfectly end-of-line comments, and imperfect ones like this.
Imperfect:
```rust
let x = (
   "first",
   "second"
) /* imperfect eol comment on tuple */
;
```
Perfect
```rust
let x =
  ("first", "second") /* perfect eol comment on tuple */
;
```


Maybe this is another category of comment called "PartialEolComment" which would be normalized to be above the item when printing. In my previous attempts, I made the mistake of not making the distinction between imperfect and perfect eol comments. That causes much difficulty.

```rust
type x =
   Variant (
   really *
   long *
   tupleType
   ...
) /* Comment */
;
```

In that example, it's an imperfect eol comment is on the tuple portion of the variant type. If we just treated it as a regular eol comment, then we have a problem, because the only way to ensure that eol comments remain as eol comments across different width formattings, is to force `~break:Never` on the entire AST item (and force break on the containing list/label). That's a problem because these imperfect eol comments are typically on very very long structures, so it will cause massively long unbroken line. The solution seems to be to consider these as different "imperfect eol comments", and to format the comment above the AST node instead.

formatTest/typeCheckedTests/expected_output/oo.re
yunxing: @jordwalke When I see it again I still like this form better? Another solution is to add an empty line after the comment:
```
class virtual stackWithAttributes 'a init => {
/* Before class */

/* The "as this" should not be formatted away because attributes. */
....
```

formatTest/typeCheckedTests/expected_output/attributes.re
yunxing: Right. The width was 50 before and it is now breaking correctly.

src/reason_pprint_ast.ml
```diff
@@ -5006,7 +5320,7 @@ class printer  ()= object(self:'self)
         )
     | Pmod_structure (s) ->
       makeList
-        ~break:IfNeed
+        ~break:Always_rec
```
yunxing: @jordwalke Don't know what's happening here. But I have to force break this for the case of 
```
let module AMod = {
  let a = 10;
};
```

formatTest/typeCheckedTests/expected_output/oo.re
jordwalke: I would recommend taking out the special case and making the code simpler to evolve/iterate on.
Curious, with this current special case, how does one distinguish a comment on the first method, vs. on the entire object returned?

Issue #625, opened 2016-07-08
[atom-reason] Add merlin-destruct


chenglou:
To automatically generate the `switch` cases based on e.g. the variant.

SanderSpies:
Working on this one.

SanderSpies:
Close please 

cc @chenglou

PR #644, opened 2016-07-16
Add merlin destruct support to Atom Reason

SanderSpies:
Mostly works when you press `alt+p`.

Except that the pattern match isn't converted to Reason yet - it's not clear yet to me where this should happen. I also don't see it properly being converted for type information (of keyword is still there...).


SanderSpies:
Related to #625.

SanderSpies:
Needs more checks, breaks too easy currently.

SanderSpies:
To clarify a bit more what this does, if you type the following

```
type cursor =
  | Default
  | Pointer
;

let foo bar:cursor =>
  bar
;
```

and press `alt + p` when the cursor is at `bar`, `bar` will be replaced by:
```
match (bar : cursor) with
| Default  -> failwith "TODO"
| Pointer  -> failwith "TODO"
```

(Yes, the match should be a switch - but not sure where this is being handled).

SanderSpies:
Is `OcamlTypeToReasonType` actually being used somewhere? I can delete it and everything builds just fine... also searching finds no usages.

chenglou:
It's supposed to be used by BetterErrors before to crawl the ocaml type errors and turn them into reason type errors.

chenglou:
Just tried this. Pretty cool, except yeah, it prints the ocaml syntax rather than the Reason one in a reason file. cc @let-def, given that #676 got implemented without any merlin change, could it be a merlin problem here?

jordwalke:
Whoa!

chenglou:
"https://github.com/facebook/reason/blob/master/src/ocamlmerlin_reason.ml doesn't have pretty print hook, so it cannot work with latest merlin-extend".

More info: https://github.com/let-def/merlin-extend/commit/1e885ee0a624c9464073e727e4274e0c26e1aa8a

Leaving this here to @SanderSpies.

jordwalke:
I'm going to try adding that pretty print hook in Reason right now. If I succeed, then this diff will work well, right?

let-def:
I think so, if codebases didn't diverge too much. I can fix issues in the next few hours.

SanderSpies:
Rebased this branch to the latest master version

chenglou:
Tested locally. Works well! @jordwalke you should try it! =D

PR #647, opened 2016-07-16
[WIP] JSX support

SanderSpies:
DO NOT MERGE

What I want to get working: 

```
let selfClosing = <Foo />;
let selfClosing = <Foo a={1} b={true} />;
let a = <Foo> <Bar c={fun a => a + 2} /> </Foo>;
let text = <Yo>"string"</Yo>;
```

which should be translated into something like:

```
let selfClosing = Foo.createElement []
let selfClosing = Foo.createElement a::1 b::true [];
let a = Foo.createElement [Bar.createElement c::(fun a => a + 2) []];
let text = Yo.createElement ["string"];
```

Although this PR is far from finished (I've only tested the feasibility in the parser) I'd like to get the discussion started. 

Current limitations: 
- using quoted strings
- spaced needed between > and <

What do you guys think of adding JSX support like this?

### DIFF
--- a/src/reason_lexer.mll
+++ b/src/reason_lexer.mll
@@ -456,6 +456,7 @@ rule token = parse
   | "&&" { AMPERAMPER }
   | "`"  { BACKQUOTE }
   | "'"  { QUOTE }
+  | "</" { LESSSLASH }
   | "("  { LPAREN }
   | ")"  { RPAREN }
   | "*"  { STAR }
@@ -469,6 +470,7 @@ rule token = parse
   | ":"  { COLON }
   | "::" { COLONCOLON }
   | ":=" { COLONEQUAL }
+  | "/>" { SLASHGREATER }
   | ":>" { COLONGREATER }
   | ";"  { SEMI }
   | ";;" { SEMISEMI }
--- a/src/reason_parser.mly
+++ b/src/reason_parser.mly
@@ -865,6 +865,7 @@ let built_in_explicit_arity_constructors = ["Some"; "Assert_failure"; "Match_fai
 %token LBRACKETAT
 %token LBRACKETATAT
 %token LBRACKETATATAT
+%token LESSSLASH
 %token SWITCH
 %token MATCH
 %token METHOD
@@ -898,6 +899,7 @@ let built_in_explicit_arity_constructors = ["Some"; "Assert_failure"; "Match_fai
 %token SEMISEMI
 %token SHARP
 %token SIG
+%token SLASHGREATER
 %token STAR
 %token <string * string option> STRING
 %token STRUCT
@@ -2491,6 +2493,21 @@ labeled_simple_pattern:
       { ("", None, $1) }
 ;
 
+jsx_arguments:
+    /* empty */ { [] }
+    | LIDENT EQUAL LBRACE expr RBRACE jsx_arguments { [1] }
+;
+
+jsx:
+  | LESS UIDENT jsx_arguments SLASHGREATER {
+     mkexp(Pexp_variant($2, None))
+  }
+  | LESS UIDENT jsx_arguments GREATER LESSSLASH UIDENT GREATER {
+       mkexp(Pexp_variant($2, None))
+    }
+  | LESS UIDENT jsx_arguments GREATER jsx LESSSLASH UIDENT GREATER {
+     mkexp(Pexp_variant($2, None))
+  }
 /*
  * Much like how patterns are partitioned into pattern/simple_pattern,
  * expressions are divided into expr/simple_expr.
@@ -2514,6 +2531,7 @@ _expr:
       let (l,o,p) = $2 in
       mkexp (Pexp_fun(l, o, p, $3))
     }
+  | jsx { $1 }
   | FUN LPAREN TYPE LIDENT RPAREN fun_def
       { mkexp (Pexp_newtype($4, $6)) }
   /* List style rules like this often need a special precendence
--- a/formatTest/unit_tests/input/jsx.re
+++ b/formatTest/unit_tests/input/jsx.re
@@ -1,0 +1,3 @@
+let selfClosing = <Foo />;
+let selfClosing = <Foo a={1} b={true} />;
+let a = <Foo> <Bar c={fun a => a + 2} /> </Foo>;
--- a/src/reason_parser.messages
+++ b/src/reason_parser.messages
@@ -1,6 +1,6 @@
 use_file: SHARP LIDENT SEMI WITH
 ##
-## Ends in an error in state: 2751.
+## Ends in an error in state: 2784.
 ##
 ## _use_file -> toplevel_directive SEMI . use_file [ # ]
 ##
@@ -12,7 +12,7 @@ use_file: SHARP LIDENT SEMI WITH
 
 use_file: SHARP LIDENT TRUE WITH
 ##
-## Ends in an error in state: 2750.
+## Ends in an error in state: 2783.
 ##
 ## _use_file -> toplevel_directive . SEMI use_file [ # ]
 ## _use_file -> toplevel_directive . EOF [ # ]
@@ -25,7 +25,7 @@ use_file: SHARP LIDENT TRUE WITH
 
 use_file: UIDENT RPAREN
 ##
-## Ends in an error in state: 2753.
+## Ends in an error in state: 2786.
 ##
 ## _use_file -> structure_item . SEMI use_file [ # ]
 ## _use_file -> structure_item . EOF [ # ]
@@ -37,28 +37,28 @@ use_file: UIDENT RPAREN
 ## This implies that, although the LR(1) items shown above provide an
 ## accurate view of the past (what has been recognized so far), they
 ## may provide an INCOMPLETE view of the future (what was expected next).
-## In state 794, spurious reduction of production constr_longident -> mod_longident
-## In state 951, spurious reduction of production _simple_expr -> constr_longident
-## In state 917, spurious reduction of production mark_position_exp(_simple_expr) -> _simple_expr
-## In state 913, spurious reduction of production simple_expr -> mark_position_exp(_simple_expr)
-## In state 921, spurious reduction of production less_aggressive_simple_expression -> simple_expr
-## In state 931, spurious reduction of production _expr -> less_aggressive_simple_expression
-## In state 960, spurious reduction of production mark_position_exp(_expr) -> _expr
-## In state 930, spurious reduction of production expr -> mark_position_exp(_expr)
-## In state 2742, spurious reduction of production post_item_attributes ->
-## In state 2743, spurious reduction of production _structure_item_without_item_extension_sugar -> expr post_item_attributes
-## In state 2744, spurious reduction of production mark_position_str(_structure_item_without_item_extension_sugar) -> _structure_item_without_item_extension_sugar
-## In state 2722, spurious reduction of production structure_item_without_item_extension_sugar -> mark_position_str(_structure_item_without_item_extension_sugar)
-## In state 2716, spurious reduction of production _structure_item -> structure_item_without_item_extension_sugar
-## In state 2746, spurious reduction of production mark_position_str(_structure_item) -> _structure_item
-## In state 2723, spurious reduction of production structure_item -> mark_position_str(_structure_item)
+## In state 746, spurious reduction of production constr_longident -> mod_longident
+## In state 960, spurious reduction of production _simple_expr -> constr_longident
+## In state 904, spurious reduction of production mark_position_exp(_simple_expr) -> _simple_expr
+## In state 900, spurious reduction of production simple_expr -> mark_position_exp(_simple_expr)
+## In state 929, spurious reduction of production less_aggressive_simple_expression -> simple_expr
+## In state 939, spurious reduction of production _expr -> less_aggressive_simple_expression
+## In state 969, spurious reduction of production mark_position_exp(_expr) -> _expr
+## In state 938, spurious reduction of production expr -> mark_position_exp(_expr)
+## In state 2775, spurious reduction of production post_item_attributes ->
+## In state 2776, spurious reduction of production _structure_item_without_item_extension_sugar -> expr post_item_attributes
+## In state 2777, spurious reduction of production mark_position_str(_structure_item_without_item_extension_sugar) -> _structure_item_without_item_extension_sugar
+## In state 2755, spurious reduction of production structure_item_without_item_extension_sugar -> mark_position_str(_structure_item_without_item_extension_sugar)
+## In state 2749, spurious reduction of production _structure_item -> structure_item_without_item_extension_sugar
+## In state 2779, spurious reduction of production mark_position_str(_structure_item) -> _structure_item
+## In state 2756, spurious reduction of production structure_item -> mark_position_str(_structure_item)
 ##
 
 <SYNTAX ERROR>
 
 use_file: UIDENT SEMI WITH
 ##
-## Ends in an error in state: 2754.
+## Ends in an error in state: 2787.
 ##
 ## _use_file -> structure_item SEMI . use_file [ # ]
 ##
@@ -70,7 +70,7 @@ use_file: UIDENT SEMI WITH
 
 use_file: WITH
 ##
-## Ends in an error in state: 2747.
+## Ends in an error in state: 2780.
 ##
 ## use_file' -> . use_file [ # ]
 ##
@@ -82,7 +82,7 @@ use_file: WITH
 
 toplevel_phrase: SHARP UIDENT EOF
 ##
-## Ends in an error in state: 2714.
+## Ends in an error in state: 2747.
 ##
 ## _toplevel_phrase -> toplevel_directive . SEMI [ # ]
 ##
@@ -93,14 +93,14 @@ toplevel_phrase: SHARP UIDENT EOF
 ## This implies that, although the LR(1) items shown above provide an
 ## accurate view of the past (what has been recognized so far), they
 ## may provide an INCOMPLETE view of the future (what was expected next).
-## In state 2614, spurious reduction of production toplevel_directive -> SHARP ident
+## In state 2647, spurious reduction of production toplevel_directive -> SHARP ident
 ##
 
 <SYNTAX ERROR>
 
 toplevel_phrase: SHARP UIDENT UIDENT DOT WITH
 ##
-## Ends in an error in state: 2621.
+## Ends in an error in state: 2654.
 ##
 ## mod_longident -> mod_longident DOT . UIDENT [ SEMI EOF DOT ]
 ## val_longident -> mod_longident DOT . val_ident [ SEMI EOF ]
@@ -113,7 +113,7 @@ toplevel_phrase: SHARP UIDENT UIDENT DOT WITH
 
 toplevel_phrase: SHARP UIDENT UIDENT WITH
 ##
-## Ends in an error in state: 2620.
+## Ends in an error in state: 2653.
 ##
 ## mod_longident -> mod_longident . DOT UIDENT [ SEMI EOF DOT ]
 ## toplevel_directive -> SHARP ident mod_longident . [ SEMI EOF ]
@@ -127,7 +127,7 @@ toplevel_phrase: SHARP UIDENT UIDENT WITH
 
 toplevel_phrase: SHARP UIDENT WITH
 ##
-## Ends in an error in state: 2614.
+## Ends in an error in state: 2647.
 ##
 ## toplevel_directive -> SHARP ident . [ SEMI EOF ]
 ## toplevel_directive -> SHARP ident . STRING [ SEMI EOF ]
@@ -145,7 +145,7 @@ toplevel_phrase: SHARP UIDENT WITH
 
 toplevel_phrase: SHARP WITH
 ##
-## Ends in an error in state: 2613.
+## Ends in an error in state: 2646.
 ##
 ## toplevel_directive -> SHARP . ident [ SEMI EOF ]
 ## toplevel_directive -> SHARP . ident STRING [ SEMI EOF ]
@@ -163,7 +163,7 @@ toplevel_phrase: SHARP WITH
 
 toplevel_phrase: UIDENT RPAREN
 ##
-## Ends in an error in state: 2717.
+## Ends in an error in state: 2750.
 ##
 ## _toplevel_phrase -> structure_item . SEMI [ # ]
 ##
@@ -174,28 +174,28 @@ toplevel_phrase: UIDENT RPAREN
 ## This implies that, although the LR(1) items shown above provide an
 ## accurate view of the past (what has been recognized so far), they
 ## may provide an INCOMPLETE view of the future (what was expected next).
-## In state 794, spurious reduction of production constr_longident -> mod_longident
-## In state 951, spurious reduction of production _simple_expr -> constr_longident
-## In state 917, spurious reduction of production mark_position_exp(_simple_expr) -> _simple_expr
-## In state 913, spurious reduction of production simple_expr -> mark_position_exp(_simple_expr)
-## In state 921, spurious reduction of production less_aggressive_simple_expression -> simple_expr
-## In state 931, spurious reduction of production _expr -> less_aggressive_simple_expression
-## In state 960, spurious reduction of production mark_position_exp(_expr) -> _expr
-## In state 930, spurious reduction of production expr -> mark_position_exp(_expr)
-## In state 2742, spurious reduction of production post_item_attributes ->
-## In state 2743, spurious reduction of production _structure_item_without_item_extension_sugar -> expr post_item_attributes
-## In state 2744, spurious reduction of production mark_position_str(_structure_item_without_item_extension_sugar) -> _structure_item_without_item_extension_sugar
-## In state 2722, spurious reduction of production structure_item_without_item_extension_sugar -> mark_position_str(_structure_item_without_item_extension_sugar)
-## In state 2716, spurious reduction of production _structure_item -> structure_item_without_item_extension_sugar
-## In state 2746, spurious reduction of production mark_position_str(_structure_item) -> _structure_item
-## In state 2723, spurious reduction of production structure_item -> mark_position_str(_structure_item)
+## In state 746, spurious reduction of production constr_longident -> mod_longident
+## In state 960, spurious reduction of production _simple_expr -> constr_longident
+## In state 904, spurious reduction of production mark_position_exp(_simple_expr) -> _simple_expr
+## In state 900, spurious reduction of production simple_expr -> mark_position_exp(_simple_expr)
+## In state 929, spurious reduction of production less_aggressive_simple_expression -> simple_expr
+## In state 939, spurious reduction of production _expr -> less_aggressive_simple_expression
+## In state 969, spurious reduction of production mark_position_exp(_expr) -> _expr
+## In state 938, spurious reduction of production expr -> mark_position_exp(_expr)
+## In state 2775, spurious reduction of production post_item_attributes ->
+## In state 2776, spurious reduction of production _structure_item_without_item_extension_sugar -> expr post_item_attributes
+## In state 2777, spurious reduction of production mark_position_str(_structure_item_without_item_extension_sugar) -> _structure_item_without_item_extension_sugar
+## In state 2755, spurious reduction of production structure_item_without_item_extension_sugar -> mark_position_str(_structure_item_without_item_extension_sugar)
+## In state 2749, spurious reduction of production _structure_item -> structure_item_without_item_extension_sugar
+## In state 2779, spurious reduction of production mark_position_str(_structure_item) -> _structure_item
+## In state 2756, spurious reduction of production structure_item -> mark_position_str(_structure_item)
 ##
 
 <SYNTAX ERROR>
 
 toplevel_phrase: WITH
 ##
-## Ends in an error in state: 2576.
+## Ends in an error in state: 2609.
 ##
 ## toplevel_phrase' -> . toplevel_phrase [ # ]
 ##
@@ -207,7 +207,7 @@ toplevel_phrase: WITH
 
 parse_pattern: EXCEPTION UNDERSCORE WITH
 ##
-## Ends in an error in state: 644.
+## Ends in an error in state: 658.
 ##
 ## _pattern_without_or -> pattern_without_or . AS val_ident [ WHEN SEMI RPAREN RBRACKET LBRACKETAT IN EQUALGREATER EQUAL EOF COMMA COLONCOLON COLON BAR AS ]
 ## _pattern_without_or -> pattern_without_or . AS error [ WHEN SEMI RPAREN RBRACKET LBRACKETAT IN EQUALGREATER EQUAL EOF COMMA COLONCOLON COLON BAR AS ]
@@ -224,7 +224,7 @@ parse_pattern: EXCEPTION UNDERSCORE WITH
 
 parse_pattern: EXCEPTION WITH
 ##
-## Ends in an error in state: 642.
+## Ends in an error in state: 656.
 ##
 ## _pattern_without_or -> EXCEPTION . pattern_without_or [ WHEN SEMI RPAREN RBRACKET LBRACKETAT IN EQUALGREATER EQUAL EOF COMMA COLONCOLON COLON BAR AS ]
 ##
@@ -236,7 +236,7 @@ parse_pattern: EXCEPTION WITH
 
 parse_pattern: LAZY WITH
 ##
-## Ends in an error in state: 619.
+## Ends in an error in state: 639.
 ##
 ## _pattern_without_or -> LAZY . simple_pattern [ WHEN SEMI RPAREN RBRACKET LBRACKETAT IN EQUALGREATER EQUAL EOF COMMA COLONCOLON COLON BAR AS ]
 ##
@@ -248,7 +248,7 @@ parse_pattern: LAZY WITH
 
 parse_pattern: LBRACE LIDENT COLON UNDERSCORE SEMI
 ##
-## Ends in an error in state: 699.
+## Ends in an error in state: 713.
 ##
 ## _or_pattern -> pattern . BAR pattern [ error RBRACE COMMA BAR ]
 ## lbl_pattern -> label_longident COLON pattern . [ error RBRACE COMMA ]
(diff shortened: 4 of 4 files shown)


bsansouci:
Wow this is amazing. Thanks for doing this!

SanderSpies:
Update: this should be fairly simple to finish.

Syntax-wise I now have support for:
```
let selfClosing = <Foo />;
let selfClosing = <Foo a={1} b={true} />;
let a = <Foo> <Bar c={fun a => a + 2} /> </Foo>;
let a = <So> <Much> <Nesting> </Nesting> </Much> </So>;
let a = <Sibling> <One test={true}/> <Two foo={b}> </Two> </Sibling>;
let a = <Foo>"testing a string here"</Foo>;
let a = <Foo>"testing a string here" <Test /> "another string" <Bar> </Bar>{ 2 + 4 }</Foo>;
```

Wiring everything up is also straightforward, I just need to figure out how to perform the nesting of children within the AST.

SanderSpies:
Ready for review.

yunxing:
woah this is sick!

jordwalke:
One thing I always wanted in JSX, was the ability to put any simple expression to the right of the `=`.

```rust
let x = <Component count=10 point={x: 0, y:0} callBack=(fun x => x) />
```

jordwalke:
I'm trying to think of ways that this could conflict with custom infix operators such as:

```rust
let ( /> ) a b => a + b;
```

If so, we'd need to provide operator swapping support for the set of operators that conflict.

chenglou:
Big request while this is still in its early phase: prop punning please.
`<Foo prop1 prop2>`
This'll prevent early flux-ification where devs start pulling props out into stores for the sake of looking at less props.

yunxing:
How do we swap "createElement" with something else? Do we ever have the need?

spicyj:
I personally see the primary advantage of JSX being the matching closing tags; the other syntax differences don't seem as valuable to me.

So I wonder if we can find something with that benefit but without special complicated syntax rules that JSX introduces (ex: whitespace parsing, entity escaping, similar spread operator but no shorthand property support). I see your PR here already is much more minimalist than JSX which I like.

chenglou:
@spicyj is it possible for us to eliminate the needs of jsx if we auto insert comments for "closing tags"?

```rust
normalFunc prop1::value 
  (normalFunc2 prop2::value) // "self-closing"
// normalFunc
```

spicyj:
Interesting. Maybe!

jordwalke:
I pitched that idea when first building support for JSX in JS and for some reason people didn't get as excited about it as true JSX support.

chenglou:
Yeah but times have changed and people like functions more than HTML now! (jk no idea what I'm saying).

The good thing about using comments is that it provides a nice down(up?)grade path, where people who don't like matching closing tags can simply remove the comment. Also this would be editor-level rather than syntax level. But I'm sure you've heard all of these before.

SanderSpies:
@chenglou - Punning support has been added.

@jordwalke Not sure how feasible omitting `{` and `}` for attribute values is as the syntax will then conflict with things like `exp GREATER exp`.

SanderSpies:
That was actually a lot easier then I thought, replacing `expr` with `simple_expr` did the trick.

bsansouci:
@SanderSpies Wow you're a magician.

SanderSpies:
Spread operator is out, as it's too troublesome to add.

SanderSpies:
- I've added namespace support, eg. `<Namespace.Foo.Bar />`. 
- Spaces are now optional between tags
- Ensured that the operators that worked before in Reason still work.

jordwalke:
> Spaces are now optional between tags

Awesome!

SanderSpies:
I also like to support lowercase tags, simply because writing `<div>` is more pleasing to my eyes then `<Div>`. This currently doesn't work because OCaml's module identifier is capitalized. I was wondering if we could make this work in some smart way. So far unsuccessful though, so open for suggestions.

jordwalke:
Lowercase could trivially be made into function calls, but come time to format, there's no way to know which lowercase functions to render as JSX and which ones not to. You could always parse (and therefore preserve at render time), some small attribute on those function application nodes which preserve its JSX-ness.

src/reason_parser.mly
```diff
@@ -788,8 +788,17 @@ let rec string_of_longident = function
 
 let built_in_explicit_arity_constructors = ["Some"; "Assert_failure"; "Match_failure"]
 
+let jsx_component module_name attrs children ~loc =
```
jordwalke: Wondering if we should call it `RSX` since we've already deviated a little bit from `JSX` (like allowing any simple expression for attributes etc).

src/reason_parser.mly
```diff
@@ -788,8 +788,17 @@ let rec string_of_longident = function
...
+    let ident = ghloc ~loc (Ldot(module_name, "createElement")) in
+    mkexp(Pexp_apply(mkexp(Pexp_ident ident), attrs @ children))
+
+let ensureTagsAreEqual startTag endTag =
+  if startTag <> endTag then
+    syntax_error ()
```
jordwalke: We definitely want a beautiful error message here, right?

src/reason_parser.mly
jordwalke: I suppose we can say we have "JSX For Reason" which is slightly different than JSX.

src/reason_parser.mly
```diff
@@ -1049,6 +1064,8 @@ conflicts.
 
 %nonassoc below_LBRACKETAT
 %nonassoc LBRACKETAT
+%nonassoc SLASHGREATERLESS
+%nonassoc GREATERLESS
```
jordwalke: Do you understand why these precedences for these tokens must be exactly as they are specified here?

src/reason_parser.mly
```diff
@@ -2491,6 +2508,121 @@ labeled_simple_pattern:
...
+     (jsx_component $1 $2, $1)
+  }
+;
+
+jsx_separate_open_close:
+  | jsx_start_tag_body GREATERLESSSLASH mod_longident {
+    (* Foo></Foo *)
+      let (component, start) = $1 in
+      let loc = mklocation $symbolstartpos $endpos in
+      let _ = ensureTagsAreEqual start $3 in
```
jordwalke: Btw: When we are in "recovering parser mode", we recover from parsing failures, by replacing whole nodes with extension point nodes (which have a single string expression describing the error). In that mode, it's good because it doesn't block parsing the rest of the file, and then merlin can work its type checking recovery magic.

src/reason_parser.mly
```diff
@@ -4095,12 +4228,16 @@ val_ident:
   | PERCENT                                     { "%" }
   | LESSGREATER                                 { "<>" }
   | LESSDOTDOTGREATER                           { "<..>" }
+  | SLASHGREATERLESS                            { "/><" }
```
jordwalke: Where is it explained why we need this special token? What errors occur in the parser if we try to keep only the tokens `/>` and `<`.

src/syntax_util.ml
```diff
@@ -72,6 +72,7 @@ let reason_to_ml_swapping_alist = [
   "\\===", "===";
   "!=", "<>";
   "!==", "!=";
+  "\\/><", "/><";
```
jordwalke: Must the same thing be done for these?
```
 "/>" { SLASHGREATER }
 "/></" { SLASHGREATERLESSSLASH }

```

src/reason_parser.mly
SanderSpies: Will explain in an upcoming commit, reason is that we need it for removing spaces between tags.

src/syntax_util.ml
SanderSpies: I tested the operators and the other two already failed without the PR, will re-check.

jordwalke:
```rust
let ( /> ) = fun a b => a + b;
```

src/reason_parser.mly
jordwalke: Yeah, I recall that and I think it should be documented somewhere why <X /><Y> would not parse while the parse rules are in terms of tokens `<` `X` `/>` `<` `Y` `>`.

src/syntax_util.ml
SanderSpies: Ah yes, you are right.

src/reason_parser.mly
SanderSpies: It's needed to solve the 74 shift/reduce I get otherwise for turning these into infix_operators. Not sure if this is correct, and I wonder if I should go for %left or %right instead. In other words: I'm lacking knowledge/experience here.

src/syntax_util.ml
SanderSpies: This has been fixed.

SanderSpies:
Is there any good reason why we should allow `<Foo></Foo>` instead of `<Foo />`? I think we can win some consistency here by not allowing tags with empty content.

jordwalke:
> Is there any good reason why we should allow <Foo></Foo> besides <Foo />? I think we can win some consistency here by not allowing tags with empty content.


Does it make the implementation easier?


Also, looking forward to the pretty printing of JSX.

SanderSpies:
Yes, it will make the implementation a little bit easier as I can remove some stuff. But it should mostly be done for providing a better developer experience. I think that it's easier to read a self-closing tag then an empty tag, but others could have a different opinion.

Working on pretty printing, hope to have something soonish.

jordwalke:
> "Yes, it will make the implementation a little bit easier as I can remove some stuff. But it should mostly be done for providing a better developer experience. I think that it's easier to read a self-closing tag then an empty tag, but others could have a different opinion."

If it's purely about the developer experience, then you can have the pretty printer format it as the preferred form, yet accept the tags as is. (It might help with copy/pasting example code from the internet etc).

Drup:
Not wanting to sound gloomy or anything, but why not just use ppx for that ... ? 

The syntax may not be as lightweight, but it has the upside of not messing up even more with operators and exported symbols.

jordwalke:
@Drup

> "The syntax may not be as lightweight"

I think that would be the main reason to not use ppx for this task. ppx was never meant to be a way to manipulate concrete syntax, which is really the level that JSX is operating at. Also, people are just familiar with JSX and really like it for some reason. I was never a huge fan, but it grew on me and now I understand the appeal.

We can go forward with JSX, and start by hiding under a flag that you can enable. (Perhaps we could use a ppx attribute at the top of the file to enable it?)

```
[@@@jsx];
```

SanderSpies:
The current JSX detection is not good, I want to move this to a JSX ppx attribute.

SanderSpies:
OTOH it might be better to keep the current check and expand it a bit  - moving to PPX would require an extra build step to remove the JSX ppx.

jordwalke:
ppx "extensions" must be removed before compiling. That requires an extra build step. However, ppx "attributes" `[@jsx]` would *not* require a build step and do not have to be removed before compiling. They can act as an annotation, preserving the fact that a block was parsed with JSX.

SanderSpies:
@jordwalke - that's good news, I will move to ppx attributes instead.

jordwalke:
If you use `ppx` attributes, then one other good piece falls out of this. We no longer need to tie ourselves to `ModuleName.createElement`. Instead, any function can be invoked with `JSX`. We would merely preserve that non-essential fact in a ppx attributes.


If we stick with named arguments, then it's perfect.

The drawback would be if we ever go to records as props. In that case, I'd probably want:

```rust
let x =
   <ModuleName
      prop1=foo
   />
```

To be parsed as:

```rust
   ModuleName.createElement {
      ...ModuleName.defaultProps,
      prop1:foo
   }
```

But you can imagine that requires type driven transformation to determine if the component even has default properties.

formatTest/unit_tests/expected_output/jsx.re
```diff
@@ -0,0 +1,180 @@
...
+let jsxInFnCall = testFunc (<Foo />);
+
+let lotsOfArguments =
+  <LotsOfArguments
+     argument1=1
+     argument2=2
+     argument3=3
+     argument4=4
+     argument5=5
+     argument6="test"
```
jordwalke: I think the common convention is to include the final `>` on the last attribute. It can simply be appended to the last item in the argument list.

formatTest/unit_tests/expected_output/jsx.re
```diff
@@ -0,0 +1,180 @@
...
+  <So><Much><Nesting /></Much></So>;
+
+let spaceBefore2 = <So><Much /></So>;
+
+let siblingNotSpaced = <So><Much /><Much /></So>;
+
+/* doesn't work (pretty print messes it up): let jsxInArray = [(<Foo />)]; */
+let testFunc b => b;
+
+let jsxInFnCall = testFunc (<Foo />);
```
jordwalke: The parens are required?

formatTest/unit_tests/expected_output/jsx.re
```diff
@@ -0,0 +1,180 @@
...
+let d = <Foo />;
+
+let spaceBefore =
+  <So><Much><Nesting /></Much></So>;
+
+let spaceBefore2 = <So><Much /></So>;
+
+let siblingNotSpaced = <So><Much /><Much /></So>;
+
+/* doesn't work (pretty print messes it up): let jsxInArray = [(<Foo />)]; */
```
jordwalke: I think this one is important to solve. With new React changes coming, we'll be able to return arrays of React components from the `render` function, can you believe it!?

formatTest/unit_tests/expected_output/jsx.re
SanderSpies: Yes, otherwise it results in a syntax error at the moment.

jordwalke: Can we just consider the JSX block to be a "simple expression" - as if `<Foo` was the opening paren, and `</Foo>` or `/>` was the closing paren?

SanderSpies: Not easily, this results in shift/reduce errors.

SanderSpies:
Review please!

src/reason_pprint_ast.ml
```diff
@@ -691,6 +691,16 @@ let default = new Pprintast.printer ()
 type funcReturnStyle =
   | ReturnValOnSameLine
 
+let rec detectJSXComponent e attributes =
+  match (e, attributes) with
+    | (Pexp_ident loc, ({txt = "JSX"; loc = _}, PStr []) :: tail) ->
+        let moduleNameList = List.rev (List.tl (List.rev (Longident.flatten loc.txt))) in
+        if List.length moduleNameList > 0 then
```
yunxing: Why do you need the "if" here?

Can't you just use
```
 Some (String.concat "." moduleNameList) 
```
?

src/reason_pprint_ast.ml
```diff
@@ -3303,11 +3313,18 @@ class printer  ()= object(self:'self)
...
-                      [SourceMap (e.pexp_loc, (self#expression2 e))];
-                      List.map self#reset#label_x_expression_param l;
-                      attributesAsList;
-                    ],
+                    let exp = match detectJSXComponent e.pexp_desc x.pexp_attributes with
+                      | Some componentName -> [
+                        [self#formatJSXComponent componentName l];
+                       ]
+                      | None ->
+                      [
```
yunxing: Maybe use "| None -> [" for consistency with the case above?

src/reason_parser.mly
```diff
@@ -2490,6 +2520,125 @@ labeled_simple_pattern:
   | simple_pattern
       { ("", None, $1) }
 ;
+jsx_arguments:
+    /* empty */ { [] }
```
yunxing: Maybe keep 2 spaces indentations here for consistency

formatTest/unit_tests/expected_output/jsx.re
```diff
@@ -0,0 +1,208 @@
...
+
+let a4 =
+  <Sibling>
+    <One test=true foo=b />
+    <Two foo=b />
+  </Sibling>;
+
+let a5 = <Foo>"testing a string here"</Foo>;
+
+let a6 =
```
yunxing: Nice formatting!

src/reason_parser.mly
```diff
@@ -2490,6 +2520,125 @@ labeled_simple_pattern:
...
+    | LIDENT jsx_arguments {
+        (* a (punning) *)
+        let loc_lident = mklocation $startpos($1) $endpos($1) in
+        [($1, mkexp (Pexp_ident {txt = Lident $1; loc = loc_lident}))] @ $2
+     }
+;
+
+jsx_name:
+| mod_longident { $1 }
+| LIDENT { Lident $1 }
```
yunxing: Missing something here: in which case can a jsx name becomes LIDENT?

yunxing: Do you support imported functions like 
<Mod.foo > ?

src/reason_parser.mly
```diff
@@ -2490,6 +2520,125 @@ labeled_simple_pattern:
...
+      let _ = ensureTagsAreEqual start $3 loc in
+      component [("", mktailexp_extension loc [] None)] ~loc
+  }
+  | jsx_start_tag_body GREATERLESS jsx_tag_siblings jsx_name {
+      let (component, start) = $1 in
+      let loc = mklocation $symbolstartpos $endpos in
+      let _ = ensureTagsAreEqual start $4 loc in
+      component [("", mktailexp_extension loc $3 None)] ~loc
+  }
+  | jsx_start_tag_body GREATER LESS jsx_tag_siblings jsx_name {
```
yunxing: Maybe you can do something like : 
great_less: 
  | GREATER LESS {}
  | GREATERLESS {}

to unify these two rules.

src/reason_parser.mly
```diff
@@ -2490,6 +2520,125 @@ labeled_simple_pattern:
...
+      let (component, start) = $1 in
+      let loc = mklocation $symbolstartpos $endpos in
+      let _ = ensureTagsAreEqual start $5 loc in
+      component [("", mktailexp_extension loc $4 None)] ~loc
+    }
+  | jsx_start_tag_body GREATERLESS jsx_tag_siblings LESSSLASH jsx_name
+    (* Foo><Tags /> </Foo *)
+  | jsx_start_tag_body GREATER jsx_siblings LESSSLASH jsx_name {
```
yunxing: what about these case? 
```
Foo><Tags /></Foo
```
and 
```
Foo> <Tags /></Foo
```

yunxing: nvm, it should be covered by the case above.

src/reason_parser.mly
```diff
@@ -2490,6 +2520,125 @@ labeled_simple_pattern:
...
+      let _ = ensureTagsAreEqual start $4 loc in
+      component [("", mktailexp_extension loc $3 None)] ~loc
+  }
+  | jsx_start_tag_body GREATER LESS jsx_tag_siblings jsx_name {
+      let (component, start) = $1 in
+      let loc = mklocation $symbolstartpos $endpos in
+      let _ = ensureTagsAreEqual start $5 loc in
+      component [("", mktailexp_extension loc $4 None)] ~loc
+    }
+  | jsx_start_tag_body GREATERLESS jsx_tag_siblings LESSSLASH jsx_name
```
yunxing: Does this mean the following incorrect syntax is also supported ? 
```
Foo><Tags /></ </Foo
```
?

src/reason_parser.mly
SanderSpies: No, not right now - is there a need for this? (could probably be easily added)

src/reason_pprint_ast.ml
SanderSpies: You are right, will change.

src/reason_parser.mly
SanderSpies: Apparently yes, I will try to fix it.

src/reason_pprint_ast.ml
SanderSpies: No actually not, I use separate parts of the identifier for both blocks. First part uses everything but the `createElement` part (so the last list item) while the second part uses the last list item which can only be one (which would be a function call).

src/reason_parser.mly
SanderSpies: It will complain about not being to reduce `GREATER LESS` and `GREATERLESS`. Not sure how I can work around that.

yunxing: Ah, you can just use "%inline" keyword (search for %inline here: http://gallium.inria.fr/~fpottier/menhir/manual.pdf)

src/reason_parser.mly
```diff
@@ -788,8 +788,28 @@ let rec string_of_longident = function
...
 
+let jsx_component module_name attrs children ~loc =
+  let firstPart = (List.hd (Longident.flatten module_name)) in
+  let lident = if firstPart = String.capitalize firstPart then
+    Ldot(module_name, "createElement")
+  else
+    Lident firstPart
+  in
+  let ident = ghloc ~loc lident in
+  let body = mkexp(Pexp_apply(mkexp(Pexp_ident ident), attrs @ children)) in
```
yunxing: You also want to pass loc here when you call mkexp (Pexp_ident ident). Otherwise, you will generate an AST node without location information and our IDE support will get confused. 

Currently, with your diff, formatting a code of: 
```
let a = <Foo a />;
```
gives us
```
[
  structure_item (try.re[1,0+0]..[1,0+18])
    Pstr_value Nonrec
    [
      <def>
        pattern (try.re[1,0+4]..[1,0+5])
          Ppat_var "a" (try.re[1,0+4]..[1,0+5])
        expression (try.re[1,0+8]..[1,0+17])
          attribute "JSX"
            []
          Pexp_apply
          expression ([0,0+-1]..[0,0+-1])
            Pexp_ident "Foo.createElement" (try.re[1,0+8]..[1,0+17]) ghost
          [
            <label> "a"
              expression ([0,0+-1]..[0,0+-1])
                Pexp_ident "a" (try.re[1,0+13]..[1,0+14])
            <label> ""
              expression (try.re[1,0+8]..[1,0+17]) ghost
                Pexp_construct "[]" (try.re[1,0+8]..[1,0+17]) ghost
                None
          ]
    ]
]
```

(You can check the generated AST by:
```
refmt -parse re -print ast try.re
```
)

src/reason_parser.mly
```diff
@@ -2490,6 +2520,140 @@ labeled_simple_pattern:
...
+        [$1] @ $3
+     }
+;
+
+jsx_siblings:
+    LESSSLASH { [] }
+  | STRING jsx_siblings
+      {
+        let (s, d) = $1 in
+        [mkexp (Pexp_constant (Const_string (s, d)))] @ $2
```
yunxing: you also need to pass loc here to mkexp

src/reason_parser.mly
```diff
@@ -2490,6 +2520,140 @@ labeled_simple_pattern:
...
+      }
+  | LESS jsx_tag_siblings
+      {
+        $2
+      }
+  | LBRACE expr RBRACE jsx_siblings
```
yunxing: Something is incorrect here.
If you try `let a = <Foo a>{1}</Foo>;`, you will notice this is not idempotent.

src/reason_pprint_ast.ml
```diff
@@ -691,6 +691,16 @@ let default = new Pprintast.printer ()
 type funcReturnStyle =
   | ReturnValOnSameLine
 
+let rec detectJSXComponent e attributes =
+  match (e, attributes) with
+    | (Pexp_ident loc, ({txt = "JSX"; loc = _}, PStr []) :: tail) ->
```
yunxing: You may also want to make sure the txt part of the "loc" has a final component of "createElement". Otherwise `Foo.bar [] [@JSX]` will also be formatted as `<Foo />;`

src/reason_pprint_ast.ml
```diff
@@ -3303,11 +3313,17 @@ class printer  ()= object(self:'self)
                   | None -> (
                     (* Standard application *)
                     (*reset here only because [function,match,try,sequence] are lower priority*)
-                    List.concat [
-                      [SourceMap (e.pexp_loc, (self#expression2 e))];
-                      List.map self#reset#label_x_expression_param l;
-                      attributesAsList;
-                    ],
+                    let exp = match detectJSXComponent e.pexp_desc x.pexp_attributes with
```
yunxing: You may also want to detect if "l" here could be a valid JSX children (at least a list). Otherwise `Foo.createElement "" [@JSX]` will be formatted as `<Foo ="" />;`, which is incorrect.

yunxing:
Mostly looking awesome! Except for some small issues to make the code more defensive.

src/reason_parser.mly
SanderSpies: This has been fixed in the printer.

SanderSpies:
@yunxing I've processed all of your feedback

chenglou:
There are still some debates on whether this should desugar to a function call with labels (might not work for Reason React, even less so for interop). But that's easy to change. I say we should just merge this soon.

yunxing:
Thanks @SanderSpies. The diff looks pretty good to me. One thing I don't like is the new tokens added, but I don't know if we can remove them.

@jordwalke feel free to merge it.

jordwalke:
It would be good to move the test case to the type checked folder. That catches more bugs when converting or escaping operators.

SanderSpies:
Sure, I'll do that. @jordwalke also mentioned some concerns about polymorphic variants which I will investigate.

jordwalke:
The only concern is just that they are well tested and the new token for `[<` doesn't mess up polymorphic variant types which can be defined like:

```rust
type atMost  = [< `Foo int | `Bar int];
type atLeast = [> `Foo int | `Bar int]
```

SanderSpies:
@jordwalke also wants to add support for fragments - which would like th E4X version:
``` <>
  <div />
  <span />
</>
```

SanderSpies:
@jordwalke polymorphic variant types are covered good enough in the current set of tests from what I see.

SanderSpies:
Remaining work:
- [ ]  add tests for added tokens to ensure they can be used as operators if needed
- [ ]  improve fragment support (currently only tags)

SanderSpies:
Check again please - esp. the added support for fragments.

src/reason_lexer.mll
```diff
@@ -510,6 +517,7 @@ rule token = parse
   | "-"  { MINUS }
   | "-." { MINUSDOT }
   | "<>" { LESSGREATER }
+  | "<><" { LESSGREATERLESS }
```
jordwalke: I think there's no test cases for this infix operator.

src/reason_pprint_ast.ml
```diff
@@ -4341,7 +4431,7 @@ class printer  ()= object(self:'self)
       | Pexp_apply _ -> true
       | _ -> false
 
-  method simple_expression x =
+  method simple_expression ?(hasJSXAttribute = false) x  =
     let (arityAttrs, docAtrs, stdAttrs) = partitionAttributes x.pexp_attributes in
```
jordwalke: Some thoughts I had while trying to rebase on master.

- whether or not this has jsx attributes would have already been evident in the attributes, had they not been stripped before calling simple_expression.

src/reason_pprint_ast.ml
```diff
@@ -4017,11 +4094,21 @@ class printer  ()= object(self:'self)
   method expression x =
     let (arityAttrs, docAtrs, stdAttrs) = partitionAttributes x.pexp_attributes in
```
jordwalke: I think this all might be cleaner if `partitionAttributes` also returned any `jsx` attributes it found. I've thought of `stdAttrs` as anything printed as `[@]`, but in the diff's current form, JSX gets added to `stdAttrs`.

src/reason_lexer.mll
SanderSpies: I tried this on the master branch at the time and it didn't work. Will try again.

src/reason_parser.mly
```diff
@@ -2491,6 +2522,136 @@ labeled_simple_pattern:
...
+jsx_separate_open_close:
+  | jsx_start_tag_body GREATERLESSSLASH jsx_name
+      {
+        (* Foo></Foo *)
+        let (component, start) = $1 in
+        let loc = mklocation $symbolstartpos $endpos in
+        let _ = ensureTagsAreEqual start $3 loc in
+        component [("", mktailexp_extension loc [] None)] ~loc
+      }
+  | jsx_start_tag_body GREATERLESS jsx_tag_siblings jsx_name
```
jordwalke: This seems to allow for there to be a space between `<` and the tag name. `< TagName`. I wonder if things become simpler if the lexer requires that there be no space.

jordwalke: We could even forbid using `<` or `>` without spaces around it: require `let x = 0 < 1` instead of allowing `let x = 0 <1`.

src/reason_pprint_ast.ml
```diff
@@ -4017,11 +4094,21 @@ class printer  ()= object(self:'self)
...
+          | _ :: tail -> findJSXAttribute tail
+          | [] -> false
+      in
+      let hasJSXAttribute = findJSXAttribute x.pexp_attributes in
       let withoutVisibleAttrs = {x with pexp_attributes=arityAttrs} in
         let itm =
-          formatAttributed
-            (self#simple_expression withoutVisibleAttrs)
-            (self#attributes stdAttrs)
+          if hasJSXAttribute then
```
jordwalke: I think this is supposed to say `if theOnlyStdAttrsWereJsxAttrs then` (if it had `stdAttrs` but all of them were `JSX` attrs).
Otherwise, as it is, I think sometimes attributes would get dropped from JSX expressions.

jordwalke:
Okay, I've got the printing working on top of the latest master (rebased).
There's some issues with how it's parsed though, and would require a bunch of hacky token splitting work. I'd rather not do the token splitting work if it's not necessary because the whole motivation for token splitting will go away once object types have `{}` syntax, so I don't want to add more use cases for token splitting when not needed. I think there's a way.
As it stands, with @SanderSpies' latest "simplification" of jsx, things like `<X><Y>` do not parse. (It fails on the `<Y`).

Here's what I believe the solution is: we need lexer rules that generate tokens for the following cases:
```
# One token
><Tag
# Two tokens
> <Tag

# One token
></Tag
# Two tokens
> </Tag

# One token
/></Tag
# Two tokens
/> </Tag

# One token
/><Tag
# Two tokens
/> <Tag
```

Where you see two tokens above one of those tokens is often one of these individual tokens:
```
</Tag
<Tag
/>
```

`/>` is the only token which *might* cause shift reduce conflicts with infix `/>` token, but I don't think it will. I think we can reuse the same token as `/>` infix without conflicts (let me know if I'm wrong). If it does cause unsolvable conflicts, then we'd have to express the infix `/>` as `\/>` but I hope it doesn't come to that.

The following should parse correctly as well:

```
   simple + simple
   <t /> + <t />
   <t> </t > + <t> </t >
```
and
```
   simple simple simple
   myFunc <t /> <t />
```

Constraints:
------------
In order for this to work, without ruining the natural ability for people to use `>` and `<` without being confused for JSX tags, the following must be true:


- Closing tags must be written as `</t>` not `</t >` (I think)
- Opening tags must be written as `<tag` and not `< tag`
- Infix `<` must be written as `x < y` not `x <y`
- Infix  `><`   must be written as `x >< y` not `x ><y`
- Infix `/></` must be written as `x /></ y` not `x/></y`.
- Infix `/><` must be written as `x />< y` not `x/></y`.

and so on. Those constraints above don't require any additional work in the parser/lexer/printer - they're just things the user must keep in mind.

If this is done in the parser, then the printer that I just rebased on master should work and we can merge this PR.

vjeux:
> @jordwalke One thing I always wanted in JSX, was the ability to put any simple expression to the right of the =.

> ```js
let x = <Component count=10 point={x: 0, y:0} callBack=(fun x => x) />
```

I think that this is a mistake to support this. I understand the desire to fix JSX in order to make it what you wished it was and to use Reason as a good forcing function to do so. But this is going to yet another small difference between JavaScript and Reason that's going to be in the way of people trying to read/write both languages, for no obviously good benefit than a nice to have.

I'm not part of the project so feel free to disregard my opinion, but I think that there's a higher chance of it being more successful if it's seen as a "better JavaScript" than a "friendlier Ocaml".

If you want to run with my theory, then you should strive to make the syntax a good subset of JavaScript and only deviate when there's a strong and clear reason to do so. This is the pattern we used for React Native where we tried as much as possible to make it a good subset of the web.

Right now it seems like there's a desire to make it look like JavaScript but in practice, it's a weird mix of ocaml and js but the code doesn't look anything like actual JavaScript.

Anyway, I'm happy to chat about it, but my 2cents :)

bobzhang:
> but I think that there's a higher chance of it being more successful if it's seen as a "better JavaScript" than a "friendlier Ocaml".

100% agree

jordwalke:
@vjeux It's definitely a tough balance to strike and it might take a couple of attempts to find the sweet spot. We definitely want to keep moving `Reason` towards JS, the good parts. But to provide a counter-argument, there can be a positive effect to having some syntactic differences and a little bit of unfamiliarity. Remember when we released React with `JSX`, how it immediately filtered out a ton of people who were quick to dismiss something new? I think in retrospect that was the best possible outcome, even if accidental. To the observer, it almost looked like `React` flopped - and we can thank JSX for that. But the truth is that there was a great community being seeded. That's not to say we should make things worse or foreign just for the sake of it. But the truth is that if people want the benefits that sound static typing brings (safety + performance (optimized compilation)), they're going to have to learn some stuff. There's just no way around it. It can't be obtained via `JavaScript` as we know it today - `JavaScript` would have to be changed into something it's not and make the programmer learn many new concepts anyways.

So maybe it's the case that having some differences in syntax is a good way to manage expectations that there will be things to learn in order to reap those benefits. It could have the same beneficial filtering effect that JSX for React had. Thoughts?

But if you have some suggestions even beyond JSX, I'd love to hear them. A few people have even requested that parenthesis be mandatory around function arguments (we're also open to that suggestion as it makes named argument syntax much better - but opinions may vary).


But for JSX, why not go the other direction? Why not work towards JSX supporting simple expressions to the right of the `=` sign?
Also, punning is really really great and we shouldn't neglect it just because JSX doesn't support it (if I could go back in time, I would have added punning to JSX from day one - instead an attribute with no `=value` is transpiled as `attribute=true`. JS itself didn't yet have punning - now it does, so maybe it's a good time to reevaluate).

jordwalke:
Oh, and one other thing @vjeux. There's actually room for a couple different "layers" here. I think with what we have for `Reason` right now, even though it's got some inspiration from JS (the good parts), I really do believe it is a more *friendly* way to use OCaml since we've fixed the top ten reported syntactic pitfalls. The benefit of this mode is that it can perfectly be converted back and forth between OCaml and Reason (in theory). All function styles interop perfectly between the two worlds.

But if you're willing to give that up, there's all kinds of syntactic transforms we could do on top, even going as far as to recreate things from JS that we really believe are worse. These transforms would not be "recoverable"(you can't go back to plain Reason from them) and that would be totally okay. There's an entire spectrum we can/should explore, even if some of those solutions (that very much resemble JS) have performance drawbacks - because at the end of the day we maintain the most important thing - we have a *way* to write parts of our in something lower that performs better in a way that interops perfectly with all the other stuff that targets the same allocator.

vjeux:
I was doing some archeology around React and you ( @jordwalke ) wrote the following in the initial diff that brought JSX to React.

> Making the sugar sweeter:
> - I'd like to omit the need for braces in as many places as possible. Not sure how difficult that
would be.

So you were indeed write saying "One thing I always wanted in JSX, was the ability to put any simple expression to the right of the =." :)

PR #653, opened 2016-07-26
Use built in variable for make

kassens:
Fixes this warning:

>  Command 'make' called directly, use the built-in variable instead

Based this fix on what I saw in https://github.com/AltGr/Camelus/issues/2

I'm not sure how to test this locally, would appreciate if someone could double check this is good.


yunxing:
Tested locally by 
```
opam upgrade reason
```
and
```
>opam lint
>opam:Passed.
```
can confirm this works. Thanks!

PR #655, opened 2016-07-30
Switch all string regex functions to use ocaml-re

rickyvetter:
This removes the c regex dependency that str uses and allows pretty printer
to be compiled by js_of_ocaml and run in the browser.

I am pretty far from an expert in ocaml build systems, so I sort of just ended up adding "re.str" to every meta file that looked relevant. I may have over-applied or missed some places. If so, please let me know and I'll update accordingly!

### DIFF
--- a/opam
+++ b/opam
@@ -26,6 +26,7 @@ depends: [
   "utop"         {>= "1.17"}
   "BetterErrors" {>= "0.0.1"}
   "menhir"       {>= "20160303"}
+  "re"           {>= "1.5.0"}
 ]
 depopts: [
 ]
--- a/.merlin
+++ b/.merlin
@@ -1,3 +1,3 @@
-PKG utop str merlin_extend compiler-libs easy-format
+PKG utop str merlin_extend compiler-libs easy-format re.str
 B _build/src
 S src
--- a/_tags
+++ b/_tags
@@ -4,8 +4,8 @@ true: warn(@5@8@10@11@12@14@23-24@26@29@40), bin_annot, safe_string, debug
 <node_modules/**>: -traverse
 "formatTest": -traverse
 "src": include
-<src/*>: package(menhirLib unix compiler-libs.common ocamlbuild findlib easy-format str BetterErrors merlin_extend)
-<src/reason_utop.*>: package(menhirLib utop str)
+<src/*>: package(menhirLib unix compiler-libs.common ocamlbuild findlib easy-format str BetterErrors merlin_extend re.str)
+<src/reason_utop.*>: package(menhirLib utop str re.str)
 
 <src_gen/*>: package(sedlex str)
 
--- a/package.json
+++ b/package.json
@@ -20,7 +20,8 @@
     "ocaml": "https://github.com/npm-ml/ocaml.git#npm-4.02.3",
     "easy-format": "https://github.com/npm-ml/easy-format.git#npm-1.2.0",
     "merlin-extend": "https://github.com/npm-ml/merlin-extend.git#npm-0.0.1",
-    "dependency-env": "https://github.com/npm-ml/dependency-env.git"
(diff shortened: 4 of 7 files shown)


opam
```diff
@@ -26,6 +26,7 @@ depends: [
   "utop"         {>= "1.17"}
   "BetterErrors" {>= "0.0.1"}
   "menhir"       {>= "20160303"}
+  "re"           {>= "1.5.0"}
```
jordwalke: Awesome! This will make many things easier - there's a ton of use cases for compiling to JavaScript:
If you update the npm package.json, to have a dependency on the [npm port of `re`](https://github.com/npm-ml/re) then it will continue to be installable via `npm`!

cristianoc:
@yunxing does this also cover the needs of https://github.com/facebook/reason/commit/7d5fe9033258cffa6ad011e733afc88e60b2ab14?

bobzhang:
It would be nice to avoid dependencies in the core of reason. Parser is the core, printer may be moving in this direction too, looking at the code, I think maybe we can avoid regex in general

chenglou:
@bobzhang regex is a dependency we _might_ need in the future anyway? (Is it?)

jordwalke:
@bobzhang: That would be nice to break up the library into parser and printer separately. But until that happens, I thought that having an `npm` installation should ease any concern for many people about dependencies. It is possible to run npm install, without building, `scp` the files to any other host, and then continue building everything in that source sandbox. That's how we would deal with our CI builds - would it work for you too?

jordwalke:
This looks straightforward to me. Any objections? (Bob, can you confirm that `npm` support helps you?)

bobzhang:
I don't have strong opinions on printer part, below are my concerns:

1. regex is mostly a bad abstraction and security hole(https://adtmag.com/blogs/dev-watch/2016/07/stack-overflow-crash.aspx), people mostly used it as a quick hack, if you look at the regex used here, they are very simple and can be replaced by a normal function easily
2. It is always good for user experience to have less dependencies, especially for windows build.
3. with regard to maintenance, I would bet more on `Str` instead of `ocaml-re` (which does not have any active maintainer, and its maintenance does require some expertise)

hcarty:
@bobzhang Not to derail the conversation too much - What gives you the impression that ocaml-re is not maintained? It seems to be maintained and updated as needed.

bobzhang:
@hcarty maybe I am wrong, but those changes are mostly not relevant to automata. ocaml-re is certainly a high quality library, but I am concerned to have such a dependency on the core part of reason, just my opinions

yunxing:
This diff looks good to me. But as @bobzhang mentioned, looks like all the usages here can be replaced with some simple utility functions that we can implement by our own to avoid having another dependency. Also, double unescaping a regex string makes the code harder to maintain. 

I think we should get this in for now, but in the long term we should have some utilities that cover the simple usages here.

yunxing:
Looks good to me.

yunxing:
@rickyvetter Actually, I think you missed this one: 

https://github.com/facebook/reason/blob/master/src/reason_pprint_ast.ml#L1288

rickyvetter:
@yunxing string_after isn't a regex operation. It takes an index and returns the substrings of all characters after it.

yunxing:
@rickyvetter I see. But doesn't this still depend on Str library, which still depend on C modules?

rickyvetter:
@yunxing it does. It gives warnings in the JS compilation but things will still build and run since the C deps never actually get called.

You're right though - probably best to get rid of it since because the warnings are definitely ideal and it'll make it a lot easier to tell if the module gets added back accidentally in the future! I'll make another PR to remove Str entirely.

chenglou:
`rtop` broke and bisected to this. cc @rickyvetter @yunxing

PR #656, opened 2016-07-30
Provide functions to pretty print the ast to string

rickyvetter:
Gives an alternative to printing to stdout. I think this could also be achieved without a new function by moving the responsibility of flush to the caller.

### DIFF
--- a/src/reason_pprint_ast.ml
+++ b/src/reason_pprint_ast.ml
@@ -891,10 +891,8 @@ let configure ~width ~assumeExplicitArity ~constructorLists = (
   configuredSettings := {defaultSettings with width; assumeExplicitArity; constructorLists}
 )
 
-let string_of_formatter (f: (Format.formatter -> 'a -> 'b)) x =
-  ignore (flush_str_formatter ());
-  f str_formatter x;
-  flush_str_formatter ()
+let string_of_formatter f x =
+  Format.asprintf "%a" f x
 
 let createFormatter () =
 let module Formatter = struct
--- a/src/reason_pprint_ast.ml
+++ b/src/reason_pprint_ast.ml
@@ -891,6 +891,11 @@ let configure ~width ~assumeExplicitArity ~constructorLists = (
   configuredSettings := {defaultSettings with width; assumeExplicitArity; constructorLists}
 )
 
+let string_of_formatter (f: (Format.formatter -> 'a -> 'b)) x =
+  ignore (flush_str_formatter ());
+  f str_formatter x;
+  flush_str_formatter ()
+
 let createFormatter () =
 let module Formatter = struct
 


jordwalke:
Super useful. Thanks!

rickyvetter:
I think we are on the same page - but maybe not. This just pushes all responsibility for formatter management on whoever is using the pretty printer. If you meant that you like having the explicit calls to return strings but that the logic should be moved down to the ReasonFormatter module then we can do it that way too.

src/reason_pprint_ast.ml
```diff
@@ -1177,9 +1179,7 @@ let easyFormatToFormatter f x =
     Easy_format.Pretty.define_styles fauxmatter html_escape html_style;
   let _ = Easy_format.Pretty.to_formatter fauxmatter x in
   let trimmed = strip_trailing_whitespace (Buffer.contents buf) |> String.trim in
-  Format.fprintf f "%s\n" trimmed;
-  pp_print_flush f ()
-
+  Format.fprintf f "%s\n" trimmed
```
jordwalke: Is this diff missing something? We've removed the flush call, but then someone else has to invoke the flush now, so where does that happen?

src/reason_pprint_ast.ml
```diff
@@ -1169,17 +1169,25 @@ let isListy = function
...
   let trimmed = strip_trailing_whitespace (Buffer.contents buf) |> String.trim in
-  Format.fprintf f "%s\n" trimmed;
+  Format.fprintf f "%s\n" trimmed
+
+let easyFormatToFormatter f x =
+  easyFormatToFormatterImpl f x;
   pp_print_flush f ()
 
+let easyFormatToStringFormatter x =
+  ignore (flush_str_formatter ());
```
yunxing: Why do you need to flush here twice?

rickyvetter: A safety thing - it clears the formatter before writing to it incase anything was added earlier.

chenglou:
@yunxing I'll let you merge this one

yunxing:
Hmm, I actually don't think this should be part of the API. The original API is already pretty generic so whoever uses it can implement the toString method by themselves.

If we really don't want to ask the caller to flush, maybe we can build a more generic API here:

```
let string_of_formatter (f: (Format.formatter -> 'a -> 'b)) x =
  ignore (flush_str_formatter ());
  f str_formatter x;
  flush_str_formatter ()
```

Then you / the caller can implement the same thing using:
```
let signature_string comments x = 
   string_of_formatter (signature comments) x 
```
without copying all the code.

rickyvetter:
That makes sense! I'll look into doing that.

bobzhang:
FYI, there is `Format.asprintf` which does exactly what you want

rickyvetter:
@bobzhang that's pretty sweet. I didn't realize that function existed so I'm not sure how necessary this helper is now. It's nice that Reason is providing this function so you don't have to think about the underlying formatter, but it isn't strictly necessary.

I tried getting it so the signature looks more like `string_of_formatter ?comments ~f ~ast` but I don't really understand how the types in Format work, so I wasn't able to get it running correctly.

yunxing:
The diff looks good to me. Let's get this merged.

Issue #652, opened 2016-07-24
BuckleScript `##` and `#=` don't compile


chenglou:
cc @yunxing @bobzhang @fxfactorial. (Related: #650) Repro:

```
## http createServer;
```

(Original OCaml code is `http##createServer` [here](https://github.com/bloomberg/bucklescript-addons/blob/cb493954288ae63bfb08726771ae79334e838628/examples/node-http-server/src/test_http_server.ml). Reason uses prefix notation).

`bsc -pp refmt -impl test.re` errors with:

```
File "test.re", line 1, characters 0-1:
Error: 0: <SYNTAX ERROR>
```

Schmavery:
Do we maybe just need to recognize it as an identifier and make sure reason doesn't die on it? Then we can run the regular ## preprocessor on the .ml code?
This seems to be blocking the regular use of js bindings. When we wrote reason code with js_of_ocaml we had to use the workaround of writing the code that the preprocessor generates (which works fine but ## is more convenient).

jordwalke:
@Schmavery : This sounds correct. It should be just a matter of looking at what OCaml's parser does and then recreating it in the Reason parser. Another Good First Task (tagging as such). Let me know if this begins to block you on something and we'll elevate the priority.

bobzhang:
If Reason moves in the direction to avoid `#` operator in the concrete syntax, that would be cool (better `cpp` compatibility is a huge win)
For example, if we decide to use `..` for Js object method dispatching, we don't need go through ppx, the `##` rewrite rule(at least for BuckleScript) is pretty simple, so we can directly construct the AST node without ppx.

chenglou:
Resolved already =)

Issue #650, opened 2016-07-21
Infix printing for `#=` and `##`


chenglou:
cc @bobzhang 

The BuckleScript readme example prints in prefix notation for reason:

```rust
let port = 3000;

let hostname = "127.0.0.1";

let create_server http => {
  let server = (## http createServer) (
  (
    fun req resp => {
      #= (## resp statusCode) 200;
      (## resp setHeader) "Content-Type" "text/plain";
      (## resp _end) "Hello world\n"
    }
  )
  [@bs]
  );
  (## server listen)
    port
    hostname
    (
    (
      fun () => Js.log (
        "Server running at http://" ^ hostname ^ ":" ^ Pervasives.string_of_int port ^ "/"
      )
    )
    [@bs]
    )
};

let () = create_server Http_types.http;
```

jordwalke:
Oh ## is a standard function identifier?

bobzhang:
OCaml treats `#_` operator in a special way, let me find the link

bobzhang:
http://caml.inria.fr/pub/docs/manual-ocaml/extn.html#sec253
Operator names starting with a # character and containing more than one # character are reserved for extensions.

jordwalke:
Thanks for the info.

chenglou:
@yunxing any solution for this? I'm not familiar with this at all

Schmavery:
Do we maybe just need to recognize it as an identifier and make sure reason doesn't die on it?  Then we can run the regular ## preprocessor on the .ml code?
This seems to be blocking the regular use of js bindings.  When we wrote reason code with js_of_ocaml we had to use the workaround of writing the code that the preprocessor generates (which works fine but ## is more convenient).

jordwalke:
It should be as easy as considering this of `Infix` precedence (there's a function in the printer that checks to see if the leading character is infix - we'd just update it to also look for `#`). This is like a GoodFirstTask.

yunxing:
@chenglou Sorry just saw this. 

@bobzhang @Schmavery I guess I'm a bit confused here. From the parser it seems like `#_` is still parsed as infix operators? How does this related to extensions ?

 https://github.com/ocaml/ocaml/blob/d6c72fc82a8950851da5f121b80a4e1c8bff377a/parsing/parser.mly#L1569

bobzhang:
@Drup may know more about it

Drup:
`#...` are infix operators (with priority higher than function application, just like `#`).
[...]

PR #657, opened 2016-08-01
Add rough support for extension operators

Schmavery:
Here's some rough work to see how we could support extension operators (##, #=) in reason.
This seems to work with bucklescript in our initial testing but we haven't managed to get everything together for js_of_ocaml (there seems to be some issue with <-). @chenglou 
Related: #650, #652

### DIFF
--- a/src/reason_lexer.mll
+++ b/src/reason_lexer.mll
@@ -446,6 +446,8 @@ rule token = parse
         lexbuf.lex_curr_p <- { curpos with pos_cnum = curpos.pos_cnum - 1 };
         STAR
       }
+  | "#" (appropriate_operator_suffix_chars | "#")
+      { SHARPOP(Lexing.lexeme lexbuf) }
   | "#" [' ' '\t']* (['0'-'9']+ as num) [' ' '\t']*
         ("\"" ([^ '\010' '\013' '"' ] * as name) "\"")?
         [^ '\010' '\013'] * newline
--- a/src/reason_lexer.mll
+++ b/src/reason_lexer.mll
@@ -446,7 +446,7 @@ rule token = parse
         lexbuf.lex_curr_p <- { curpos with pos_cnum = curpos.pos_cnum - 1 };
         STAR
       }
-  | "#" (appropriate_operator_suffix_chars | "#")
+  | "#" (appropriate_operator_suffix_chars | "#")+
       { SHARPOP(Lexing.lexeme lexbuf) }
   | "#" [' ' '\t']* (['0'-'9']+ as num) [' ' '\t']*
         ("\"" ([^ '\010' '\013' '"' ] * as name) "\"")?
--- a/src/reason_pprint_ast.ml
+++ b/src/reason_pprint_ast.ml
@@ -379,7 +379,7 @@ let rec sequentialIfBlocks x =
 
 let prefix_symbols  = [ '!'; '?'; '~' ] ;;
 let infix_symbols = [ '='; '<'; '>'; '@'; '^'; '|'; '&'; '+'; '-'; '*'; '/';
-                      '$'; '%'; '\\' ]
+                      '$'; '%'; '\\'; '#' ]
 let operator_chars = [ '!'; '$'; '%'; '&'; '*'; '+'; '-'; '.'; '/';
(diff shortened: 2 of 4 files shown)


facebook-github-bot-7:

hhugo:
js_of_ocaml (ppx) syntax is `obj##meth`, `obj##.prop` and `obj##.prod := exp`

yunxing:
@Schmavery What's the issue with "<-" ?

src/reason_lexer.mll
```diff
@@ -446,6 +446,8 @@ rule token = parse
         lexbuf.lex_curr_p <- { curpos with pos_cnum = curpos.pos_cnum - 1 };
         STAR
       }
+  | "#" (appropriate_operator_suffix_chars | "#")
```
yunxing: This will only work with operators that are two characters long, right? I think you need to use a "?" after the parens.

Schmavery: I was actually just cleaning this up to address my previous issues with it and added a "+", which I believe is more appropriate?

yunxing: You are right, we should use "+" instead of "?".

Schmavery:
It looks like the [jsoo code](https://github.com/ocsigen/js_of_ocaml/blob/master/examples/wiki/main.ml#L38) that we tried to use as a test example was actually using the old ocamlp4 proprocessor.  I added the '+' that I forgot on the lexer regex (based on [what ocaml does](https://github.com/ocaml/ocaml/blob/2c96a267dc3bf715d64aa134c031602d398d90ab/parsing/lexer.mll#L517)) and now I seem to be able to compile both jsoo and bucklescript examples (using the ppx syntax).

chenglou:
cc @bobzhang, apparently this PR also works with BuckleScript!

yunxing:
@Schmavery
Looks good to me. 

Maybe add some test cases with these operators? Especially how they interact with "<-" (you want to make sure it's left associative and has higher precedence than "<-")

src/reason_pprint_ast.ml
```diff
@@ -405,6 +405,9 @@ let rules = [
       (fun s -> (`Right, s = "asr"));
     ];
     [
+      (fun s -> (`Left, s.[0] == '#'));
```
jordwalke: Just to confirm - this order in the rules matches the order of token precedence defined in `reason_parser.mly`?

jordwalke:
Looking great. Test cases would be very good. I'm currently rewriting all of that precedence checking/printing code (for infix combined with assignment `=`, `:=`, etc). But if the test cases are in place, it will ensure I don't break it!

yunxing:
@jordwalke Good point about "<-".

@Schmavery :

`<-` is represented as `=` in Reason.
So in the example you were trying, it should be written as
```
preview##style##border = Js.string "1px black dashed";
```

jordwalke:
So it looks great as long as the precedence is correct, and some test cases are added so that I make sure I don't break it in the diffs I'm about to send.

Some ideas for test cases:
- Making sure that it abides by the left precedence.  `x##y##z` should be parsed and printed exactly the same as `(x##y)##z` - both should be printed without parens. Likewise `x##(y##z)` should have its parens preserved when printing due to the left biased precedence.

jaredly:
@jordwalke what could `x##(y##z)` possibly mean? surely not that `y##z` returns a string and then you access that property on `x`? I should think that property access would only be left-recursive

bobzhang:
@jaredly x##method_name is still static, so in most cases, x##(y##z) is a compile error

chenglou:
I'm literally doing `let result = js_method_run1 (!react)#createElement foo;` right now. Someone merge this before I rage merge this and break reason

jordwalke:
This has merge conflicts. If someone resolves them, we can insta-merge as soon as the tests pass.

chenglou:
It's just `reason_parser.messages`. I'll resolve locally and merge.

jordwalke:
okay

jordwalke:
You *coullldd* add a quick little test case too :D

chenglou:
Ok, will add test when I wake up.

PR #673, opened 2016-08-10
[atom-reason] Hyperclick improvements

chenglou:
- Disable underline when no reference can be found.
- New feature: display why a reference can't be found using Notiflyer (good stuff).
- Small refactors, less JS.


chenglou:
Merging; retrospective comments welcome.

PR #674, opened 2016-08-11
Fix rtop

chenglou:
All these meta files are making me hungry!

See notes here:
https://github.com/facebook/reason/pull/659/files#r73099166

This fixes it.

### DIFF
--- a/src/rtop_init.ml
+++ b/src/rtop_init.ml
@@ -1,5 +1,6 @@
 #!/usr/bin/env ocaml
 #require "menhirLib";;
+#require "re.str";;
 #require "reason";;
 #use ".reasoninit";;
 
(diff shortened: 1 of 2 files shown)


jordwalke:
Thank you, @chenglou !

chenglou:
**Don't merge this yet**! #675 should currently be failing CI but it's not. Once that one successfully fails, we can merge this fix.

chenglou:
LOL talk about timing lol. Alright well, I'll see how I can fix #675.

src/rtop_init.ml
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env ocaml
 #require "menhirLib";;
+#require "re.str";;
```
rickyvetter: Gah! I didn't know this file existed. I kept just making the other change and not being able to fix it. Thank you :)

PR #675, opened 2016-08-11
Simple rtop integration test

chenglou:
To prevent regression.


jordwalke:
Great this would have saved at least two issues.

chenglou:
Wait don't merge yet.

chenglou:
YES! It finally failed!

PR #676, opened 2016-08-12
[atom-reason] Make things work with ocaml

chenglou:
- Hyperclick
- Formatter
- Diagnosis
- Recoverable syntax
- Autocomplete (shows correct syntax)
- Type hint (prints to the correct syntax)
- Multiselect in scope

Works alongside other reason files.


chenglou:
I need to thanks @let-def for the amazing merlin work. This is a ridiculously low amount of change needed to support OCaml (actually, no change on merlin's side at all...).

cristianoc:
This must be the highest ratio of achievements per code changes I've seen.

PR #677, opened 2016-08-13
Activate atom plugin for ocaml too.

pvolok:
Here is my photo after 2 hours of trying to understand why atom-reason loads in reason repo but does not in flow repo: ლ(ಠ益ಠლ)

### DIFF
--- a/editorSupport/atom-reason/package.json
+++ b/editorSupport/atom-reason/package.json
@@ -8,7 +8,8 @@
     "Reason"
   ],
   "activationHooks": [
-    "language-reason:grammar-used"
+    "language-reason:grammar-used",
+    "language-ocaml:grammar-used"
   ],
   "devDependencies": {
     "jengaboot": "0.0.22"


jordwalke:
@chenglou, do you know why this was needed, and why [your diff](https://github.com/facebook/reason/pull/676/files) was not sufficient?


Thanks for your pull request @pvolok !

chenglou:
Oh crap, so sorry. I tried on a hybrid repo but not on a pure ocaml one... I assume this is the fix?

pvolok:
@chenglou From http://flight-manual.atom.io/hacking-atom/sections/package-word-count/#packagejson:
> `activationHooks`: an Array of Strings identifying hooks that trigger your package's activation. The loading of your package is delayed until one of these hooks are triggered.

>  Currently, the only activation hook is `language-package-name:grammar-used` (e.g., `language-javascript:grammar-used`)

Now atom-reason will be activated when either reason or ocaml file is open. So yes, assuming we want to delay atom-reason activation, this should be a fix.

chenglou:
Great. Sorry for the oversight.

jordwalke:
thank you!

PR #679, opened 2016-08-14
Add outline for atom-reason.

pvolok:


chenglou:
Fuck, this is awesome...
Can I have a screenshot please?

pvolok:
Of course :)

outline

jordwalke:
Whoa! Awesome!

chenglou:
Great, thanks! I'll try it =D

pvolok:
Oh no! I forgot the order is reversed... One sec.

chenglou:
Oops, already merged. While we're at it, can you make sure that you run `npm start` one last time to get to latest js/map artifacts?

chenglou:
Also, can we have these printed as Reason code when it's in a `re`/`rei` file?

pvolok:
.map file also contains lines like this: `"/Users/chenglou/.opam/4.02.3/lib/js_of_ocaml/fs.js"`. I will find-replace my username for now.

I will check how reason signatures are different later. Have to fix the order and go to the bed.

chenglou:
Yeah the source map absolute path problem is a KP (cc @SanderSpies). Currently it's not too important, as it's developer-facing and if you were checking these things locally, after the build command it'll be pointing to your local absolute path anyway (thus loading the source maps correctly).

I see that the current `val` and `sig` and such outline labels are hard-coded; This is easy to fix with another layer of hard-coding. Btw, [here's a handy ml <-> re conversion trick](https://github.com/facebook/reason/wiki/Newcomer-Tips-&-Tricks#ocaml---reason-conversion-tricks).

PR #680, opened 2016-08-14
Reverse items order in the atom outline.

pvolok:
Merlin reports outline in the reversed order.


chenglou:
Left a comment on https://github.com/facebook/reason/pull/679#issuecomment-239699966

PR #682, opened 2016-08-16
[atom-reason] Outline: `val` for OCaml, `let` for Reason.

pvolok:
This one seems to be the only difference in reason/ocaml signatures.

As per https://github.com/facebook/reason/pull/679#issuecomment-239699966: How about gitignoring `output_byte_debug_js/app.map` to keep diffs cleaner?

### DIFF
--- a/editorSupport/atom-reason/src/AtomReasonOutline.re
+++ b/editorSupport/atom-reason/src/AtomReasonOutline.re
@@ -10,14 +10,17 @@ let buildToken kind value => Js.Unsafe.obj [|
   ("value", Js.Unsafe.inject (Js.string value))
 |];
 
-let makeTokens data => {
+let makeTokens grammar data => {
   let merlinKind = Js.to_string (Js.Unsafe.get data "kind");
   let name = Js.to_string (Js.Unsafe.get data "name");
   let kind =
     switch merlinKind {
-    | "Value" => "val"
+    | "Value" =>
+      switch grammar {
+      | `Reason => "let"
+      | `OCaml => "val"
+      }
     | "Constructor" => "ctor"
-    | "Signature" => "sig"
     | kind => String.lowercase kind
     };
   let nameToken =
@@ -26,8 +29,7 @@ let makeTokens data => {
     | "Class"
     | "Constructor"
     | "Exn"
-    | "Module"
-    | "Signature" => buildToken "class-name" name
+    | "Module" => buildToken "class-name" name
     | "Type" => buildToken "type" name
     | kind => buildToken "plain" name
     };
@@ -40,7 +42,7 @@ let makeTokens data => {
   )
 };
 
-let rec convertOutlines entries => {
+let rec convertOutlines grammar entries => {
   let entries =
     List.rev_map
       (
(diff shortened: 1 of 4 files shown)


chenglou:
Yeah we should ignore `app.map`. Because of the absolute paths, it's useless for anyone else's computer anyway.

editorSupport/atom-reason/src/AtomReasonOutline.re
```diff
@@ -26,8 +29,7 @@ let makeTokens data => {
     | "Class"
     | "Constructor"
     | "Exn"
-    | "Module"
-    | "Signature" => buildToken "class-name" name
```
chenglou: That's intended right

pvolok: Oh. I didn't find `Signature` [here](https://github.com/the-lambda-church/merlin/blob/2e014d8a8bd44af25796b292706889c5513599c1/src/frontend/protocol.mli#L66), and thought it was added by mistake. Will bring it back.

chenglou:
ah yeah, you'd have to remove it from https://github.com/facebook/reason/blob/3bba4793005c5a5e745d829f640983252f8b81f9/pkg/build.ml#L57. https://travis-ci.org/facebook/reason/builds/152614269#L1140

chenglou:
Can you split out the `.map` change into another PR so that it can be easily reverted if needed? The current build process is a bit manual.

editorSupport/atom-reason/src/AtomReasonOutline.re
chenglou: Yeah that's weird. It's described [here](https://github.com/the-lambda-church/merlin/blob/master/doc/dev/PROTOCOL.md#completion)

pvolok:
Sure.

One thing about Signature. Marlin outline reports this code as a `Signature`:
```ocaml
module type X_int = sig val x : int end;;
```
I'm not sure how to show it in the outline. `module type` is too verbose imo. What do you think?

chenglou:
cc @jordwalke: got any preference? Context: this is a hard-coded translation of e.g. `val foo` to `let foo` in the nuclide outline (which you should totally try btw):
screenshot 2016-08-16 02 18 49

Also @pvolok: how come it's `"Constructor"` and all and not `"constructor"` lower-case as written [here](https://github.com/the-lambda-church/merlin/blob/05d20b68275b9a817ab42a62d34c4583e0109380/doc/dev/PROTOCOL.md#completion)?

pvolok:
@chenglou https://github.com/the-lambda-church/merlin/blob/2e014d8/src/frontend/IO.ml#L108

chenglou:
Somewhat related: what if we map `Constructor` to `|` instead? And `label` to nothing for a visual parallel with the actual record?

pvolok:
That sounds great. I'm being somewhat conservative since I'm still newbie in ocaml and don't know  in what contexts different types may appear.

chenglou:
We're hard-coding the strings already and there's no precedent, so I guess we can have some liberty here. If these are used in unexpected situations we can come back and reconsider what they display.

pvolok:
Changed for ctors and labels: 
outline

I left signatures as `module type` for now. Until Jordan answers you. So please don't merge yet.

chenglou:
Ehhh... now that `|` looks confusing because it looks like no constructor carries a value anymore. Maybe I'll get used to it. Will check back in the morning.

chenglou:
Comment from jordan: "can we jus use what's there"?
That'd be a solution that works for both ocaml and reason syntax. Though kinda verbose ¯\_(ツ)_/¯.

pvolok:
I'd just go with `Signature -> module type` to make it more intuitive. As for ctor and label, I'd be cool with either display option. You choose and we move on :)

PR #684, opened 2016-08-16
Fix "Various Differences" example.

pvolok:


chenglou:
Hah, this has been there for a while now! I thought it was fixed. Thanks!

PR #685, opened 2016-08-16
Gitignore output_byte_debug_js.

pvolok:


chenglou:
ccing @jordwalke for good measure. Context is https://github.com/facebook/reason/pull/682#issuecomment-240037340

PR #687, opened 2016-08-16
[atom-reason] Remove useless menu items

chenglou:
OS X bar and right click, they all have these boilerplate toggle commands.


chenglou:
Did the same over at https://github.com/chenglou/atom-reason-loader/commit/9b2ce374d47df4d4abaa265a7db22604bcb78175

cc @frantic. Thanks for discovering these.

PR #692, opened 2016-08-17
Give a syntax error when `of` is used before a constructor argument.

### LINKED ISSUE #690: Old reason variant syntax with `of` accidentally prints correctly into ocaml
cc @sanderspies
Repro:

```
type foo = A of string;
```

prints to:

```
type foo =
  | A of of* string
```

Without good tooling, it's hard to know why this happened. Maybe we should put a temp warning. Or maybe we should just let this be since people will upgrade.

SanderSpies:
Fixes #690.


yunxing:
LGTM.

PR #695, opened 2016-08-17
[atom-reason] Linux support

chenglou:
See comments on https://github.com/chenglou/atom-reason-loader/commit/7c91672670efe84a0a170adeb9276709e4146548

Big thanks to @vramana for helping me debug this =)


chenglou:
**Don't merge yet!**

jordwalke:
So to confirm, does this make anything worse on Mac, for recent Atom versions? Does it make it better on Linux? If the answer to both is yes, shouldn't we merge?

Also, the long term fix is to set environment variables from the project sandbox, which doesn't rely on loading the environment/PATH from the user's shell, so long term this won't be an issue. Cc @yunxing

vramana:
[This](https://github.com/ioquatix/script-runner/blob/v1.5.0/lib/script-runner.coffee#L45-L63) looks like a possible fix as described [here](https://discuss.atom.io/t/atom-command-doesnt-pass-environment-variables-to-atom/1596/6)

chenglou:
Closed in favor of #699

PR #696, opened 2016-08-18
[parser] Nit: better error for `of`

chenglou:
See https://github.com/facebook/reason/commit/e2fb37146d738afce3fac55715d3c35875d2282c#diff-93c201b41bf10dfca8380188d820970aR3603

Also, utop uses the outcome printer and sucks at printing this:
screenshot 2016-08-17 17 44 03

### DIFF
--- a/src/reason_parser.mly
+++ b/src/reason_parser.mly
@@ -3600,7 +3600,7 @@ sig_exception_declaration:
 generalized_constructor_arguments:
     /*empty*/                                   { ([],None) }
   | as_loc(OF)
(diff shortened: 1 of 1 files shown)


SanderSpies:
Yes, much better

chenglou:
Yunxing came up with a better one: 
```
"of" is not needed in reason, just use `type a = Foo a` instead.
```

Probably a bit more impressive =).

hcarty:
@chenglou I like that minus the 'just' to keep the message direct.

yunxing:
Commit updated and merged via 2f072aabf1d6a3af9eecfe8c3cbb662860c10d92

PR #698, opened 2016-08-19
[atom-reason] Prevent hyperclick notiflyer banner during hover

chenglou:
Hyperclick unfortunately activates too eagerly; this includes simply pressing cmd. We'll have to delay showing notiflyer upon click.


jordwalke:
Awesome. Let's try this.

PR #700, opened 2016-08-19
Docs into master

vramana:
Since Github Pages can be [generated][url] from master branch's docs folder, I am submitting this PR.  The main advantage of this is 
> This lets you maintain documentation and code together on one branch, and open source maintainers can accept contributions for both in a single pull request.

It seems to build correctly. Check [here][reason] 

[reason]: https://vramana.github.io/reason/
[url]:  https://github.com/blog/2228-simpler-github-pages-publishing


chenglou:
woah 61 commits

vramana:
I wanted to preseve the history of the files in the master branch. Although the github doesn't show it. You can track the history by looking that parent commits. If you think history isn't so important, then I can squash them.

chenglou:
Yeah I think erasing history here is fine. cc @jordwalke

vramana:
@chenglou Updated the PR

chenglou:
@vramana what's the second point [here](https://help.github.com/articles/configuring-a-publishing-source-for-github-pages/#publishing-your-github-pages-site-from-a-docs-folder-on-your-master-branch) about?

"your repository must not follow the repository naming scheme `<username>.github.io` or `<orgname>.github.io`"

Also, can you add an `.npmignore` to the root and ignore `docs/`?

vramana:
@chenglou They are telling the repo name can't be `facebook.github.io`. 

> Also, can you add an .npmignore to the root and ignore docs/?

Yes. I will do that

jordwalke:
Will this make it so jQuery is returned in git grep and github search?

vramana:
@jordwalke It seems so.

jordwalke:
I think you can fix that with a simple .gitattributes file. Want to see if that will work? Thanks for sending this PR!

jordwalke:
Why is CI so slow?

chenglou:
The whole Facebook organization shares the same travis limit I think. So we're queuing behind other facebook/ projects. Faster at night.

vramana:
@jordwalke I didn't get you. I have never used gitattributes in the past.

chenglou:
@vramana see my comment. `.gitattributes` specifies which file (or globs) are supposed to be treated as binary. These receive special treatment, including not being displayed in the diff view when they change, and probably not being indexed by github search.

vramana:
@chenglou Seems like I missed your previous comment. Which files should I mark at binary? All the vendor files?

vramana:
Also do we need to keep the examples folder? It seems completely unrelated.

vramana:
How to mark a folder recursively as binary in gitattributes? Currently I supress all jQuery mentions from invidually target the folders. But I don't think it's ideal.

jordwalke:
Does double star `**` work?

vramana:
@jordwalke It does. What else work is pending in this PR?

jordwalke:
I think this is good to commit now. We may have to revert (or find the right fix) if jQuery ends up showing up in searches.

vramana:
@jordwalke I have checked from git grep. It doesn't show up.

jordwalke:
@chenglou What else do we have to do? Do we just merge this and then delete the docs branch or can we keep it around?

chenglou:
`.npmignore`, `.gitattributes`, can't think of anything else. I think we're good merging this =)

chenglou:
Seems working. Lemme test by making changes to the new master docs.

chenglou:
Alright, coordinated with the internal folks to make the change! Thanks @vramana! =D

jordwalke:
thanks!

PR #702, opened 2016-08-20
[Reason] Fix Many Parsing Bugs (including incorrect ternary) by Refac…

jordwalke:
When converting the css-layout library to Reason, I found the following "bug".

```rust
x.contents = e1 ? e2 : e3;
```

Most people (and JS) would expect that to be parsed as:
```rust
x.contents = (e1 ? e2 : e3);
```

But Reason was parsing/printing it as:

```rust
(x.contents = e1) ? e2 : e3;
```

It was very difficult to fix this without a major refactor, because there's many combination of infix operators, update token `=` and ternary `?` that didn't fit into the existing "framework" of printing. So I implemented light version of "unparsing", which resembles a more general unparsing strategy. It uses rules tables that rank token precedences (and also custom `%prec` rules that we've used) in order to determine when to add precedence `()` wrappers on various parts of the printing. Now, this is not generalized, so we cannot automatically generate a printing algorithm from new parsing rules we add, but it's an exploration in that direction.

In the process, it helped me fix a few bugs that were outstanding, eliminate a few unnecessary parens, and generally consolidate all the crazy logic about "fixity" (`Infix/`Prefix/`Normal) into one place.

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/attributes.re
+++ b/formatTest/typeCheckedTests/expected_output/attributes.re
@@ -68,7 +68,7 @@ let x = "hello" [@onHello] ^ "goodbye";
 
 let x = "hello" ^ "goodbye" [@onGoodbye];
 
-let x = (^) "hello" "goodbye" [@onEverything];
+let x = ("hello" ^ "goodbye") [@onEverything];
 
 let x = 10 + 20 [@on20];
 
@@ -78,7 +78,7 @@ let x = 10 [@on10] + 20;
 
 let x = 10 [@on10] + 20;
 
-let x = (+) 10 20 [@attrEverything];
+let x = (10 + 20) [@attrEverything];
 
 let x = 10 - 20 [@on20];
 
@@ -88,7 +88,7 @@ let x = 10 [@on10] - 20;
 
 let x = 10 [@on10] - 20;
 
-let x = (-) 10 20 [@attrEntireEverything];
+let x = (10 - 20) [@attrEntireEverything];
 
 let x = true && false [@onFalse];
 
@@ -98,7 +98,7 @@ let x = true [@onTrue] && false;
 
 let x = true [@onTrue] && false;
 
-let x = (&&) true false [@attrEverything];
+let x = (true && false) [@attrEverything];
 
 /* now make sure to try with variants (tagged and `) */
 
@@ -117,21 +117,20 @@ let res =
 let res =
   (if true {false} else {false}) [@onEntireIf];
 
-let add a b => (+) (a [@onA]) b [@onEverything];
+let add a b => (a [@onA] + b) [@onEverything];
 
 let add a b =>
-  (+) (a [@onA]) (b [@onB]) [@onEverything];
+  (a [@onA] + b [@onB]) [@onEverything];
 
 let add a b => a + b [@onB];
 
 let both = (fun a => a) [@onEntireFunction];
 
-let both a b =>
-  (&&) (a [@onA]) b [@onEverything];
+let both a b => (a [@onA] && b) [@onEverything];
 
 let both a b => a [@onA] && b [@onB] [@onB];
 
-let both a b => (&&) a b [@onEverything];
+let both a b => (a && b) [@onEverything];
 
 let thisVal = 10;
 
@@ -139,19 +138,17 @@ let x =
   20 + (- add thisVal thisVal [@onFunctionCall]);
 
 let x =
-  (+) 20 (- add thisVal thisVal) [@onEverything];
+  (20 + (- add thisVal thisVal)) [@onEverything];
 
 let x = - add thisVal thisVal [@onFunctionCall];
 
-let x =
-  (~-) (add thisVal thisVal) [@onEverything];
+let x = (- add thisVal thisVal) [@onEverything];
 
 let bothTrue x y => {contents: x && y};
 
 let something =
-  (!)
-    (bothTrue true true)
-    [@onEverythingToRightOfEquals];
+  !(bothTrue true true)
+  [@onEverythingToRightOfEquals];
 
 let res =
   add 2 4 [@appliesToEntireFunctionApplication];
@@ -162,7 +159,8 @@ let myObj = {method p () => {method z () => 10}};
 
 let result =
   (myObj#p () [@attOnFirstSend])#z
-    () [@onSecondSend];
+  ()
+  [@onSecondSend];
 
 type recordFunctions = {
   p: unit => recordFunctions [@onUnit],
@@ -179,7 +177,8 @@ and unused = ();
 
 let result =
   (myRecord.p () [@attOnFirstSend]).q
-    () [@onSecondSend];
+  ()
+  [@onSecondSend];
 
 type variantType =
   | Foo int [@onInt] | Bar (int [@onInt]) | Baz
--- a/formatTest/typeCheckedTests/expected_output/attributes.re
+++ b/formatTest/typeCheckedTests/expected_output/attributes.re
@@ -150,6 +150,9 @@ let something =
   !(bothTrue true true)
   [@onEverythingToRightOfEquals];
 
+let something =
+  !(bothTrue true true [@onlyOnArgumentToBang]);
+
 let res =
   add 2 4 [@appliesToEntireFunctionApplication];
 
--- a/formatTest/typeCheckedTests/expected_output/attributes.re
+++ b/formatTest/typeCheckedTests/expected_output/attributes.re
@@ -60,45 +60,45 @@ let x = "hello" [@onHello];
 
 let x = "hello" [@onHello];
 
-let x = "hello" ^ ("goodbye" [@onGoodbye]);
+let x = "hello" ^ "goodbye" [@onGoodbye];
 
-let x = ("hello" [@onHello]) ^ "goodbye";
+let x = "hello" [@onHello] ^ "goodbye";
 
-let x = ("hello" [@onHello]) ^ "goodbye";
+let x = "hello" [@onHello] ^ "goodbye";
 
-let x = "hello" ^ ("goodbye" [@onGoodbye]);
+let x = "hello" ^ "goodbye" [@onGoodbye];
 
-let x = ("hello" ^ "goodbye") [@onEverything];
+let x = (^) "hello" "goodbye" [@onEverything];
 
-let x = 10 + (20 [@on20]);
+let x = 10 + 20 [@on20];
 
-let x = 10 + (20 [@on20]);
+let x = 10 + 20 [@on20];
 
-let x = (10 [@on10]) + 20;
+let x = 10 [@on10] + 20;
 
-let x = (10 [@on10]) + 20;
+let x = 10 [@on10] + 20;
 
-let x = (10 + 20) [@attrEverything];
+let x = (+) 10 20 [@attrEverything];
 
-let x = 10 - (20 [@on20]);
+let x = 10 - 20 [@on20];
 
-let x = 10 - (20 [@on20]);
+let x = 10 - 20 [@on20];
 
-let x = (10 [@on10]) - 20;
+let x = 10 [@on10] - 20;
 
-let x = (10 [@on10]) - 20;
+let x = 10 [@on10] - 20;
 
-let x = (10 - 20) [@attrEntireEverything];
+let x = (-) 10 20 [@attrEntireEverything];
 
-let x = true && (false [@onFalse]);
+let x = true && false [@onFalse];
 
-let x = true && (false [@onFalse]);
+let x = true && false [@onFalse];
 
-let x = (true [@onTrue]) && false;
+let x = true [@onTrue] && false;
 
-let x = (true [@onTrue]) && false;
+let x = true [@onTrue] && false;
 
-let x = (true && false) [@attrEverything];
+let x = (&&) true false [@attrEverything];
 
 /* now make sure to try with variants (tagged and `) */
 
@@ -117,44 +117,40 @@ let res =
 let res =
   (if true {false} else {false}) [@onEntireIf];
 
-let add a b => ((a [@onA]) + b) [@onEverything];
+let add a b => (+) (a [@onA]) b [@onEverything];
 
 let add a b =>
-  ((a [@onA]) + (b [@onB])) [@onEverything];
+  (+) (a [@onA]) (b [@onB]) [@onEverything];
 
-let add a b => a + (b [@onB]);
+let add a b => a + b [@onB];
 
 let both = (fun a => a) [@onEntireFunction];
 
 let both a b =>
-  ((a [@onA]) && b) [@onEverything];
+  (&&) (a [@onA]) b [@onEverything];
 
-let both a b => (a [@onA]) && (b [@onB] [@onB]);
+let both a b => a [@onA] && b [@onB] [@onB];
 
-let both a b => (a && b) [@onEverything];
+let both a b => (&&) a b [@onEverything];
 
 let thisVal = 10;
 
-let x = 20 + (
-  - (add thisVal thisVal [@onFunctionCall])
-);
+let x =
+  20 + (- add thisVal thisVal [@onFunctionCall]);
 
 let x =
-  (20 + (- (add thisVal thisVal)))
-    [@onEverything];
+  (+) 20 (- add thisVal thisVal) [@onEverything];
+
+let x = - add thisVal thisVal [@onFunctionCall];
 
 let x =
-  -
-  (add thisVal thisVal [@onFunctionCall]);
-
-
-let x =
-  (- (add thisVal thisVal)) [@onEverything];
+  (~-) (add thisVal thisVal) [@onEverything];
 
 let bothTrue x y => {contents: x && y};
 
 let something =
-  !(bothTrue true true)
+  (!)
+    (bothTrue true true)
     [@onEverythingToRightOfEquals];
 
 let res =
--- a/formatTest/typeCheckedTests/expected_output/imperative.re
+++ b/formatTest/typeCheckedTests/expected_output/imperative.re
@@ -55,9 +55,11 @@ try (
 };
 
 let result =
-  while false {
-    ()
-  } ==
+  (
+    while false {
+      ()
+    }
+  ) ==
   () ?
     false : true;
 
--- a/formatTest/typeCheckedTests/expected_output/imperative.re
+++ b/formatTest/typeCheckedTests/expected_output/imperative.re
@@ -55,12 +55,10 @@ try (
 };
 
 let result =
-  (
-    while false {
-      ()
-    }
-  ) ==
-    () ?
+  while false {
+    ()
+  } ==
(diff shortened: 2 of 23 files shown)


formatTest/typeCheckedTests/expected_output/attributes.re
```diff
@@ -60,45 +60,45 @@ let x = "hello" [@onHello];
 
 let x = "hello" [@onHello];
 
-let x = "hello" ^ ("goodbye" [@onGoodbye]);
+let x = "hello" ^ "goodbye" [@onGoodbye];
```
jordwalke: Currently, are parsed as if they are final arguments to a function, and apply to that entire function call. The parens were unnecessary. I still don't like this convention (I'd rather have attributes come before the function), but we can fix that later.

formatTest/typeCheckedTests/expected_output/attributes.re
```diff
@@ -60,45 +60,45 @@ let x = "hello" [@onHello];
...
+let x = 10 + 20 [@on20];
 
-let x = (10 [@on10]) + 20;
+let x = 10 [@on10] + 20;
 
-let x = (10 [@on10]) + 20;
+let x = 10 [@on10] + 20;
 
-let x = (10 + 20) [@attrEverything];
+let x = (+) 10 20 [@attrEverything];
```
jordwalke: To simplify the implementation (and also learning how to use attributes), I just made it so that if you want to apply an attribute to an infix function, you must use the same syntactic patters as with regular function application. Every infix/prefix identifier can be used as a regular identifier, so it works. This simplifies things greatly. We likely had bugs with attributes with various combinations of infix/prefix/unary and now that everything must follow this syntax for attributes, I'm much more confident that everything is correct.

formatTest/typeCheckedTests/expected_output/attributes.re
```diff
@@ -117,44 +117,40 @@ let res =
...
 let x =
-  -
-  (add thisVal thisVal [@onFunctionCall]);
+  (+) 20 (- add thisVal thisVal) [@onEverything];
 
+let x = - add thisVal thisVal [@onFunctionCall];
 
 let x =
-  (- (add thisVal thisVal)) [@onEverything];
+  (~-) (add thisVal thisVal) [@onEverything];
```
jordwalke: Since the unary `-` minus function application had an attribute on the entire application, it is converted to normal application syntax, and in the case of unary minus, the *actual* identifier is `~-`. It's just that when written as "prefix" application, its true nature is hidden away.

formatTest/typeCheckedTests/expected_output/imperative.re
```diff
@@ -55,12 +55,10 @@ try (
 };
 
 let result =
-  (
-    while false {
-      ()
-    }
-  ) ==
-    () ?
+  while false {
```
jordwalke: Also, these parens were not needed. It makes the monadic bind code `>==` much more beautiful as well. For example, @samoht found that our function printing was far too verbose:

```rust
entry_of_file t index file mode sha1 blob >>= (
  fun | None => Lwt.return_unit 
      | Some e => e
)
```
(Imagine when they are nested, the extra parens really disrupt the flow).

But with the new logic, they will be printed much better (still not perfect, but progress!)

```rust
entry_of_file t index file mode sha1 blob >>=
fun | None => Lwt.return_unit 
    | Some e => e
```

formatTest/unit_tests/expected_output/basicStructures.re
```diff
@@ -57,12 +75,21 @@ let parensNeededAroundFirst = (!blah)#foo#bar;
 
 let parensNeededAroundSecond = (!blah#foo)#bar;
 
-/* The following two have an error in formatting! We must check for
- * *consecutive* prefix operators and either space separate or guard in
- * parens. */
-let x = !!foo.bar;
```
jordwalke: The refactor also made it easier to fix this bug.

formatTest/unit_tests/expected_output/infix.re
```diff
@@ -348,3 +358,568 @@ let myFunc
...
+  val bigArr = "goodThingThisIsntTypeChecked";
+  val str = "string";
+  method testCases () => {
+
+    /**
+     * The lowest precedence token is =, followed by :=, and then ?, then :.
+     *
+     * The following text
+     *
+     *     x.contents = tenaryTest ? ifTrue : ifFalse
```
jordwalke: The problem that started it all.

src/reason_pprint_ast.ml
```diff
@@ -99,16 +99,61 @@ let exprDescrString x =
...
-type listConfig = {
+(* Describes the "fixity" of a token, and stores its *printed* representation
+   should it be rendered as infix/prefix (This rendering may be different than
+   how it is stored in the AST). *)
+and tokenFixity =
+  (* Such as !simple_expr and ~!simple_expr. These function applications are
+     considered *almost* "simple" because they may be allowed anywhere a simple
+     expression is accepted, except for when on the left hand side of a
+     dot/send. *)
+  | AlmostSimplePrefix of string
```
jordwalke: We should get rid of this, and have only unary precedence in the future. Once `!` is swapped with boolean inversion `not`, it will make more sense anyways that `! myFunc a b` should be parsed like unary prefix `!(myFunc a b)`, instead of like today's "almost simple prefix" `(!myFunc) a b`.

src/reason_pprint_ast.ml
```diff
@@ -482,133 +527,239 @@ let rec sequentialIfBlocks x =
...
-      (fun s -> (`Right, s = "||"));
-    ];
-    [
-      (fun s -> (`Right, s = ":="))
-    ]
+  [
+    (TokenPrecedence, (fun s -> (Nonassoc, isSimplePrefixToken s)));
+  ];
+  [
+    (CustomPrecedence, (fun s -> (Nonassoc, s = "prec_unary_minus")));
```
jordwalke: Support for the custom `%prec` precedences in the yacc file!

src/reason_pprint_ast.ml
```diff
@@ -1992,19 +2140,15 @@ let formatCoerce expr optType coerced =
  *     );
  *
  *)
-let formatIndentedApplication delimiter headApplicationItem argApplicationItems =
+let formatIndentedApplication headApplicationItem argApplicationItems =
```
jordwalke: I was able to remove the delimiter, though I see @yunxing had recently added a new location field inside of the delimiter. The test cases seem to do the right thing, so maybe not so important?

src/reason_pprint_ast.ml
```diff
@@ -2084,16 +2215,10 @@ let protectIdentifier txt =
...
-let combinedAppItems (regularItems, sugaredItems, delimiter, loc) =
-    (regularItems @ sugaredItems, delimiter, loc)
-
 class printer  ()= object(self:'self)
   val pipe = false
   val semi = false
   (* The test and first branch of ternaries must be guarded *)
-  val inTernary = false
-  method underTernary = {<inTernary=true>}
-  method resetInTernary = {<inTernary=false>}
```
jordwalke: Thanks to the unparsing pseudo-automated rules, these are no longer needed. We could do the same thing for pipe/semi I imagine.

src/reason_pprint_ast.ml
```diff
@@ -2996,94 +3092,142 @@ class printer  ()= object(self:'self)
...
+     parens. It's simply a matter of changing [higherPrecedenceThan] to
+     [higherOrEqualPrecedenceThan].
+   *)
 
+  (* Ensures a rule doesn't reduce until *after* `reducesAfterRight` gets a chance
+     to reduce. Example: The addtion rule which has precedence of rightmost
+     token "+", in `x + a * b` should not reduce until after the a * b gets
+     a chance to reduce. This function would determine the minimum parens to
+     ensure that. *)
+  method ensureContainingRule ~withPrecedence ~reducesAfterRight =
```
jordwalke: These are the two main utilities that help "uparse".

src/reason_pprint_ast.ml
```diff
@@ -2996,94 +3092,142 @@ class printer  ()= object(self:'self)
...
+      that determines how to print based on precedence of tokens and rules.
+      The end goal is that this should be completely auto-generated from the
+      Menhir parsing tables. We could move more and more into this function.
 
-       let x = Red 10 20;
+      You could always just call self#expression, but `unparseExpr` will render
+      infix/prefix/unary/terary fixities in their beautiful forms while
+      minimizing parenthesis.
   *)
+  method unparseExpr x =
```
jordwalke: This is now the main entrypoint to printing expressions. It falls through to `self#expression` if the special "unparser" can't handle that kind of expression yet.

src/reason_pprint_ast.ml
```diff
@@ -2996,94 +3092,142 @@ class printer  ()= object(self:'self)
...
-            | None -> None
-      )
-      | _ -> None
+  method unparseExprApplicationItems x =
+    match self#unparseExprRecurse x with
+    | SpecificPrecedence ({reducePrecedence; shiftPrecedence}, leftRecurse) -> ([leftRecurse], Some x.pexp_loc)
+    | VeryHighPrecedence itms -> (itms, Some x.pexp_loc)
 
-  method consecutiveInfixApplicationItems x =
+  method unparseExprRecurse x =
```
jordwalke: This is now the main entrypoint to printing expressions. It falls through to `self#expression` if the special "unparser" can't handle that kind of expression yet.

src/reason_pprint_ast.ml
```diff
@@ -2996,94 +3092,142 @@ class printer  ()= object(self:'self)
...
-          | (`Infix infixStr, lbls) -> (None, [self#expression x])
-          | (`Prefix s, _) -> (None, [self#simple_expression x])
-          (* We know that this can only ever occur at a recursive call - the
-             first call into this would have been infix. There is overlap with
-             this and the simple_enough_to_be_separated_by_infix case. Either
-             path would work. *)
-          | (`Normal, _) -> (None, [self#expression x])
-        )
-      | _ -> (None, [self#simple_enough_to_be_separated_by_infix x])
+    | Pexp_apply (e, ls) -> (
```
jordwalke: Unparsing of function application, including the various potential forms (such as infix/prefix/sugary setting of array fields etc).

src/reason_pprint_ast.ml
```diff
@@ -2996,94 +3092,142 @@ class printer  ()= object(self:'self)
...
+          let attributesAsList = (List.map self#attribute stdAttrs) in
+          let theFunc = SourceMap (e.pexp_loc, (self#simple_expression e)) in
+          (*reset here only because [function,match,try,sequence] are lower priority*)
+          let theArgs = List.map self#reset#label_x_expression_param ls in
+          let theArgsAndAttrs = List.concat [theArgs; attributesAsList] in
+          VeryHighPrecedence (theFunc::theArgsAndAttrs)
+      )
+    )
+    (* TODO: Should protect this identifier *)
+    | Pexp_setinstvar (s, rightExpr) ->
```
jordwalke: And then the rest of this function is the unparsing of other syntax elements such as setting of fields/record updates, or ternary.

jordwalke:
There's one remaining issue:
Parenthesis are preserved in the first case (correctly), but not the second, which changes the meaning:

```rust
myFunc >>= (
  x y z >>= a  b c
);

myFunc >>= (
  fun x y z => 0 >>= fun a b c => 10
);
```

cristianoc:
Is this taking on some functionality of Easyformat, or is it purely about precedence?

As a meta-comment, the original printer is completely changed now, so we could also consider making more structural changes, e.g. get rid of the object methods for pretty printing and use normal functions instead.

jordwalke:
This isn't taking on any of the burden of Easy-format. It is mostly about precedence. And yeah, we can change the printer to use a bunch of mutually recursive functions. (perhaps once we merge the `4.03` PR it would be a good time to do that).

formatTest/typeCheckedTests/expected_output/imperative.re
yunxing: Why did you say this is not perfect? what else can be improved in your mind?

src/reason_pprint_ast.ml
```diff
@@ -482,133 +527,239 @@ let rec sequentialIfBlocks x =
...
+ *)
+let almost_simple_prefix_symbols  = [ '!'; '?'; '~'] ;;
+(* Subset of prefix symbols that have special "unary precedence" *)
+let unary_minus_prefix_symbols  = [ "~-"; "~-."] ;;
+let unary_plus_prefix_symbols  = ["~+"; "~+." ] ;;
 let infix_symbols = [ '='; '<'; '>'; '@'; '^'; '|'; '&'; '+'; '-'; '*'; '/';
                       '$'; '%'; '\\' ]
+(* There are just a couple of infix operators that start with `!`. All others that
+   start with ! would be prefix, so we keep a list of exceptions here. *)
+let infix_starting_with_bang = ["!="; "!=="]
```
yunxing: Do you really need this? "!=" and "!==" are already `special_infix_strings`

src/reason_pprint_ast.ml
```diff
@@ -482,133 +527,239 @@ let rec sequentialIfBlocks x =
...
+   may have resulted from Pexp -> Texp -> Pexp translation, then checking
+   if all the characters in the beginning of the string are valid infix
+   characters. *)
+let printedStringAndFixity  = function
+  | s when List.mem s special_infix_strings -> Infix s
+  | s when List.mem s.[0] infix_symbols || List.mem s infix_starting_with_bang -> Infix s
+  (* Correctness under assumption that unary operators are stored in AST with
+     leading "~" *)
+  | s when List.mem s.[0] almost_simple_prefix_symbols &&
+           not (List.mem s infix_starting_with_bang) &&
```
yunxing: This is redundant given the second match case.

src/reason_pprint_ast.ml
yunxing: Can't remember why I added the location info. Probably not important.

src/reason_pprint_ast.ml
```diff
@@ -2033,28 +2177,15 @@ let formatAttachmentApplication finalWrapping (attachTo: (bool * layoutNode) opt
...
           | (hd::tl, None) ->
             (* Args that are "attached to nothing" *)
-            let appList = makeAppList delimiter attachedList in
-            let attachedArgs =
-              match delimiter with
-                | None -> appList
-                | Some (s, loc) -> label ~space:true appList (atom s)
-            in
+            let appList = makeAppList attachedList in
+            let attachedArgs = appList in
```
yunxing: Duplicated.

src/reason_pprint_ast.ml
```diff
@@ -3099,18 +3243,19 @@ class printer  ()= object(self:'self)
...
+     associative on the S side of the function application, we don't need to
      wrap in parens. In more detail:
 
-     -Add parens
-       Exception 1:Unless we are a left-assoc X operator in the left branch of an X operator.
-       Exception 2:Unless we are a right-assoc X operator in the right branch of an X operator.
-       Exception 3:Unless we are a _any_-assoc X operator in the _any_ branch of an Y operator where X has greater precedence than Y.
+     -Add parens around infix binary function application
+       Exception 1: Unless we are a left-assoc operator of precedence X in the left branch of an operator w/ precedence X.
+       Exception 1: Unless we are a right-assoc operator of precedence X in the right branch of an operator w/ precedence X.
```
yunxing: typo

src/reason_pprint_ast.ml
```diff
@@ -2855,118 +2980,89 @@ class printer  ()= object(self:'self)
...
+
+  method simple_get_application e =
+    match (e.pexp_desc, e.pexp_attributes) with
+    | (_, attrHd::attrTl) -> None (* Has some attributes - not simple *)
+    | (Pexp_apply (eFun, ls), []) -> (
+      match (printedStringAndFixityExpr eFun, ls) with
+      (* We must take care not to print two subsequent prefix operators without
+         spaces between them (`! !` could become `!!` which is totally
+         different).  *)
+      | (AlmostSimplePrefix prefixStr, [("", rightExpr)]) ->
```
yunxing: Is it an invariant that a "AlmostSimplePrefix" always has its argument list as [("", rightExpr)]?

src/reason_pprint_ast.ml
jordwalke: I check it later on to make sure it's not interpreted as prefix - but I think I can just check special_infix_strings in that place. I'll try to remove this.

src/reason_pprint_ast.ml
jordwalke: Is it redundant? It seems to describe the exact opposite. I thought this was good because it makes each individual match case truthful on its own, and the entire set capable of being reordered without changing the meaning of the entire pattern match.

src/reason_pprint_ast.ml
jordwalke: fixing.

src/reason_pprint_ast.ml
jordwalke: No, and that would make a great test case. I will add it. For example, you can write the (almost simple) prefix application:

```rust
(!!!) 20 30;
```

formatTest/unit_tests/expected_output/infix.re
```diff
@@ -348,3 +368,586 @@ let myFunc
...
+    let result =
+      2 + (
+        - (- (- add 4 0 [@onAddApplication]))
+      );
+
+    /**
+     * TODO: Move all of these test cases to attributes.re.
+     */
+    /* Attribute on the prefix application */
+    let res = (- something blah blah) [@attr];
```
jordwalke: This is an important test case. See the difference between

`- something blah blah [@onSomethingCall]` and `(- something blah blah) [@onPrefixApplication]`.

formatTest/unit_tests/expected_output/infix.re
```diff
@@ -348,3 +368,586 @@ let myFunc
...
+     * And this
+     */
+    let res = !(- callThisFunc ());
+    /* Should be parsed (and should remain printed as: */
+    let res = !(- callThisFunc ());
+    let res = !x [@onApplication];
+    let res = !(x [@onX]);
+    let res = !(x [@onX]);
+    (something.contents = "newvalue")
+    [@shouldBeRenderedOnEntireSetField];
```
jordwalke: This is also another fix/improvement that the refactor made natural/easy. Before we were dropping attributes on some expressions such as `x.y = foo`.

formatTest/typeCheckedTests/expected_output/attributes.re
```diff
@@ -88,7 +88,7 @@ let x = 10 [@on10] - 20;
 
 let x = 10 [@on10] - 20;
 
-let x = (-) 10 20 [@attrEntireEverything];
```
jordwalke: I kind of liked it better in my first round of changes - if applying attributes to infix operations, then render the infix applications as normal-fix. Oh well, we'll revise attribute syntax anyways.

PR #703, opened 2016-08-20
Fix #688

### LINKED ISSUE #688: Polymorphic constraints not parsing
In `rtop` master:

```rust
type  extensible 'a =
  'a
  constraint 'a = [`Base int];
```

Does not parse. The same is true with `refmt`.

In OCaml, it does parse:

```ocaml
type  'a extensible =
   'a
   constraint 'a = [`Base int];;
```

This is a good first task: Fix the parse issue, and add a test case.

pvolok:
[Menhir doc](http://gallium.inria.fr/~fpottier/menhir/manual.pdf) says that `Parsing.ml` is deprecated whereas `symbol_rloc()` calls `Parsing.symbol_start_pos`.

But! ``type  extensible 'a = 'a constraint 'a = [`Base int];`` DOES work in refmt, and I have no idea why. RTop fails as expected.

There are other usages of deprecated ocamlyacc functions. I can remove them in this PR later today or tomorrow.

### DIFF
--- a/src/reason_parser.mly
+++ b/src/reason_parser.mly
@@ -90,12 +90,6 @@ open Syntax_util
      loc_ghost = true;
    }
 
-   let rhs_loc n = {
-     loc_start = Parsing.rhs_start_pos n;
-     loc_end = Parsing.rhs_end_pos n;
-     loc_ghost = false;
-   }
-
    ast_helper.ml:
    ------------
    module Typ = struct
@@ -483,8 +477,8 @@ let expecting_pat nonterm =
   else
     expecting nonterm
 
-let not_expecting pos nonterm =
-    raise Syntaxerr.(Error(Not_expecting(rhs_loc pos, nonterm)))
+let not_expecting start_pos end_pos nonterm =
+    raise Syntaxerr.(Error(Not_expecting(mklocation start_pos end_pos, nonterm)))
 
 let bigarray_function ?(loc=dummy_loc()) str name =
   ghloc ~loc (Ldot(Ldot(Lident "Bigarray", str), name))
@@ -527,10 +521,10 @@ let bigarray_set ?(loc=dummy_loc()) arr arg newval =
                         "", mkexp ~ghost:true ~loc (Pexp_array coords);
                         "", newval]))
 
-let lapply p1 p2 =
+let lapply p1 p2 start_pos end_pos =
   if !Clflags.applicative_functors
   then Lapply(p1, p2)
-  else raise (Syntaxerr.Error(Syntaxerr.Applicative_path (symbol_rloc())))
+  else raise (Syntaxerr.Error(Syntaxerr.Applicative_path (mklocation start_pos end_pos)))
 
 let exp_of_label label =
   mkexp ~loc:label.loc (Pexp_ident {label with txt=Lident(Longident.last label.txt)})
@@ -637,11 +631,11 @@ let mkcf_attrs ?(loc=dummy_loc()) d attrs =
 let mkctf_attrs d attrs =
   Ctf.mk ~attrs d
 
-let add_nonrec rf attrs pos =
-  match rf with
+let add_nonrec rf attrs =
+  match rf.txt with
   | Recursive -> attrs
   | Nonrecursive ->
-      let name = { txt = "nonrec"; loc = rhs_loc pos } in
+      let name = { txt = "nonrec"; loc = rf.loc } in
         (name, PStr []) :: attrs
 
 type let_binding =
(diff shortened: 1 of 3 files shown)


src/reason_parser.mly
```diff
@@ -2013,16 +2007,12 @@ parent_binder:
 
 value:
 /* TODO: factorize these rules (also with method): */
-    override_flag MUTABLE VIRTUAL as_loc(label) COLON core_type {
-      if $1 = Override then not_expecting 7 "not expecting equal - cannot specify value for virtual val"
-      else $4, Mutable, Cfk_virtual $6
-    }
```
pvolok: This case looks weird. Should equal sign be prohibited for any virtual field? If so, then the next case handles that.

src/reason_parser.mly
```diff
@@ -2283,7 +2273,10 @@ value_type:
       { $1, Immutable, Concrete, $3 }
 ;
 constrain:
-        core_type EQUAL core_type          { $1, $3, symbol_rloc() }
```
pvolok: This `symbol_rloc()` was causing the problem with constraint.

pvolok:
Replaced all deprecated position functions.

jordwalke:
> Menhir doc says that Parsing.ml is deprecated whereas symbol_rloc() calls Parsing.symbol_start_pos.

I can't find anywhere that confirms the official standard library has `Parsing` deprecated. I see it in the Menhir manual though.

src/reason_parser.mly
```diff
@@ -2283,7 +2273,10 @@ value_type:
...
+  | core_type EQUAL core_type {
+    let loc = mklocation $symbolstartpos $endpos in
+    $1, $3, loc
```
jordwalke: If you can remember to, I suggest forming tuples like `($1, $3, loc)` to make it more intuitive for noobs to understand what's going on (slightly closer to the Reason syntax too).

jordwalke:
> There are other usages of deprecated ocamlyacc functions. I can remove them in this PR later today or tomorrow.

Are you able to reproduce any of the `rtop` failures for those cases?

src/reason_parser.mly
jordwalke: Oh, I think the error message was just incorrect and misleading. I believe we should keep this case, and change the error message to say something like `"members marked virtual may not also be marked overridden."`

jordwalke:
This is an excellent find. Thank you! I still don't understand how `Parsing` is deprecated, but if this fixes the issue then great! If @yunxing has no suggestions or comments, we'll just merge (after you put back the error handling case for `virtual` `override` values).

pvolok:
Sorry, I meant `Parsing` is deprecated when used with menhir. Seems like menhir doesn't update this val: https://github.com/ocaml/ocaml/blob/520fb2d/stdlib/parsing.ml#L88

Usually position functions from Parsing should return the dummy position. But looking at [this line](https://github.com/ocaml/ocaml/blob/520fb2d/stdlib/parsing.ml#L192) I would say that it should throw all the time when `env.rule_len >= 2` (`end.asp` was 0 when I invistigated the issue). I'd prefer not to investigate any further as the doc says we should not use these functions with menhir anyway.

Another failed example: `type nonrec tt += ()`. utop points to the word `nonrec` but rtop doesn't (since `Parsing` returned dummy pos).

`virtual override` case is back.

jordwalke:
Thanks a ton, @pvolok !

yunxing:
@jordwalke Those parsing functions are deprecated in menhir in a way that they return dummy location. This is the main challenge when I converted the codebase to menhir.

@pvolok Great find. See some minor comments in the code.

src/reason_parser.mly
```diff
@@ -637,11 +631,11 @@ let mkcf_attrs ?(loc=dummy_loc()) d attrs =
 let mkctf_attrs d attrs =
   Ctf.mk ~attrs d
 
-let add_nonrec rf attrs pos =
+let add_nonrec rf attrs start_pos end_pos =
```
yunxing: There is a data structure called "loc" which encapsulate a object with it's location. 

If we can convert "rf" to "loc", we will be able to say
```
let add_nonrec rf attrs =
match rf.txt with 
...

let name = { txt = "nonrec"; loc = rf.loc }
...
```
which I found could make the code more concise.

I made a small utility to create a "loc" object, see comments below.

src/reason_parser.mly
yunxing: Great detective work.

src/reason_parser.mly
```diff
@@ -3459,7 +3457,7 @@ many_type_declarations:
       let loc = mklocation $symbolstartpos $endpos in
       [Type.mk ident
        ~params:params ~cstrs:constraints
-       ~kind ~priv ?manifest ~attrs:(add_nonrec $2 $4 2) ~loc]
+       ~kind ~priv ?manifest ~attrs:(add_nonrec $2 $4 $startpos($2) $endpos($2)) ~loc]
```
yunxing: In the grammar rule above, we can say:
```
| TYPE as_loc(nonrec_flag) type_declaration_details post_item_attributes
```
This turns nonrec_flag into a "loc" object, and we can then call it with
```
~kind ~priv ?manifest ~attrs:(add_nonrec $2 $4) ~loc]
```

pvolok:
@yunxing Thanks for such detailed explanation! Updated.

yunxing:
Looks good. Merging.

PR #704, opened 2016-08-21
Fix record spread with no labels.

### LINKED ISSUE #689: Invalid record spread syntax
```
{...foo};
```

This is invalid syntax. `[...bar]` works and prints as plain `bar`; for consistency, `{...foo}` should print as `foo`.

pvolok:
Fix #689

### DIFF
--- a/formatTest/unit_tests/expected_output/basicStructures.re
+++ b/formatTest/unit_tests/expected_output/basicStructures.re
@@ -539,6 +539,8 @@ let anotherRecord = {
   age: testRecord.age + 10
 };
 
+let copiedRecord = {...testRecord};
+
 let makeRecordBase () => {
   name: "Joe",
   age: 30,
--- a/formatTest/unit_tests/input/basicStructures.re
+++ b/formatTest/unit_tests/input/basicStructures.re
@@ -448,6 +448,9 @@ let anotherRecord = {
   name: "joe++",
   age: testRecord.age + 10
 };
+let copiedRecord = {
+  ...testRecord
+};
 
 let makeRecordBase () => {name: "Joe", age: 30, occupation: "Engineer"};
 let anotherRecord = {
--- a/src/reason_parser.messages
+++ b/src/reason_parser.messages
@@ -6723,6 +6723,7 @@ implementation: LBRACE DOTDOTDOT UIDENT RBRACKET
 ##
 ## Ends in an error in state: 1112.
 ##
+## record_expr -> DOTDOTDOT expr_optional_constraint . [ error RBRACE ]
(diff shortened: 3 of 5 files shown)


yunxing:
Looks great, thanks!

chenglou:
What happens when there's a single mutable field?

jordwalke:
It appears to give a compiler error because record fields cannot have zero fields extended.

```
Reason # let x = {contents: 10};
let x : ref int = {contents : 10}                                                                                                                                                                                                                                               Reason # x.contents = 100;
- : unit = ()                                                                                                                                                                                                                                                                   Reason # x;
- : ref int = {contents : 100}                                                                                                                                                                                                                                                  Reason # let y = {...x};
Error: broken invariant in parsetree: Records cannot be empty.
```

pvolok:
Forgot to test. Sorry...

I'll try to hack on ocaml.

jordwalke:
I think the change to ocaml might be somewhat difficult since the type cannot be inferred without one label to work off of.

```rust
type t = {someField: int};
let myFun y => {...y, someField: foo};
let myFun y = {...y};
```

The first function will be inferred as `t=>t` because the record label gives enough information to work from, but the second would need to be inferred as `anyRecordType => anyRecordType`, but OCaml does not currently have any notion of "any type that is a record".

In other words, it would be neat to have this feature, but I'm not sure it will be worth the benefit (compared to other things we could be building). (Unless you already know how to implement this in the OCaml core compiler).

jordwalke:
So I'd say this diff is fine as it is, but we could make a slightly better error message perhaps.

pvolok:
No, I'm new to the ocaml world. So yeah, you talked me out of it for now. Hopefuly will dig it one day :)

Should we fail in the parser, but with a better message?

yunxing:
Found some discussion about this issue before: http://caml.inria.fr/pub/ml-archives/caml-list/2005/01/9deef19c1b8dde84b81b554a7caa482e.en.html, which is basically what @jordwalke mentioned. 

I think there are two options for us now:
  1) Print `{...bar}` into `bar` -- this may confuse people as people may expecting the syntax {...bar} to make a shallow copy of `bar`. 
  2) Make a better syntax error when people write `{...bar}`.

pvolok:
If we autoformat `{...bar}` into `bar`, it is not only confusing, but can also introduce bugs. Why would someone write `{...bar}` rather than trying to make a shallow copy?

yunxing:
@pvolok Sorry just saw this. Maybe we here we should just provide a better syntax error here.

jordwalke:
Agree.

Issue #707, opened 2016-08-23
[parser] Nested object type error


chenglou:
Repro:

```
type foo = <
bar : <a: int>>;
```

This works:

```
type foo = <
  bar : <a: int>
>;
```

pvolok:
The problem is that `>>` is being lexed as a token `INFIXOP0(">>")` while the parser expects `GREATER GREATER`.

One idea I had is to lex `>...` to something like `GREATER_WITOUT_SPACE_AFTER ...` and treat it as an operator beginning when we expect operator. But this would also require changing precedence rules.

I will not tackle this issue now as I first need to learn menhir and ocaml operators better. So posted it in case someone takes it.

yunxing:
Another idea is we can build some utilities in our parse to "split" a token into two:
In the lexer, 
  when we see ">>", we tokenize it as two tokens ["GREATER", "GREATER"]. 

In the parser when we see, 
  when we see GREATER GREATER, we reduce them to INFIXOP0(">>"). 

The benefit of doing this is we don't need create fake tokens like "/><" in JSX. As it will be split as ["/", ">", "<"] correctly. Do you think this can make your JSX diff easier ? @SanderSpies @jordwalke

yunxing:
C++03 actually has the same problem, which took people a while to fix: https://en.wikipedia.org/wiki/C%2B%2B11#Right_angle_bracket

jordwalke:
Yeah, Ideally we'd split the tokens into two when possible.

In this case of objects, I was hoping that when we revised the method dispatch syntax, we could also improve the object type syntax. Right now, Reason improved the "class type" syntax but not the object type syntax.

Taking a step back: OCaml has a syntax for object types (which we're seeing here with the `<>`) and a syntax for class types (which look like `object method blah: int -> int end`). In Reason, we've improved the syntax for class types (which you can see in our `oo.re` test file) so that it looks like:

```rust
 {method x : int; method y : int;}
```

I think that we can continue to not only improve that syntax, but also make it so that *object type* syntax is identical. I don't know why OCaml has different syntax for class types vs. regular object types and it's part of the reason why I think no one uses objects in OCaml. After doing some deep exploration of the syntax of objects, I think it's possible to unify class types and object types *syntactically* while always being able to keep the distinction in the AST, once parsed.

All that to say - I think the `>>` issue would go away, but we should still consider breaking it into multiple tokens if possible.

jordwalke:
Alternative proposals:

```rust
[...]

PR #708, opened 2016-08-23
support nested object type

yunxing:
One possible solution for #707.


SanderSpies:
Ah, this is what you meant - looks interesting. I think this approach could definitely help with making JSX easier.

yunxing:
Rebased against master.

PR #395, opened 2016-05-02
Use cmdliner

fxfactorial:
Implement replacement of Arg with cmdliner while passing all tests for `refmt_imp` along with some style changes.

@jordwalke You'll probably want to commit on top of this to replace the proper description in the man page's main `DESCRIPTION`.

PR #709, opened 2016-08-23
Better help message for refmt

jaredly:
- added example usages
- show help if no filename given
- show help if `-h` given

Test Plan:
`make build && ./_build/bin/refmt -h`

```
:~/clone/rsn/Reason$ ./_build/bin/refmt -h hi
Reason: Meta Language Utility

[Usage]: refmt [options] some-file.[re|ml]

   // translate ocaml to reason on stdin
   echo 'let _ = ()' | refmt -print re -parse ml -use-stdin true -is-interface-pp false

   // print the AST of a file
   refmt -parse re -print ast some-file.re

   // reformat a file
   refmt -parse re -print re some-file.re

[Options]:

  -ignore ignored
  -is-interface-pp <interface>, parse AST as <interface> (either true or false)
  -use-stdin <use_stdin>, parse AST from <use_stdin> (either true, false). You still must provide a file name even if using stdin for errors to be reported
  -recoverable Enable recoverable parser
  -assume-explicit-arity If a constructor's argument is a tuple, always interpret it as multiple arguments
  -parse <parse>, parse AST as <parse> (either 'ml', 're', 'binary_reason(for interchange between Reason versions')
  -print <print>, print AST in <print> (either 'ml', 're', 'binary(default - for compiler input)', 'binary_reason(for interchange between Reason versions)', 'ast (print human readable directly)', 'none')
  -print-width <print-width>, wrapping width for printing the AST
  -heuristics-file <path>, load path as a heuristics file to specify which constructors are defined with multi-arguments. Mostly used in removing [@implicit_arity] introduced from OCaml conversion.
                example.txt:
                Constructor1
                Constructor2
  -h  Display this list of options
  -help  Display this list of options
  --help  Display this list of options
:~/clone/rsn/Reason$
```

### DIFF
--- a/src/refmt_impl.ml
+++ b/src/refmt_impl.ml
@@ -35,6 +35,22 @@ let reasonBinaryParser use_stdin filename =
   let (magic_number, filename, ast, comments, parsedAsML, parsedAsInterface) = input_value chan in
   ((ast, comments), parsedAsML, parsedAsInterface)
 
+let usage =
+"Reason: Meta Language Utility\n\
+\n\
+[Usage]: refmt [options] some-file.[re|ml]\
+\n\
+\n   // translate ocaml to reason on stdin\
+\n   echo 'let _ = ()' | refmt -print re -parse ml -use-stdin true -is-interface-pp false\
+\n\
+\n   // print the AST of a file\
+\n   refmt -parse re -print ast some-file.re\
+\n\
+\n   // reformat a file\
+\n   refmt -parse re -print re some-file.re\
+\n\
+\n[Options]:\n"
+
 (*
  * As soon as m17n vends comments, this should be replaced with what is
  * effectively m17n's parser.
@@ -49,7 +65,8 @@ let () =
   let assumeExplicitArity = ref false in
   let heuristics_file = ref None in
   let print_width = ref None in
-  let () = Arg.parse [
+  let print_help = ref false in
+  let options = [
     "-ignore", Arg.Unit ignore, "ignored";
     "-is-interface-pp", Arg.Bool (fun x -> intf := Some x), "<interface>, parse AST as <interface> (either true or false)";
     "-use-stdin", Arg.Bool (fun x -> use_stdin := x), "<use_stdin>, parse AST from <use_stdin> (either true, false). You still must provide a file name even if using stdin for errors to be reported";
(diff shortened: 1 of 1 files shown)


jordwalke:
Also, see this PR from @fxfactorial who showed us how to use cmdliner for beautiful help messages automatically: https://github.com/facebook/reason/pull/395

src/refmt_impl.ml
```diff
@@ -35,6 +35,22 @@ let reasonBinaryParser use_stdin filename =
   let (magic_number, filename, ast, comments, parsedAsML, parsedAsInterface) = input_value chan in
   ((ast, comments), parsedAsML, parsedAsInterface)
 
+let usage =
```
jordwalke: Have you seen strings like this?

```ocaml
let x = {|
   Hey look it's a string that makes embedding newlines very easy!
   See?
|};

Comment (jordwalke) [inline src/refmt_impl.ml]: If you want to change to the sweet multiline form, I'll merge this immediately after.

Comment (jaredly) [inline src/refmt_impl.ml]: that's exactly what I wanted!

**Comment by yunxing** [comment]
Nice, let's wait for CI (which is taking forever today) and merge this.

**Comment by jaredly** [comment]
also what is up with the `-ignore` argument? It's literally unused.

**Comment by jordwalke** [comment]
It's, ignored :)

File: src/refmt_impl.ml [hunk 2]
```diff
@@ -62,11 +79,16 @@ let () =
...
-  (fun arg -> filename := arg)
-  "Reason: Meta Language Utility"
-  in
+    "-h", Arg.Unit (fun () -> print_help := true), " Display this list of options";
+  ] in
+  let () = Arg.parse options (fun arg -> filename := arg) usage in
+  let print_help = !print_help in
   let filename = !filename in
+  let _ =
+    if filename = "" || print_help then
```
yunxing: The filename will be "" if we pass -use-stdin. 

Should check if that option is set or not here.

fxfactorial:
FWIW, my code can still be used. The biggest hassle was all the little shell scripts and whatnot.

yunxing:
@fxfactorial Long time no see @fxfactorial ! Could you elaborate a bit on that? I tried to checkout your code and run tests, but didn't see any failure.

fxfactorial:
@yunxing (sorry, I moved overseas and trying to build a hackerspace in Armenia, introduce people to ReactJS & friends.)

I mean to say that my original PR can still work, essentially. The things that are/were breaking are the shell scripts in the tests directory, last time I checked. 

Bonus points, you get a free and real man page.

jaredly:
@jordwalke ok I think it's ready?

yunxing:
Looks good. Was waiting for CI to complete. Let's look at if we can get @fxfactorial's stuff in.

PR #711, opened 2016-08-24
Change `let open` to `open` (fix #641).

### LINKED ISSUE #641: change `let open Foo` to be `open Foo`.
I think the `let` is excessive, let's just use `open Foo` even when opening locally in scope.

ak3n:


yunxing:
Thanks man. Diff looks good to me. Let's wait for the CI to complete.

ak3n:
Not sure... but looks like I broke travis somehow.

screen shot 2016-08-25 at 02 52 48

yunxing:
Our CI looks pretty unstable recently. Let me see what I can do.

chenglou:
Precision: travis CI runs are queued per organization. So the entire facebook/ runs on the same queue. This is busy during the day. There might be some repos that introduced expensive builds recently.

ak3n:
@chenglou it makes sense (and a little bit sad). Thanks for info.

chenglou:
Thanks!

PR #713, opened 2016-08-25
Change the precedence of ternary expression (fixes #642).

### LINKED ISSUE #642: "Ternary" not parsed correctly
I discovered the following example is not parsed correctly.

```
 something.contents = isUndefined (childHeight.contents) ? X : Y;
```

Is parsed as.
```
 (something.contents = isUndefined (childHeight.contents)) ? X : Y;
```

Which I suppose makes sense as a lexical interpretation of the statement, but it's not intuitive, and not what you would expect. It's also never a valid program (the result of an imperative assignment is `unit`, which could never be `true/false`).

ak3n:


ak3n:
Ooops. Have not seen #702.

Issue #710, opened 2016-08-24
Class type ppx attributes disappear


chenglou:
Repro ocaml code:

``` ocaml
class type _y = object 
  method height : int [@@bs.set {no_get}]
  (* [height] is setter only *)
end [@bs]
```

No attributes when converted. Related: #693

yunxing:
How to attack this problem for first time contributor:
- Install reason according to the [readme page](git@github.com:facebook/reason.git).
- create a **ocaml** file test.ml, with the content 
```
class type _y = object 
  method height : int [@@bs.set {no_get}]
end [@bs]
```
- run `refmt -parse ml -print re try.ml`, notice that the [@bs] attributes are disappeared from the output.

You job is to change "reason_pprint_ast.ml" so the conversion preserves the attributes.

ngzhian:
Hi! I would like to claim this issue :)
I have started investigating, and found the old closed PR that fixes this. However this causes indentation issues.

I printed the intermediate easy-format representation and found there because of the fix, there is a `attach_std_attrs` call that wraps it in another list. I have yet to figure out how to fix this, but would like to jot down my findings here. Thanks!

This is the fixed output:
```
 List: open  close  sep  break Force_breaks_rec 
   List: open  close  sep  break Force_breaks_rec 
     Atom:''
     Label (break = Auto): 
       left 
       List: open  close  sep  break No_breaks 
         Label (break = Auto): 
           left 
           Atom:'class type'
           right 
           Atom:'_y'
         Atom:'='
       right 
       List: open  close  sep  break No_breaks  <<< here is the extra enclosing List
         List: open { close } sep  break Force_breaks_rec 
           Label (break = Auto): 
             left 
             Label (break = Auto): 
               left 
               Atom:'pub'
               right 
               List: open  close  sep  break Never_wrap 
                 Atom:'height:'
             right 
             List: open  close  sep  break No_breaks 
               Atom:'int'
         List: open  close  sep  break Never_wrap 
           List: open  close  sep  break No_breaks 
             Atom:'[@bs]'
             Atom:';'
```

This is the original output
```
 List: open  close  sep  break Force_breaks_rec                                                                                                                                                           [2/69]
   List: open  close  sep  break Force_breaks_rec 
     Atom:''
     Label (break = Auto): 
       left 
       List: open  close  sep  break No_breaks 
         Label (break = Auto): 
           left 
           Atom:'class type'
           right 
           Atom:'_y'
         Atom:'='
       right 
[...]

PR #716, opened 2016-08-26
Preserve attributes on class types

IwanKaramazow:
See https://github.com/facebook/reason/issues/710

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/attributes.re
+++ b/formatTest/typeCheckedTests/expected_output/attributes.re
@@ -319,3 +319,28 @@ let myFun
    */
 /* Bucklescript FFI item attributes */
 external imul : int => int => int = "Math.imul" [@@bs.val];
+
+class type _x = {
+                  method height: int
+                } [@bs];
+
+class type _y = {
+                  method height: int [@@bs.set]
+                } [@bs];
+
+class type _z = {
+                  method height: int
+                  [@@bs.set] [@@bs.get]
+                } [@bs];
+
+class type _A = {
+                  method height: int
+
+                    [@@bs.set
+                      {
+                        no_get;
+                        null
+                      }
+                    ]
+
+                } [@bs];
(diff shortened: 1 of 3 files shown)


SanderSpies:
Is it possible to make the output more like: https://github.com/IwanKaramazow/reason/blob/c218d8bd507141e020e19aac4af27538951589ff/formatTest/typeCheckedTests/expected_output/attributes.re#L231 ?

formatTest/typeCheckedTests/expected_output/attributes.re
```diff
@@ -319,3 +319,28 @@ let myFun
    */
 /* Bucklescript FFI item attributes */
 external imul : int => int => int = "Math.imul" [@@bs.val];
+
+class type _x = {
+                  method height: int
```
yunxing: The indentation seems to be incorrect here.

jordwalke: In most cases `~inline:(true, false) ~wrap:("{", "}")` will fix it.

wokalski: @jordwalke This bug is introduced when `[@bs]` attribute is [appended](https://github.com/facebook/reason/pull/716/files#diff-7dce44be99776d19af8c2d7b0d14e2c7R4737). It makes braces to be aligned on one column. 

I know nothing about this codebase, but after experimenting a little bit, it seems like the problem is somewhere else. I think that brace indentation logic is causing it but cannot figure out where it happens. Any tips appreciated.

jordwalke: by default, lists are `break:Never`. It could be that some list needs to specify `~break:IfNeed`. Other than that, the lists that append to labels *typically* need to be `~inline:(true, false)` to wrap in pleasing ways. So look around `class_types` in the printer to see if any lists and/or labels sound like they could be the culprit.

Issue #660, opened 2016-08-02
/* let a = [%bs.obj {hi: a" [@bs]}]; */ fails to parse


chenglou:
cc @yunxing

jordwalke:
The string is not balanced inside of the comment, correct?

chenglou:
Right. I keep forgetting ocaml tries to parse what's inside...

yunxing:
@chenglou Interesting. I didn't ocaml tries to parse what's inside.

chenglou:
Nested comments work

jordwalke:
It's actually reason lexer that is doing this - forked from OCaml's.

jaredly:
the PR was abandoned b/c it would break ocaml compat

PR #717, opened 2016-08-26
Remove string quotes checking inside comments

yunxing:
#660


jordwalke:
Will this fail to parse if you convert this back to OCaml syntax (assuming you can preserve comments?)

jordwalke:
I wish this could still be checked but changed to a warning which could be suppressed.

yunxing:
@jordwalke You are right, this is not compatible with the original OCaml parser.

yunxing:
abandoning this as it won't be compatible with ocaml.

PR #718, opened 2016-08-26
Update package.json to include utop

### LINKED ISSUE #667: Lighter weight top level.
`rtop` has a ton of dependencies. There is a custom top level inside of `reason` which is not `rtop`/`utop`, but the simply default top level that parses using Reason.

Our `npm` version of `reason` doesn't even include `rtop` because it has too many dependencies which all need to be compiled at installation, and because we'd have to port them all to npm.
We should take that minimal top level and make it usable and include it with the `npm` installation.

I believe the following work needs to be done.

- Simply wrap it with `rlwrap` (and model `rlwrap` as an npm dependency (actually, there is likely already a package).
- Ensure it prints outcomes using KC's custom outcome printer for Reason.

yunxing:
This should fix #667

### DIFF
--- a/opam
+++ b/opam
@@ -22,7 +22,6 @@ build-test: [
 depends: [
   "easy-format"  {>= "1.2.0"}
   "ocamlfind"    {build}
-  "ounit"        {test}
   "utop"         {>= "1.17"}
   "BetterErrors" {>= "0.0.1"}
   "menhir"       {>= "20160303"}
--- a/package.json
+++ b/package.json
@@ -1,47 +1,65 @@
 {
-  "name": "reason",
-  "version": "1.2.1",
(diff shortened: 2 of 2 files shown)


jordwalke:
Wow, it's working already!?

I saw you ported all of these [packages](https://github.com/npm-opam/), but there's no git history, should we keep the git history and just tag certain commits for versions?

yunxing:
@jordwalke Unfortunately we don't have a history for the automated conversion -- the source code are obtained via reading opam files (for example, the "source" field in https://opam.ocaml.org/packages/aifad/) and the git repo is recreated on top of that. 

We can definitely tag the commits though. I was using master before for the ease of hacking (so I can just go to github and edit the file on their web UI).

PR #723, opened 2016-08-27
Add Haskell-like infix syntax (fix #401).

### LINKED ISSUE #401: [ideas] Having haskell infix syntax
It would be generally useful to allow:

``` ocaml
a `infix_fun` b
```

with this, there is no need to make `lsl`, `lsr` and `mod` keyword

ak3n:


jordwalke:
We were thinking of using \` to represent multiline strings, instead of the current approach, which is to inherit OCaml's `{| |}` multiline syntax. One justification for backticks for multiline strings, is that they will then render correctly when opening in source mapped browser debuggers.

screen shot 2016-08-27 at 3 59 24 pm


So, I imagine you could use another leading character perhaps. Is there a lot of demand for arbitrary text identifiers to be parsed as infix? If so, then I imagine you'd also want to parse the identifier with a ppx attribute attached, and then use that attachment to keep the infix identifier as rendered infix when it comes time to print (in this PR, you'll see the output is rendered not-infix because the fixity wasn't preserved anywhere - ppx attributes can be used to solve that).

ak3n:
What do you think about ``` 1 ``add`` 2 ```?

Thanks for suggestion about ppx attribute. Good idea.

yunxing:
@ak3n I miss this a lot, thanks. 

You can also try something like `1 ~add 2`, not sure if that would work, though. 

For ppx attributs, take a look at the function "simple_ghost_text_attr" to create an attribute and pass the result to mkexp in your diff.

freebroccolo:
I'd kind of like `x ~op y` syntax too if it's possible. Is `~` being used for anything else now?

yunxing:
@freebroccolo `~` is currently used to downgrade a binary operator into a unary operator ("~-" means the negate operator)

src/reason_pprint_ast.ml
```diff
@@ -697,6 +699,16 @@ let rules = [
   [
     (TokenPrecedence, (fun s -> (Nonassoc, s = "=>")));
   ];
+  [
+    (TokenPrecedence, (fun s ->
+      let is_infix_op = Hashtbl.mem infix_operators s in
```
yunxing: Just curious why do you have this mechanism to check if a infix_operator is registered or not. Why not just do
(TokenPrecedence, (fun s -> (Left, s.[0] = "~"))); ?

ak3n: I thought about similar to Haskell's `infix`, `infixl` and `infixr` mechanism. To allow user to choose associativity of operator.

yunxing: Maybe useful, but how do we allow user to choose the associativity? AFAIK Haskell uses a two-passes algorithm which determines associativity and precedence in the first pass.

ak3n: Something like:

```
infixr add
...
let add a b => a + b;
```

The default choice is left associativity.

yunxing: The puzzle is now we have `%left VARIABLEINFIXOP` in the parser, saying this is a left-assoc operator.

With your proposal we have to somehow tell the parser to parse it as %right VARIABLEINFIXOP. Not sure how to do it with Menhir (our parser generator).

ak3n: Oh, sorry, forgot about it. Ok, I will remove it.

src/reason_parser.mly
```diff
@@ -782,6 +782,21 @@ let rec string_of_longident = function
...
 
+(* Replace dot with bangdot
+   e.g.: "Int32.rem" ~> "Int32!.rem" *)
+let prepare_name_for_call name =
+  if String.contains name '.' then
+    let rec replace str n = match str with
+        | "" -> ""
+        | str ->
+          let ch = (String.get str 0) in
+          let new_str = if Char.compare ch '.' == 0 then "!." else String.make 1 ch in
```
yunxing: Not very familiar with the logic with Longident.parse -- what's the magic here to replace "." with "!."?

formatTest/unit_tests/expected_output/basicStructures.re
```diff
@@ -635,3 +635,9 @@ let anotherRecord = {
   name: "joe++",
   age: testRecord.age + 10
 };
+
+let add a b => a + b;
+
+(15 ~add 4) ~add 5;
```
yunxing: Do you know how to remove the unnecessary parens here?

src/reason_parser.mly
ak3n: Ooops. It's useless function, which I wrote before Longident.parse.

src/reason_parser.mly
```diff
@@ -2705,6 +2707,13 @@ _simple_expr:
         let loc = mklocation $symbolstartpos $endpos in
         bigarray_get ~loc $1 $4
       }
+  | simple_expr VARIABLEINFIXOP simple_expr
```
yunxing: You are making a rule of
```
simple_expr:
  | simple_expr VARIABLEINFIXOP simple_expr
```
`simple_expr` is not a right item here. Simple expression means a special kind of expression where you can just pass it as a function argument directly without wrapping them into parens. But that's not the case with infix operators.

We should put this under `expr`.

src/reason_parser.mly
```diff
@@ -932,6 +933,7 @@ conflicts.
 
 */
 /* Question: Where is the SEMI explicit precedence? */
+%left VARIABLEINFIXOP
```
yunxing: Order matters here. 

putting VARIABLEINFIXOP as first item means this operator has lowest precedence. From what you are doing in the printer and what Haskell uses, I think we need it to be the highest precedence.

src/reason_parser.mly
ak3n: Any variant with `expr` has conflicts.

```
Error: 47 states have shift/reduce conflicts.
Error: 29 states have reduce/reduce conflicts.
Warning: 823 shift/reduce conflicts were arbitrarily resolved.
Warning: 1173 reduce/reduce conflicts were arbitrarily resolved.
```

yunxing: what's your grammar rule there?

Try add this under "expr as(loc(infix_operator) expr":
```
  | expr VARIABLEINFIXOP expr
      {
        let n = prepare_name_for_call $2 in
        let attribute = simple_ghost_text_attr "infix" in
        let name = mkexp ~ghost:true ~attrs:(attribute) (Pexp_ident(ghloc (Longident.parse n))) in
        mkexp (Pexp_apply (name, ["", $1; "", $3]))
      }
```
Notice we have "mark_position_exp" after expr so you don't need to calculate ~loc and pass is as an argument to mkexp.

ak3n: Nice!

formatTest/unit_tests/expected_output/basicStructures.re
ak3n: They are added here https://github.com/facebook/reason/blob/master/src/reason_pprint_ast.ml#L3118

src/reason_parser.mly
```diff
@@ -2560,6 +2563,12 @@ _expr:
       }
   | expr as_loc(infix_operator) expr
       { mkinfix $1 $2 $3 }
+  | expr VARIABLEINFIXOP expr
+      {
+        let attribute = simple_ghost_text_attr "infix" in
+        let name = mkexp ~ghost:true ~attrs:(attribute) (Pexp_ident(ghloc (Longident.parse $2))) in
```
yunxing: You need to pass location of the second argument to both mkexp here and ghloc here. Otherwise this will confuse our IDE support. 

Take a look at how mkinfix does it.

src/reason_parser.mly
```diff
@@ -1053,6 +1054,8 @@ conflicts.
           NEW NATIVEINT PREFIXOP STRING TRUE UIDENT
           LBRACKETPERCENT
 
+%left VARIABLEINFIXOP
```
yunxing: hmm.. Maybe the precedence should be lower than LPAREN

src/reason_parser.mly
ak3n: Something like this?

```
  | expr as_loc(VARIABLEINFIXOP) expr
      {
        let attribute = simple_ghost_text_attr "infix" in
        let name = mkexp ~ghost:true ~attrs:(attribute) ~loc:($2.loc)
          (Pexp_ident(ghloc ~loc:($2.loc) (Longident.parse $2.txt))) in
        mkexp (Pexp_apply (name, ["", $1; "", $3]))
      }
```

src/reason_parser.mly
jordwalke: I believe it should be right by the other infix ops (you have to decide if it's higher or lower precedence than `INFIXOP4` etc.

src/reason_parser.mly
yunxing: Right

src/reason_pprint_ast.ml
```diff
@@ -697,6 +697,9 @@ let rules = [
   [
     (TokenPrecedence, (fun s -> (Nonassoc, s = "=>")));
   ];
+  [
+    (TokenPrecedence, (fun s -> (Left, s.[0] == '~')));
```
jordwalke: The order in this list should correspond to the ranking in precedence from `reason_parser.mly`. Notice how it corresponds for things like `&&` and `||`. The new infix `~` should similarly be represented.

ak3n:
Should I fix pretty printer to get applications without parentheses? Like `10 ~add 4 ~add 5`?

yunxing:
@ak3n Could be a more complex problem. Maybe just open an issue once this is merged?

yunxing:
Just found one thing that could be adding the parens, if you search on "printedStringAndFixity"

ak3n:
Is list of unary operators fixed? We can union `unary_minus_prefix_symbols`, `unary_plus_prefix_symbols` and other unary operators to get a list for filtering in `printedStringAndFixity`.

ak3n:
Or we can keep a set of infix operators...

bsansouci:
I don't think there is much value in arbitrary infix operators.

I can't think of an example where a word carries more "commonly accepted meaning" than a symbol. If I compare `->>|` with a name like `map`, I don't find that `map` carries more meaning. In both cases I need to know what that function does given the specific context. (A counter example would be `+++` vs `dot_product` but I'd argue that `dot_product v1 v2` isn't much worse than `v1 ~dot v2` because it avoids the precedence issues like `v1 ~dot v2 + v3`).

Another reason that I don't think such freedom is useful is that most of the infix operators that I know of are for math and math doesn't need `add` because `+` is pretty darn universal. Monad composition could very well be done with normal function call and clever uses of `|>`.

Lastly, operators have precedence built into them. I'll agree that figuring out the precedence of a user defined operator might be unwieldy but it's at least possible. It doesn't seem to be the case with literal identifiers like here.

yunxing:
@bsansouci Thanks for the input. I totally agree that it is more intuitive to represent a mathematical operator in the symbols we are familiar with. However the problem is in ocaml, without modular implicit, we have to come up with a different symbol for the same operation across different data structures....

A classical example is "+". We have "+" for integers, "+." for floats. But how about for other [Magma](https://en.wikipedia.org/wiki/Magma_(algebra)) types? As a example, in compiler lib we have a data structure called "loc" that acts like intervals and we would like to have an operation that "combines" two intervals together. Which symbol should we use? "+" and "+." are already take, we can use "+.." or "+!", how about "++"? It just quickly becomes unmanageable. In cases like that, I would actually prefer to have an English word as the name of the operator.

yunxing:
@hakuch Good point. We do have the same support in Reason. Not sure about what the best practice is here. Do people do this often? My concern is this may shadow Pervasive and limit other functions in the module to write (+).

```
module Foo = struct
  (** Add a new operator for string concatenation. *)
  let ( + ) x y =
    x ^ y
  (** Another API)
  let plusOne x = x Pervasive.(+) 1 
end
```

hakuch:
I do this often in OCaml code, and many libraries (such as `cmdliner`) do as well. It's particularly helpful for monadic (and other such common) operators like `>>=`. For example, I might have a `Parser` module that has a monadic interface (with `unit` and `bind` functions), and then a module `Parser.Infix` that exposes `>>=`.

bsansouci:
@yunxing Part of my point was that english words wouldn't necessarily be more meaningful than `++`, because either way it's something that isn't common and the user of the library has to lookup what the symbol or word means.

freebroccolo:
I use this feature frequently in Haskell and would have used it on a number of occasions if it were available in OCaml.

I don't see this feature as having anything to do with "which way is more meaningful". It's to let you use functions which are already defined somewhere conveniently as operators either on a limited basis or when usual operators are not enough or too cumbersome for some reason.

For example:

1. Writing functions in infix form can make certain nested expressions easier to read. Think DSLs or combinators for constructing certain kinds of syntax trees. Sometimes writing everything in prefix form is harder to understand.

2. It's often simply not worth defining an operator if it is only used in a few places locally.

3. OCaml's operator functionality is already fairly inflexible and limited in usefulness due to the way operator symbol prefix determine fixity. If you're trying to write a DSL, this becomes a problem once you hit around 4 or more primary operators with different precedence.

I can't speak for others but in my own code I would not redefine operators like `(+)` or `(*)` locally unless they had a very specific meaning as some sort of algebra because I think it's too confusing. Even then, I wouldn't really like doing that until we have modular implicits. I agree that it can make more sense for some special cases like `(>>=)` but I am more skeptical about the general case.

So I don't really view that as an alternative to this proposal in general.

Another issue with the local operator approach is that you can't (afaik) have qualified operators. So if you have two different modules `M0` and `M1` with `(+)`, you aren't going to be able to open and use them both in the same scope. But `~M0.add` and `~M1.add` shouldn't be a problem. Qualified operators would also be nice to have but that's another issue.

tekknolagi:
Is there by chance any motion on this?

Issue #706, opened 2016-08-22
Parse into bleeding edge "modern" AST data structure. Downgrade automatically.


jordwalke:
### How can we use new OCaml features immediately?

Here's what we can do to start using all of the upcoming and recently released features `Ocaml` features in `Reason` the moment they are available in `ocaml`.
- Create a package called `MLAst` which only models the AST. It is independent of the compiler but for the most part is the exact same dat structure as the most recent version of the OCaml AST. The only difference is that it would have syntactic support for algebraic effects, and modular implicits in the AST, simultaneously, even though neither is in upstream ocaml yet (we don't need to add parsing rules for them in the first step of creating the package).
- Create a converter package for each major version of OCaml, that converts `MLAst` to the each respective version of the ocaml compiler's AST. Then when using Reason with `4.02.3` we would parse into `MLAst` which resembles `4.04` (or even higher), but would downgrade it to `4.02.3` at the final step so it can be used in older versions. If you use a feature that's only supported in `4.04` then the downgrade to `4.02.3` would give a helpful error such as ("This feature is only available in `4.04`").
### What does this give us?

This not only helps us immediately use any new features the moment they're merged, (even experimenting with ones _before_ they're merged), but it also makes it so that we can write the Reason parser once, keeping it up to date with new merged features (such as inline records in `4.03`/`4.04`), but use that one implementation in 4.02.3. So this is just as much about using these new upcoming features very soon, as it is about our own productivity maintaining the Reason parser/printer. Right now, we can't merge the PR for upgrading to `4.03` because then it would be incompatible with `4.02.3`.
### Why not just only support `4.04+`?

Because this isn't the only time we're going to need to keep up with compiler changes. This proposal here helps us keep up to date, without having to maintain several forks of `Reason` for each underlying compiler version.

yunxing:
Like this idea. We need to be careful about some internal representation changes like this one: https://github.com/ocaml/ocaml/commit/9eca3e45191ef52ce519e2d5bb4688da62bf7496 where they store integer as strings in the AST tree.

jaredly:
Oooh and can we have the `MLAst` types have names that are actually easy to read? :P been reading through ast code a lot, and the names are pretty dense

jordwalke:
@jaredly I agree, they're not very readable. We can clean them up if we like, if we're okay that it may make it more difficult to keep up to with newer OCaml compiler changes. The AST would likely contain the union of all upcoming compiler features (Implicits, Algebraic Effects).

[...]

PR #724, opened 2016-08-28
JSON support for -parse and -print

jaredly:
To make this work, I copied in the interfaces wholesale from https://github.com/ocaml/ocaml/blob/4.02/parsing/parsetree.mli as well as other files that I needed to access, so that `@@deriving yojson` could work its magic.
If anyone knows of a less labor-intensive way of doing this, feel free to speak up :)

Here's what it looks like

```
:~/clone/rsn/Reason$ cat > example.re
print_endline "hello" /* world */
^C
:~/clone/rsn/Reason$ ./refmt_impl.native -print json example.re > out.json
:~/clone/rsn/Reason$ cat out.json
[[{"pstr_desc":["Pstr_eval",{"pexp_desc":["Pexp_apply",{"pexp_desc":["Pexp_ident",{"txt":["Lident","print_endline"],"loc":{"loc_start":{"pos_fname":"example.re","pos_lnum":1,"pos_bol":0,"pos_cnum":0},"loc_end":{"pos_fname":"example.re","pos_lnum":1,"pos_bol":0,"pos_cnum":13},"loc_ghost":false}}],"pexp_loc":{"loc_start":{"pos_fname":"example.re","pos_lnum":1,"pos_bol":0,"pos_cnum":0},"loc_end":{"pos_fname":"example.re","pos_lnum":1,"pos_bol":0,"pos_cnum":13},"loc_ghost":false},"pexp_attributes":[]},[["",{"pexp_desc":["Pexp_constant",["Const_string","hello",null]],"pexp_loc":{"loc_start":{"pos_fname":"example.re","pos_lnum":1,"pos_bol":0,"pos_cnum":14},"loc_end":{"pos_fname":"example.re","pos_lnum":1,"pos_bol":0,"pos_cnum":21},"loc_ghost":false},"pexp_attributes":[]}]]],"pexp_loc":{"loc_start":{"pos_fname":"example.re","pos_lnum":1,"pos_bol":0,"pos_cnum":0},"loc_end":{"pos_fname":"example.re","pos_lnum":1,"pos_bol":0,"pos_cnum":21},"loc_ghost":false},"pexp_attributes":[]},[]],"pstr_loc":{"loc_start":{"pos_fname":"example.re","pos_lnum":1,"pos_bol":0,"pos_cnum":0},"loc_end":{"pos_fname":"example.re","pos_lnum":1,"pos_bol":0,"pos_cnum":21},"loc_ghost":false}}],[[" hello ",["EndOfLine"],{"loc_start":{"pos_fname":"example.re","pos_lnum":1,"pos_bol":0,"pos_cnum":21},"loc_end":{"pos_fname":"example.re","pos_lnum":1,"pos_bol":0,"pos_cnum":33},"loc_ghost":false}]]]
:~/clone/rsn/Reason$ ./refmt_impl.native -print re out.json
print_endline "hello"; /* hello */
```

### DIFF
--- a/_tags
+++ b/_tags
@@ -4,7 +4,7 @@ true: warn(@5@8@10@11@12@14@23-24@26@29@40), bin_annot, safe_string, debug
 <node_modules/**>: -traverse
 "formatTest": -traverse
 "src": include
-<src/*>: package(menhirLib unix compiler-libs.common ocamlbuild findlib easy-format BetterErrors merlin_extend re.str)
+<src/*>: package(menhirLib unix compiler-libs.common ocamlbuild findlib easy-format BetterErrors merlin_extend re.str ppx_deriving.std yojson ppx_deriving_yojson)
 <src/reason_utop.*>: package(menhirLib utop)
 
 <src_gen/*>: package(sedlex)
--- a/.merlin
+++ b/.merlin
@@ -1,3 +1,3 @@
-PKG utop merlin_extend compiler-libs easy-format re.str yojson
+PKG utop merlin_extend compiler-libs easy-format re.str yojson result
 B _build/src
 S src
--- a/.merlin
+++ b/.merlin
@@ -1,3 +1,3 @@
-PKG utop merlin_extend compiler-libs easy-format re.str
+PKG utop merlin_extend compiler-libs easy-format re.str yojson
 B _build/src
 S src
--- a/opam
+++ b/opam
@@ -26,6 +26,9 @@ depends: [
   "BetterErrors" {>= "0.0.1"}
   "menhir"       {>= "20160303"}
   "re"           {>= "1.5.0"}
+  "ppx_deriving" {>= "4.0"}
+  "ppx_deriving_yojson" {>= "3.0"}
+  "yojson"       {>= "1.3.2"}
 ]
 depopts: [
 ]
--- a/pkg/build.ml
+++ b/pkg/build.ml
@@ -52,7 +52,6 @@ let () =
     Pkg.bin  ~auto:true "src/reactjs_jsx_ppx" ~dst:"reactjs_jsx_ppx";
     Pkg.share "editorSupport/emacs/refmt.el" ~dst:"../emacs/site-lisp/refmt.el";
     Pkg.share "editorSupport/emacs/reason-mode.el" ~dst:"../emacs/site-lisp/reason-mode.el";
-
     (* atom-reason *)
     (* Unfortunately we have to specificy each individual file *)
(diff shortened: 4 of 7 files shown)


_tags
```diff
@@ -4,7 +4,7 @@ true: warn(@5@8@10@11@12@14@23-24@26@29@40), bin_annot, safe_string, debug
 <node_modules/**>: -traverse
 "formatTest": -traverse
 "src": include
-<src/*>: package(menhirLib unix compiler-libs.common ocamlbuild findlib easy-format BetterErrors merlin_extend re.str)
+<src/*>: package(menhirLib unix compiler-libs.common ocamlbuild findlib easy-format BetterErrors merlin_extend re.str ppx_deriving.std yojson ppx_deriving_yojson)
```
jordwalke: I've always wondered what flags *actually* get passed to the compiler. I want to know so that we can support ppx in jengaboot (instead of ocamlbuild `_tags` which are super annoying).

jordwalke:
I'll just write some thoughts as I think of them:

- We currently have two serialization formats currently. One is the `binary` format, which is the binary serialization (that OCaml requires for `-pp`), and the other called `binary_reason`, is just like `binary`but with comments added, and a bit of metadata (such as some of the initial flags).
- That second form `binary_reason` is used for upgrading between versions of Reason (for a fixed version of the AST).
- It seems like this JSON format might be a good candidate for replacing the `binary_reason` format. (If it was enhanced to include comment data and flags in the JSON output). No sense having *three* formats, right?
- The fact that you copy/pasted a version of the OCaml AST is not an issue. In fact, it's part of the [*plan*] (https://github.com/facebook/reason/issues/706). The main difference here, is that we actually want to parse into the most modern version of the AST. Then we need to convert from that format back into `4.02.3` etc depending on which version of the compiler is available.

jordwalke:
I can definitely see the benefits of having the `json` format be the canonical output. We could write tooling/scripts in any language more easily, and in a loosely typed way. I think part of getting this in should be removing the `reason_binary` format.

jordwalke:
Also, @yunxing or I should check in a fresh copy of the AST data structure without any of the ppx attributes added, and then you could rebase on top of it to add the ppx attributes (licensing).

jaredly:
sounds great! I've got the comments stuff included as well now

yunxing:
@jaredly let's wait a bit on this diff as we are thinking to move to 4.03. I will work on #706 tomorrow to get a latest copy of parsetree into the repo.

_tags
yunxing: I think there is a -ppx option in ocamlc which let you specify ppx plugins

yunxing:
Still working #706, got some progress

Issue #725, opened 2016-08-29
Make rebuild support ocamlc (in addition to ocamlopt)


yunxing:
Just realize we only have a script called "reopt" which is a wrapper for ocamlopt, also need a "rec" so we can support ocamlc functionalities (like compiling cmo files). 

@freebroccolo

PR #727, opened 2016-08-29
Make rebuild support ocamlc

yunxing:
Add a new command "rec" to be used in rebuild as a replacement for ocamlc. This is needed to
support ocamlc specific functionalities, such as compiling cmo files.

#725


yunxing:
Merging this to unblock @freebroccolo

PR #732, opened 2016-08-31
Enable `-parse binary` for refmt

jaredly:
refmt can export it, why not import as well? I wanted this to debug a ppx (b/c
 they output binary ast) and I wanted to print it back as ocaml.


jordwalke:
This appears to be missing the ability to parse binary for interfaces. (And I think the docs need to mention the option).

jaredly:
all set I think

yunxing:
Looks great, thanks!

jordwalke:
Thanks

PR #737, opened 2016-09-03
Fix pretty printing #=

### LINKED ISSUE #736: Infix #= doesn't print correctly
cc @Schmavery 

```
id #= (Some 2)
```

Gets printed as

```
id #= Some 2
```

And throws `Error: The constructor Some expects 1 argument(s),
       but is applied here to 0 argument(s): ...`

Funny printing with functions:

```
let foo a => a;
id #= (foo 2);
```

Becomes 

```
let foo a => a;
id #= foo 2;
```

Then becomes

```
let foo a => a;
(id #= foo) 2;
```

bsansouci:
Fixes #736.
Special casing `#` followed by any of `['!' '$' '%' '&' '+' '-' '.' ':' '<' '=' '>' '?' '@' '^' '|' '~']` or
`('\\' '*') | ('\\' '/')` to be printed properly with the right
precedence and associativity.

I added tests in a new file that I called "bucklescript", I'm not sure that's the best name for it.

### DIFF
--- a/formatTest/unit_tests/expected_output/bucklescript.re
+++ b/formatTest/unit_tests/expected_output/bucklescript.re
@@ -4,24 +4,19 @@ bla#=(Some 10);
 
 bla#=(someFunc (Some 10));
 
-bla#=((someFunc someFunc2) 10);
-
 test##var#=(Some (-10));
 
 obj##.prop;
 
 obj##.prod := exp;
 
-preview##style##border#=Js.string "1px black dashed";
+preview##style##border#=(Js.string "1px black dashed");
+
+preview##(style##border)#=args somenum;
 
 x##y##z#=(xxxx##yyyy##zzzz);
 
-let result =
-  js_method_run1 (!react)#createElement foo;
-
-let foo a => a;
-
-id#=(foo 2);
+let result = js_method_run1 (!react)#createElement foo;
 
 add zz##yy xx##ww;
 
@@ -65,6 +60,6 @@ res#=?!!z##q;
 
 res#=?!!z##(q##a);
 
-let result = myFunction (x y#=z) (a b#=c);
+let result = myFunction (x y##z) (a b#=c);
 
 (!x)##y##(b##c);
--- a/formatTest/unit_tests/expected_output/bucklescript.re
+++ b/formatTest/unit_tests/expected_output/bucklescript.re
@@ -66,3 +66,5 @@ res#=?!!z##q;
 res#=?!!z##(q##a);
 
 let result = myFunction (x y#=z) (a b#=c);
+
+(!x)##y##(b##c);
--- a/formatTest/unit_tests/expected_output/bucklescript.re
+++ b/formatTest/unit_tests/expected_output/bucklescript.re
@@ -1,0 +1,68 @@
+bla#=10;
+
+bla#=(Some 10);
+
+bla#=(someFunc (Some 10));
+
+bla#=((someFunc someFunc2) 10);
+
+test##var#=(Some (-10));
+
+obj##.prop;
+
+obj##.prod := exp;
+
+preview##style##border#=Js.string "1px black dashed";
+
+x##y##z#=(xxxx##yyyy##zzzz);
+
+let result =
+  js_method_run1 (!react)#createElement foo;
+
+let foo a => a;
+
+id#=(foo 2);
+
+add zz##yy xx##ww;
+
+/* These should print the same */
+let res = x##y + z##q; /* AST */
+
+let res = x##y + z##q; /* Minimum parens */
+
+/* These should print the same */
+let res = y + z##q##a; /* AST */
+
+let res = y + z##q##a; /* Min parens */
+
+/* Make sure it's actually parsed as left precedence
+ * and that is maintained when printed */
+let res = z##(q##a); /* AST */
+
+let res = z##(q##a); /* Min parens */
+
+/* These should print the same */
+let res = !x##y; /* AST */
+
+let res = !x##y; /* Minimum parens */
+
+/* These should print the same */
+let res = !z##q##a; /* AST */
+
+let res = !z##q##a; /* Min parens */
+
+/* These should print the same */
+let res = ?!!x##y; /* AST */
+
+let res = ?!!x##y; /* Minimum parens */
+
+/* These should print the same */
+let res = ?!!z##(q##a); /* AST */
+
+let res = ?!!z##(q##a); /* Min parens */
+
+res#=?!!z##q;
+
+res#=?!!z##(q##a);
+
+let result = myFunction (x y#=z) (a b#=c);
--- a/formatTest/unit_tests/expected_output/bucklescript.re
+++ b/formatTest/unit_tests/expected_output/bucklescript.re
@@ -10,13 +10,16 @@ obj##.prop;
 
 obj##.prod := exp;
 
-preview##style##border#=(Js.string "1px black dashed");
+preview##style##border#=(
+                          Js.string "1px black dashed"
+                        );
 
 preview##(style##border)#=args somenum;
 
 x##y##z#=(xxxx##yyyy##zzzz);
 
-let result = js_method_run1 (!react)#createElement foo;
+let result =
+  js_method_run1 (!react)#createElement foo;
 
 add zz##yy xx##ww;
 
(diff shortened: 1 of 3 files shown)


facebook-github-bot-0:

facebook-github-bot-1:
Thank you for signing our Contributor License Agreement. We can now accept your code for this (and any) Facebook open source project. Thanks!

src/reason_pprint_ast.ml
```diff
@@ -630,9 +630,6 @@ let rules = [
     (TokenPrecedence, (fun s -> (Right, s = "asr")));
   ];
   [
-    (TokenPrecedence, (fun s -> (Left, s.[0] == '#')));
```
jordwalke: I think we still want this, but it's precedence should match the precedence in the parser (I think I commented about this on the original diff).

src/reason_pprint_ast.ml
```diff
@@ -3014,6 +3011,23 @@ class printer  ()= object(self:'self)
         in
         let rightItm = self#simplifyUnparseExpr rightExpr in
         Some (label ~space:forceSpace (atom prefixStr) rightItm)
+      | (Infix s, [(_, e1); (_, e2)]) when s.[0] = '#' ->
+        (*
+          If the expression parsed is
+            expr1##expr2
+          then keeping no spaces between that infix operator and the expressions
+          is important as ## signifies "js property access" and therefore looks
```
jordwalke: I think it signifies "extern access" - it could be interoperating with an C++ object etc.

src/reason_pprint_ast.ml
```diff
@@ -3014,6 +3011,23 @@ class printer  ()= object(self:'self)
         in
         let rightItm = self#simplifyUnparseExpr rightExpr in
         Some (label ~space:forceSpace (atom prefixStr) rightItm)
+      | (Infix s, [(_, e1); (_, e2)]) when s.[0] = '#' ->
```
jordwalke: I think it's wrong that this is considered `Infix`.  Infix function application cannot be passed as an argument without guarding parens `add (3 + 2) (4 + 5)`. But I believe `##` can be passed as an argument to a function. `add x##y z##q`

bsansouci: Are you saying the naming / AST node is wrong? I just tried and `add x##y z##q` parses and prints correctly.

formatTest/unit_tests/input/bucklescript.re
```diff
@@ -0,0 +1,20 @@
...
+
+bla #= Some 10;
+
+bla #= someFunc (Some 10);
+
+test##var #= Some 10;
+
+obj##.prop;
+
+obj##.prod := exp;
```
jordwalke: I think we need more tests for how it interacts with other infix. For example, make tests that verify parsing and printing match the left vs. right precedence and priority.

```rust
/* These should print the same */
let res = ((x##y) + (z##q));   /* AST */
let res = x##y + z##q; /* Minimum parens */

/* These should print the same */
let res = (y + (z##q)##a);   /* AST */
let res = y + z##q##a;   /* Min parens */


/* Make sure it's actually parsed as left precedence
 * and that is maintained when printed */
let res = (z##(q##a));   /* AST */
let res = z##(q##a);   /* Min parens */
```

src/reason_pprint_ast.ml
bsansouci: Pretty printing using just the `rules` array with `##` being an `Infix` op does this:
```
bla #= 10;

(bla #= Some) 10;

(bla #= someFunc) (Some 10);

(test ## var #= Some) 10;

obj ##. prop;

obj ##. prod := exp;

(preview ## style ## border #= Js.string) "1px black dashed";

x ## y ## z #= (xxxx ## yyyy) ## zzzz;

let result = js_method_run1 (!react)#createElement foo;

let foo a => a;

id #= foo 2;

add (zz ## yy) (xx ## ww);

/* These should print the same */
let res = x ## y + z ## q; /* AST */

let res = x ## y + z ## q; /* Minimum parens */

/* These should print the same */
let res = y + z ## q ## a; /* AST */

let res = y + z ## q ## a; /* Min parens */

/* Make sure it's actually parsed as left precedence
 * and that is maintained when printed */
let res = z ## (q ## a); /* AST */

let res = z ## (q ## a); /* Min parens */
```

This does the right thing for `##` but might need to be special cased for `#=` which has a special meaning. How many of those symbols starting with `#` have special meaning (i.e.: aren't expected to have the same semantics as `##`)?

formatTest/unit_tests/input/bucklescript.re
bsansouci: Added.

src/reason_pprint_ast.ml
bsansouci: What exactly is that rules array used for? I added it at the right place (just higher precedence that INFIXOP4)

src/reason_pprint_ast.ml
jordwalke: > Are you saying the naming / AST node is wrong? I just tried and add x##y z##q parses and prints correctly.

Correct, I'm saying we incorrectly classify it as `Infix`. If we leave it as is, there should at least be a comment saying "This isn't *actually* infix, it's simple". You might be able to easily change it to `Simple` though instead of `Infix` in this diff.

src/reason_pprint_ast.ml
jordwalke: It should match the order that tokens are specified in the `.mly` file. I believe that means it should match the ranking of `SHARPOP` in the list of tokens. I believe that puts it higher than `isSimplePrefixToken` (which would be things like a leading `BANG` `!`).

bsansouci: Its precedence is below bang though https://github.com/facebook/reason/blob/499c11d886913c5a499d8e5f6d2a606386675440/src/reason_parser.mly#L1051, which is why I put it in 2nd place. (This diff is old too).

jordwalke: Oh, it's misleading because it uses `below_DOT_AND_SHARP` in the actual parse rule. So I think my suggestion for where its location in the list should be still stands, right?

jordwalke: So `!`, `PREFIXOP` are parsed with precedence `below_DOT_AND_SHARP`, which is, well - below dot and sharp/`SHARPOP`. That `below_DOT_AND_SHARP` is what `isSimplePrefixToken` tests. So that means the `SHARPOP` identifiers need to be considered higher precedence than the test for `isSimplePrefixToken`.

The fact that the tests pass even with the incorrect ordering in `rules`, means we're missing a test case!

I might suggest adding the failing test case, then changing the order to what I suggested, and you would then see it pass!

The incorrect precedence ranking is between any `SHARPOP` and anything that returns true for `isSimplePrefixToken ` (`PREFIXOP`, `!` etc).

So some good test cases would be:

```rust
/* These should print the same */
let res = (! (x##y));   /* AST */
let res = !x##y;        /* Minimum parens */

/* These should print the same */
let res = (!(z##q)##a);   /* AST */
let res = !z##q##a;   /* Min parens */

/*
 * And then copy/paste those tests, but with another `isSimplePrefixToken`
 * instead of `!` (for example `?!!` should have the same precedence as !)
 */
```

bsansouci: Fixed. Though this sounds a bit awful, why has `PREFIXOP` a precedence but then uses `below_DOT_AND_SHARP`? Can we fix that?

jordwalke: We wanted `PREFIXOP` and `BANG` to have the exact same precedence as each other (but only in *some* use cases of `BANG` - not others). The only way to do that that I could see was to create a custom `%prec` and then make both of them use that custom precedence.

bsansouci: Oh I see, thanks for explaining. Maybe a little comment regarding that would be nice then :)

formatTest/unit_tests/expected_output/bucklescript.re
```diff
@@ -0,0 +1,68 @@
...
+
+bla#=(someFunc (Some 10));
+
+test##var#=(Some (-10));
+
+obj##.prop;
+
+obj##.prod := exp;
+
+preview##style##border#=(
```
jordwalke: Should fix this in a follow up.

PR #740, opened 2016-09-05
[Reason] Add pretty_print to ocamlmerlin_reason

jordwalke:
Summary: This adds the hook needed to bump merlin/merlin-extend to master.

IMPORTANT
=========
The main docs still point to old tags, and that's fine - those are stable.
But for developing, we now need to pin merlin/merlin-extend from master, and I've updated the README to reflect this.

Test Plan:

Reviewers:let-def, yunxing, sanderspies

CC:

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,28 +33,6 @@ opam pin add -y reason 'https://github.com/facebook/reason.git#0.0.6'
 
 ```
 
-Install master (For contributors)
-----------
-```
-# On OSX, install opam via Homebrew:
-brew update
-brew install opam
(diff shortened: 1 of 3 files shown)


jordwalke:
This diff just brings in @let-def's changes, and adds the missing pieces. I haven't tested the new merlin features thoroughly with this diff but the old merlin features work.

https://github.com/let-def/Reason/commit/361a319a8785084d32b7bbb4d2e6e5f17da58bab

jordwalke:
Merging. Will revert if this causes build failures.

PR #746, opened 2016-09-07
[npm] use utop-bin

yunxing:
utop-bin is a manually generated package that prebuilts utop on linux and darwin systems. 

Using it as a dependency reduces half of the Reason installation time [down to 3.5 minutes on a 2014 Mac] through npm.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,9 +55,9 @@
         "substs": "https://github.com/yunxing/substs.git",
         "easy-format": "https://github.com/npm-opam/easy-format",
         "merlin-extend": "https://github.com/npm-opam/merlin-extend",
-        "opam-installer-bin": "https://github.com/yunxing/opam-installer-bin.git",        
+        "opam-installer-bin": "https://github.com/yunxing/opam-installer-bin.git",
         "nopam": "https://github.com/yunxing/nopam.git",
-        "utop": "https://github.com/npm-opam/utop",
+        "utop-bin": "https://github.com/reasonml/utop-bin",
         "menhir": "https://github.com/npm-opam/menhir"
     },
     "scripts": {


jordwalke:
Wow!
How can we make it so that we can rebuild `utop` optionally? We will need to do this for some CI hosts.

jordwalke:
So, I imagine being able to set an environment variable before running `npm install` such as:

`$REASON_ALWAYS_RECOMPILE`

And then the binaries will be ignored, and rebuilt instead.

yunxing:
@jordwalke I'm thinking we can provide a package.json.source, which has dependencies in source code version,in the directory, as the first step of the CI, one can replace package.json with `package.json.source`.

jordwalke:
We can also look into shrinkwrap which might help us provide overrides for packages. Also, I realized this problem will never occur with utop because no one needs to use utop on production CI. We just need to somehow setup a test to make sure we can always build utop from source again on every commit. So maybe the main use case is actually github CI building utop from source every commit. Not production CI for software built on reason.

jordwalke:
Beautiful!

jordwalke:
Merging because I think it was previously passing, but then my accidental closing/reopening caused the build to be re-queued up.

Filed an issue here to track the ability to rebuild from source optionally. https://github.com/facebook/reason/issues/747

package.json
```diff
@@ -55,9 +55,9 @@
         "substs": "https://github.com/yunxing/substs.git",
         "easy-format": "https://github.com/npm-opam/easy-format",
         "merlin-extend": "https://github.com/npm-opam/merlin-extend",
-        "opam-installer-bin": "https://github.com/yunxing/opam-installer-bin.git",        
+        "opam-installer-bin": "https://github.com/yunxing/opam-installer-bin.git",
         "nopam": "https://github.com/yunxing/nopam.git",
-        "utop": "https://github.com/npm-opam/utop",
+        "utop-bin": "https://github.com/reasonml/utop-bin",
```
jordwalke: I think the way shrinkwrap would work, is that we still *call* this `utop` in the JSON key, then shrinkwrap allows us to specify *which* `utop` to use - the binary build or the source build.

Issue #747, opened 2016-09-07
A way to force rebuilding binary dependencies from source.


jordwalke:
We (thankfully) have `utop-bin` and `jenga-bin`. This _massively_ speeds up build times of projects. However:

We still need a way to build from source on some CIs. (Typically only `jenga`, typically no need to build `utop` from source).

We can look into `npm`'s "shrinkwrap" feature, which could allow this.

jaredly:
I think this is stale - we have a much different build environment now

PR #752, opened 2016-09-09
use utop-bin (again!)

yunxing:
Previously we ran into some issues with utop-bin and we disabled it. But with some recent changes it works great now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,9 +55,9 @@
         "substs": "https://github.com/yunxing/substs.git",
         "easy-format": "https://github.com/npm-opam/easy-format",
         "merlin-extend": "https://github.com/npm-opam/merlin-extend",
-        "opam-installer-bin": "https://github.com/yunxing/opam-installer-bin.git",        
+        "opam-installer-bin": "https://github.com/yunxing/opam-installer-bin.git",
         "nopam": "https://github.com/yunxing/nopam.git",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -55,9 +55,9 @@
         "substs": "https://github.com/yunxing/substs.git",
         "easy-format": "https://github.com/npm-opam/easy-format",
         "merlin-extend": "https://github.com/npm-opam/merlin-extend",
-        "opam-installer-bin": "https://github.com/yunxing/opam-installer-bin.git",        
+        "opam-installer-bin": "https://github.com/yunxing/opam-installer-bin.git",
         "nopam": "https://github.com/yunxing/nopam.git",
-        "utop": "https://github.com/npm-opam/utop",
+        "utop-bin": "https://github.com/reasonml/utop-bin",
```
jordwalke: Do you want to just name it `utop`? that way people can specify overrides for their top level app to build it from source?

```
"utop": "https://github.com/reasonml/utop-bin"
```

jordwalke: (That would require that `utop-bin`'s `package.json` has its name specified as `"utop"` even if the repo is named `utop-bin`).

yunxing: maybe let's keep it as `utop-bin` for now, until we find a good solution to https://github.com/facebook/reason/issues/747?

PR #757, opened 2016-09-10
[atom-reason] Add more operators so that ligatures render correctly

vramana:
image

Both `|>` and `>>=` both render correctly. Are there any more common operators that we need to add?

cc @jordwalke


jordwalke:
!==
===
...

Worth checking those.

jordwalke:
Uggh it didn't render correctly.

vramana:
`...` works I am adding `!=` and `!==`.

chenglou:
Is this good to merge?

Issue #712, opened 2016-08-25
Better syntax for e.g. `bs.obj`


chenglou:
I don't know whether we should do this, but it'd be nice to have e.g. `#{a: 1, b: 2}` instead of `[%bs.obj {a: 1, b: 2}]`. Maybe this would couple us too tightly with BuckleScript, or maybe there's a more general approach.

Good for react (props, style).

(obj set & get could be better too, but that's another discussion.)

cc @jordwalke @bobzhang

jordwalke:
`#{a: 1, b: 2}` can be for FFI objects, and `{"a": 1, "b": 2}` could be sugar for string map data structures. There is no parser ambiguity for any of these with records/objects.

chenglou:
By string map, do you mean js obj?

bobzhang:
note `bs.obj` does more than that, collections also work

```ocaml
[%bs.obj [ { x = 3 } ; { x = 4} ] ]
```
But here `#` is so light weight so it does not really matter, I have a bunch of cool ideas to make FFI syntax exactly like typescript/flow, for example, the flow declaration below can also work in Reason+ Buckle

```js
declare module Misc {
  declare var DEBUG: bool;
  declare function isLeapYear(year: string): bool;
}
```

jordwalke:
Clever.

jordwalke:
@chenglou : By `string` map I meant an actual map data structure.

chenglou:
do we need first-class support for Map? People still don't use it as much in the wild. I feel that `{
"foo": bar}` syntax could be much better spent on js obj. And that hash syntax could be spent on... hash map =)

jordwalke:
Very well. :D

jordwalke:
I suppose `{"foo": bar}` would be suitable for JS objects, since JS does have string keys and this is acknowledging that fact.

SanderSpies:
@bobzhang - could you write your ideas down for using Flow syntax for FFI? I would like to help out with this.

bobzhang:
@SanderSpies 

so 
```js
declare module Misc {
  declare var DEBUG: bool;
  declare function isLeapYear(year: string): bool;
}
```

can be interpreted as 
```ocaml
external debug : bool = "DEBUG" [@@bs.module "Misc"]
external isLeapYear : year:string -> bool = "" [@@bs.module "Misc"]
```

This translate can be done either in Reason side or Buckle side

SanderSpies:
[...]

PR #759, opened 2016-09-11
Support js obj literal syntax

bsansouci:
This is a simple PR to support `{"key": 10}` to be internally represented as `[%bs.obj {key: 10}`. It allows for a nicer js interop through BuckleScript.
Related discussion: #712 

I'd love some feedback over this.

This supports punning which looks a bit weird `{"a": a}` becomes `{"a"}`.


formatTest/unit_tests/expected_output/bucklescript.re
```diff
@@ -0,0 +1,88 @@
...
+
+let c = {
+  "a",
+  "b",
+  "func": fun a => a##c#=(func 10)
+};
+
+let d = {
+  "a": a2,
+  "b", /* punned */
```
jordwalke: I haven't looked at the parser part of this diff but I believe this would cause an issue if only a single key was punned. It would be indistinguishable from a sequence expression with a single item - a string.
```
let x = {"hi"};
```

After that, `x` is `"hi"`.

bsansouci: I think this isn't allowed because I used the same code as `Pexp_record`. It checks if there's only one identifier (they're fake identifiers) and prevents punning (same way records would disallow `{a}`.

jordwalke: Nice! But the reason that this is justified for records, was because no one (okay hardly anyone) ever has a record with only one field. However, with objects, I think it's much more likely because objects are extensible (in terms of their rows) and you can make a function that expects an object (that has a name, and possibly more) which encourages people to make an object with just the one/two fields required by the function.

jordwalke: How bad would it be to give up punning?

bsansouci: Good point. Also it does look a bit weird in my opinion. Not bad at all. Will fix soon.

chenglou:
@bsansouci rebase and I'll rage merge

chenglou:
cc @bobzhang reason "tied" to BuckleScript is a bit biased. Is it possible to have a more generic name than `bs.obj`? Like `js.obj`?

bobzhang:
@chenglou Actually I would like to rename all bs into js, the main concern is to avoid making other ocaml/js tools unhappy..

chenglou:
Ben said it was good for merging!

chenglou:
Nvm reverting (and force pushing to erase the merge). Fixing ben's files

PR #762, opened 2016-09-12
Add node_modules to .gitignore

yunxing:
Somethings we want to test npm installation and it will generate node_modules.


yunxing:
Should be harmless

PR #764, opened 2016-09-14
BuckleScript bs.obj support

chenglou:
See #759, this is a non-badly-merged version.

> This is a simple PR to support `{"key": 10}` to be internally represented as `[%bs.obj {key: 10}`. It allows for a nicer js interop through BuckleScript.
Related discussion: #712

> This supports punning which looks a bit weird `{"a": a}` becomes `{"a"}`.

### DIFF
--- a/formatTest/unit_tests/expected_output/bucklescript.re
+++ b/formatTest/unit_tests/expected_output/bucklescript.re
@@ -66,3 +66,23 @@ res#=?!!z##(q##a);
 let result = myFunction (x y##z) (a b#=c);
 
 (!x)##y##(b##c);
+
+let a = {"key": 10};
+
+let b = {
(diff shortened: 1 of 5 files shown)


chenglou:
@bsansouci punning `let foo = {"a"}` into `let foo = "a"` is wrong. How do we fix it?

chenglou:
Nvm that's intentional.

chenglou:
cc @bsansouci @jordwalke @yunxing note that printing to ocaml with this feature is currently broken. It's very edge-casey though.

```sh
$ echo 'let a = {"a-a": b}' | refmt -use-stdin true -is-interface-pp false -parse re -print ml
let a = [%bs.obj { a-a = b }]
```

See more context in the referred BS issue.

Issue #766, opened 2016-09-14
New object syntax proposal


chenglou:
From @jordwalke. I'm just paraphrasing here.

``` rust
let x = {
  pub myMethod argOne => ...
  pri myPrivateMethod argOne => ...
};

/* typed, open */
let x : {myMethod: t, ..} = someObjectDeclaration;
/* typed, closed. Rarely used? */
let x : {myMethod: t, .} = someObjectDeclaration;
```
## Proposal 1 for method call syntax:

``` rust
x.myMethod argOne;
```

This requires changing record function calling to

``` rust
record->myFunc argOne
```

The justification being that record function is rare.
## Proposal 2

Keep the same record function syntax, use the following for method call (notice the `:`):

``` rust
myObject.method: argOne argTwo
```
## Tradeoffs

The first proposal for method calling would play nicely if function application of multiple curried arguments were expressed via standard `(,,)` syntax - for example:

``` rust
let adder = {
   pub add (a, b) => a + b
};
let result = adder.add(100, 200);
/* Partially applying 4 */
let addFourTo = adder.add(4);
let seven = addFourTo(3);
```

The second proposal for method calling would play nicely with named arguments of the form `label:` - for example:

``` rust
let myObject = {
   pub someMethod x: xx y: yy => x + y
};
let result = myObject.someMethod: x:10 y:200;
```

vramana:
Is there any drawback in making methods private by default?

jordwalke:
the `pub` and `pri` are needed to distinguish the object from a block expression. After the `{` there always needs to be something that can determine the kind of object being parsed. Record vs. block expression, vs. method.

jordwalke:
You need something there, even if it's not `pub`/`pri` (for example `met` for "method")

SanderSpies:
Close please.

PR #767, opened 2016-09-15
New object syntax

SanderSpies:
Related to #766


src/reason_lexer.mll
```diff
@@ -112,6 +112,7 @@ let keyword_table =
     "match", MATCH; (* Including MATCH for source transforming compat *)
     "method", METHOD;
     "module", MODULE;
+    "pub", PUB;
```
jordwalke: I think we need to make suer `pub` and `pri` can be valid identifiers. Imagine converting the following code from `ML`:

```
let pri = "private"
```

formatTest/typeCheckedTests/expected_output/oo.re
```diff
@@ -38,15 +38,15 @@ class virtual stackWithAttributes 'a init => {
   /* Virtual member */
   val virtual dummy: unit;
   val mutable v: list 'a = init;
-  method virtual implementMe: int => int;
-  method pop =
+  pub virtual implementMe: int => int;
+  pub pop =
```
jordwalke: I thought we had some test cases that had a `private` method.

SanderSpies: For types yes, but not for methods - I added one but I don't think all the private method cases are covered.

SanderSpies:
I will update this PR with the new method invocation syntax once we have made a decision on that.

jordwalke:
> I will update this PR with the new method invocation syntax once we have made a decision on that.

Can we actually do them separately? This diff is still good even without that right?

src/reason_lexer.mll
jordwalke: Is there now a test case for this (and a fix?)

jordwalke:
(I notice there's some merge conflicts)

formatTest/typeCheckedTests/expected_output/oo.re
```diff
@@ -235,16 +235,16 @@ class myClassWithAnnotatedReturnType3
 class myClassWithAnnotatedReturnType3_annotated_constructor:
   int =>
   new {
-    method x: int;
-    method y: int
+    pub x: int;
```
jordwalke: Do you think we should eventuallyunify these `class_instance_type` with object instance types too? They allow much more to be specified than object types (for example public vs. private, and even values). But could it be that class instance types could be a syntactic superset of object instance types?

SanderSpies:
@jordwalke 
- Rebased on the latest master
- I still need to add extra test cases
- Not sure about unifying with `class_instance_type`, it's not clear to me what we would gain from it.

formatTest/typeCheckedTests/input/oo.re
```diff
@@ -113,77 +113,77 @@ and secondRecursiveClass init => {
  * Anonymous objects.
  */
 
-type closedObj = <>;
+type closedObj = {.};
 
 let (<..>) a b => a + b;
 let five = 2 <..> 3;
 
-type nestedObj = < bar : <a: int>>;
```
jordwalke: Heh, we can kill the token splitting hack now! ;)

src/reason_lexer.mll
SanderSpies: Done

formatTest/typeCheckedTests/input/oo.re
SanderSpies: I've remove `splitToken` also now.

cc @yunxing

SanderSpies:
This PR is ready for review.

src/reason_lexer.mll
jordwalke: Great!

sgrove:
Might be nice to simultaneously get some documentation update for http://facebook.github.io/reason/ on the new syntax

jordwalke:
If so, then we might want to put it somewhere deep like in the "Diving Deeper" section because the object syntax is likely to change much (even though object type syntax might not much).

SanderSpies:
@sgrove @jordwalke 

As requested I've added objects documentation in the "Diving Deeper" section.

docs/index.html
```diff
@@ -1021,6 +1021,81 @@
 let justTwo = hasExactlyTwoCars myCarList;  /* true! */
 ```
 
+Objects
```
jordwalke: Looking really great!

SanderSpies:
Rebased

cc @chenglou @jordwalke

freebroccolo:
This looks pretty nice. I like the shorter names for methods.

I'm a little bit concerned about putting the '.' at the end of the block though. This would mean we couldn't use that information to determine the parsing context in any of the textmate based plugins (atom, sublime, vscode) because they can't do backtracking. So we would have to try to do some approximate parsing of both record and object types at the same time which could be messy especially as the syntax changes more in the future.

I would suggest changing it to where the syntax is `{.` and `{..`. Not having a space between `{` and `.` there also makes the parsing for the highlighting much easier and more accurate.

Another alternative though that I think might be more visually distinct would be to use `{=` for the closed objects and `{>` for the open ones. This might work thematically because of similarity to the notation for polymorphic variants.

Thoughts?

jordwalke:
Those are great points, @freebroccolo. Let's also consider how that would work with some kind of syntax for record types that could be inherited (not that we have a working implementation, but it would be a useful feature). Something like:

```rust
type point = {
  ...point2D,
  z: int
};
```

Which matches the record spread syntax.

Perhaps `..` could be used for objects?

```rust
type hasZ 'a = {.., z: int} as 'a
```

And we can later add sugar for that common pattern to be written as:

```rust
type hasZ 'a = {..'a,  z: int};
```

Closed objects types could be:

```rust
type hasZ = {. z: int};
```

(Though those are less common, so I don't mind them looking a little strange.


I recall one reason why I preferred the `..` at the end of the record was because the following looks a little strange:


```rust
type hasZ 'a = {.., z: int} as 'a
```

With a comma after the `..`, and omitting the comma looks confusing


```rust
type hasZ 'a = {.. z: int} as 'a
```

Because it looks like you're extending `z`, instead of *appending* `z`.

SanderSpies:
I'm growing more fond of the `{=` and `{>` syntax for several reasons:
- `..` is too ambiguous like mentioned before
- `{=` and `{>` have a more heavy feeling which I like, while also being more descriptive (all subjective of course). And I don't think it clashes with existing syntax concepts.

freebroccolo:
@jordwalke those suggestions look pretty good to me. I agree that this case looks a bit wierd:

```rust
type hasZ 'a = {.. z: int} as 'a
```

But with the comma after the `..` I don't think it looks so bad, although I would still prefer the sugared version instead of the `{ … } as 'a` version.

I don't feel too strongly about which tokens are chosen at the moment but I do think that @SanderSpies makes a good point about `..`. The fact that we have `...` for one part in the syntax and `..` or `.` for a different part seems potentially confusing. I think it would be ideal if we could just stick to `...` but I guess that is problematic because we use curly brackets for both objects and records now.

So as far as I understand the current issues, I might suggest trying a slightly different approach. What if we change the syntax to something like this?

```rust
// open
type hasZ 'a = #{ ...'a,  z: int };

// closed
type hasZ = #{ z: int };
```

The advantages of this syntax are that it's still lightweight and looks close to the record syntax but it allows us to reuse the `...` token without ambiguity. It also doesn't require that the `...` be connected to the opening `{` just to be easier to handle for syntax highlighters. And hopefully the `#` is suggestive of intent since `#` is used for method invocation.

formatTest/typeCheckedTests/expected_output/mlSyntax.re
```diff
@@ -100,3 +100,9 @@ let module EM = {
 };
 
 exception Ealias = EM.E;
+
+let switch_ = "switch";
+
+let pub_ = "public";
```
jordwalke: Wait, I see a bug. What if the original `.ml` file had an identifier named `pub_`, and another identifier named `pub`. They both will get converted to `pub_` - which could not only fail to compile, but could also cause an unintentional bug. We always need to make sure that:

1. Every AST can be parsed in each syntax.
2.  In each syntax, `Parser(Printer(AST)) == AST` for all valid `AST`, for all syntaxes.

I think in this case for an AST where an identifier is "pub_" *in the AST*, it will be printed as `pub_`, and then it will be parsed into the AST as the identifier `pub`, which is not what we began with so it violated rule 2.

One easy way to that ensure that these rules hold, is we can simply *swap* the text *printing* of two already valid identifiers. For example, we could swap `mod` with `%`.
- When parsing the `%` token, we would always store it as identifier `"mod"` in the AST, and when we print we know to print the identifier `"mod"` as `%`. 
- When parsing the `mod` token, we would always store it as identifier `"%"` in the AST, and when we print we know to print the identifier `"%"` in the AST as `mod`.
- Therefore every AST is representable (rule number one), and `Parse(Printing(AST)) == AST`.

This case is a little different though because `%` and `mod` were both valid identifiers in each language. In our case, `public` is not a valid identifier in ML and `pub` is not a valid identifier in `Reason`. In either case, since ML is the canonical syntax, an AST that has an identifier named `"public"` is not a valid AST, so we don't need to even consider ASTs of that form.

That means, that we have one more option at our disposal. We can swap a *valid* identifier, with an identifier that could *never* occur. In `Reason`, we could parse the identifier `"public"` (as in `let public = "foo"`), and then store it in the AST as `"pub"`, then always ensure that `Reason` prints out the AST identifier `"pub"` as `"public"`. That allows every *valid* ML AST to be printed in Reason, and parsed back in Reason, and every valid `Reason` parse, to be printed in `ML` without any loss of information or changes in the AST. (note: If you parsed an ML file that had `let pub = ` it would be stored as `"pub"` in the AST and printed out in `Reason` as `let public =`.)

jordwalke: There's some utilities in here to do some of the swapping on identifiers: [here](https://github.com/facebook/reason/blob/ad1494ce5d3cb09590be3d0305949e5138407db4/src/syntax_util.ml).

bsansouci:
I like @freebroccolo's suggestion the best. Though I'm wondering if we could push it further and allow implicit open objects thanks to the `#` in front.
Something like `type bla = #{foo: int};` would really mean `type bla = #{...'a, foo : int};`.
I don't really know how the `...'a` can be used, but I'm guessing that you can restrict that "rest" to be a "concrete type" rather than "abstract" like `'a`? If so, then we definitely should support that syntax.

freebroccolo:
There is another advantage to using the `type hasZ 'a = #{ ...'a,  z: int }` syntax that I did not mention in the previous post. If we also used `#{` for object terms, this would allow us to elide the `pub` and reuse the syntax for record fields. Maybe we would want to adjust it a little still but in any case not having to write `pub` could be nice I think.

SanderSpies:
@freebroccolo I believe this PR already comes close to what you want (by mixing objects and records). I think we should do this in another PR though.

jordwalke:
@bsansouci: In `type hasZ 'a = #{ ...'a, z: int }` you can think of the `'a` as a type variable representing representing "the rest of the rows not named. That's a bit of an over-simplification, because it *actually* represents the *total* set of rows. Most of the time we just use it as "the rest of the rows". What's really going on is the following: `type hasZ 'a = `{...'otherRows, z:int} as 'a`. That is, there *is* a hidden variable for "and any other rows", but you can't bind it to a variable.  So therefore, you have to say "okay, and that *whole* set of rows should be considered `'a`, and we'll just be polymorphic in terms of the whole set, which is just as good. In Reason, we can simply make an over-simplification of that in the syntax because this is really the most common use case.

Regardless of the syntax, it allows you to be explicitly polymorphic in terms of the total set of rows in your annotations.

```
let myFunction :
  'a . #{...'a, z: int} => string =
    fun o => string_of_int o#z;
```

If you leave out the type definition, the row polymorphism will be inferred.

jordwalke:
A couple of thoughts on `#{}` vs `{ x: ..}`. I'm torn. Our feedback that we've gotten is that people don't like the hash. I do like that the hash is "up front", so you know immediately that the type is an object instead of record. Whereas with `{x: ..}` you have to wait until the end (imagine a long definition).

Things I like about the hash proposal:
- We can make object definitions resemble it without needing `pub/pri` `let x = #{bar: "hi"}`.
- I like that it's "up front" so when reading long definitions, you know immediate that they're not record definitions.

Things I dislike about the hash:
- You end up having to put `pri` anyways to mark some methods private.
- The mandatory `pub/pri` makes it possible to get rid of semicolons in objects (like JS).

I'm wondering  if we can get the best of both without any hashes at all. It might also simplify the parser greatly - it would no longer need to wait until the "end" to see what type of object/record it is.

```rust
/* Closed object type : These are very rare anyways */
type objType = {. doSomething: int => unit, foo: unit};

/* Row polymorphic object type - two dots. */
type objType 'a = {..'a, doSomething: unit => int, foo: unit};

/* Row polymorphic object type - two dots. */
type recordType = {doSomething: int => unit, foo: unit};

/* Faux-polymorphic records (one day?) Three dots */
type recordType 'a = {...'a, doSomething: int => unit, foo: unit};


/* Now if we give objects a comma, they look more like their definitions */
let myObject = {
  pub doSomething i => (),
  pub foo => ()
};
/* But the commas are optional! Printer will remove them  */
let myObject = {
  pub doSomething i => ()
  pub foo => ()
};

let myRecord = {doSomething: fun i => (), foo: ()};

/* We can also make the syntax for method dispatch be object.methodName: with
 * the trailing colon being attached to the method name like a label. They would be
 * highlighted as one token in the editor.
 */
let myRes = myObject.doSomething: 100;
let myRes = myObject.foo:;
let myRes = myRecord.doSomething 100;

```

The `methodName:` isn't easy to get right because we need to ensure that type annotations always have spaces around the `:` like `let res = (someValue.recordField : int)`.

But the big advantage is that we can cary the `label:` to named arguments like:

```
let res = myObject.doSomething: withNamedArg:10 andAnotherNamedArg:[];
```

I'm also realizing that my original objection to `type o 'a = {..'a, blah: int}` was that that it looks strange when writing it in the `as 'a` form `type o 'a = {.., blah: int} as 'a}`. But I think that wouldn't be an issue if we just always write/print `objType as 'a` as `{..'a, fields}`. That is, if you wrote `{.._, fields} as 'a`, it would even be printed as `{..'a, fields}`. When making type definitions (as opposed to annotations), you're always *forced* to acknowledge the row polymorphism anyways so it's not likely we'd get stuck with something like `{.._, fields}` where there is no `'a`. The only time you omit the type parameter is on inline definitions of *values*, where the type parameters are inferred.

```
let myFunc = fun (o: {.._, field: int}) => {
  o.field: + o.field:
};
``` 

In that case type of `myFunc` would be:
```
'unNamedType . {..'unNamedType, field: int} => int
```

We could provide special sugar for `{.._, field: int}` in the form of `{_, field: int}`.

bsansouci:
@jordwalke Thanks for writing all of this up! You inspired me to write down a bunch of thoughts.
Here's my (lengthy) object syntax proposal: https://gist.github.com/bsansouci/ca4dee603ed2bd8ed0a8c4e939f9b2d2.

It's basically inlined with what this PR did for type declaration, but reuses that for object declarations. My goal with this was to make using objects basically like a record as simple as possible. Because sometimes I need the row polymorphism and I don't really care for anything else about Objects.

jordwalke:
@bsansouci I replied on that gist.

jordwalke:
We can always evolve the object syntax rapidly, but for now the best result that incorporates all the feedback here seems to be:

`{..'a, fields}`, `{. fields}` for open and closed object types respectively, where `{..'a, fields}` is simply sugared rewriting of `{.. fields} as 'a`). I believe it may also make this diff much simpler to implement (you might be able to delete some parts of it that degrade readability of the parser?). For now, I think we can even just implement it as `{.. fields} as 'a` and save the super sugared `{..'a, fields}` for later. That would by itself make a #goodFirstTask.

jordwalke:
The only little weirdness to the resulting syntax would be that object values are of the form `{field ; field}`, but their types are of the form `{field, field}`. As a next step (in another diff) we should think about fixing that, by either making objects look more like records `{field, field}`, or something else.

SanderSpies:
The dots have been moved to the front. 

What's important to note is that I also changed `private` for types to `pri` (see also: https://caml.inria.fr/pub/docs/manual-ocaml/extn.html#sec220).

formatTest/typeCheckedTests/expected_output/mlSyntax.re
```diff
@@ -101,8 +101,6 @@ module EM = {
 
 exception Ealias = EM.E;
 
-let switch_ = "switch";
+let match = "match";
```
jordwalke: I see you took the opportunity to improve this?

SanderSpies: Yeah, basically:

.ml -> .re
switch -> match
pub -> method
pri -> private

and vice versa.

src/syntax_util.ml
```diff
@@ -83,6 +83,10 @@ let reason_to_ml_swapping_alist = [
   "\\===", "===";
   "!=", "<>";
   "!==", "!=";
+  "switch", "match";
+  "match", "switch";
```
jordwalke: I thought this list already implies the bidirectional. For example, the first entry ` "===",  "=="` means:

> In Reason, when you write `===`, it means the ML AST identifier `==`. Therefore, the ML identifier `==` is printed as `===` when printing Reason.

So, I *think* that you only need ` "match", "switch"`, which means:

> In Reason, when you write `match`, it means the ML AST identifier `switch`. Therefore, the ML identifier `switch` is printed as `match` when printing Reason.

There is *one* AST we are describing here - the ML AST.  In that ML AST, it is (apparently) impossible to express the identifier named `match`:

```ocaml
utop # let match = "asdf";;
Error: Syntax error
```

So therefore, it shouldn't be necessary to include `"switch", "match";`. Do I understand correctly?

docs/index.html
```diff
@@ -1021,6 +1021,81 @@
...
+##### Declaring an object type
+An object can have an object type to define its structure.
+
+```
+type tesla = {
+  .
+  color: string
+};
+```
+The extra dot at the end is to indicate that this is a closed object type, which means that
```
jordwalke: at the *beginning* now.

docs/index.html
```diff
@@ -1021,6 +1021,81 @@
...
+    hasEnvy.contents = envy
+  };
+};
+```
+This object is of object type tesla and has a public method `drive`. It also contains a
+private method `enableEnvy` that is only accesible from within the object.
+
+The following example shows an open object type which uses a type as parameter. The
+closed object type parameter is required to implement all the methods of the open object
+type.
```
jordwalke: Instead of "The closed object type parameter is required to implement all the methods of the open object type."

I think you should remove the word "closed" because it's distracting. It doesn't matter if the type parameter is closed or open, it just needs to implement all of the fields of `tesla`.

I have another long doc I've been working on that explains OCaml's row polymorphism.

formatTest/typeCheckedTests/expected_output/mlSyntax.re
```diff
@@ -100,3 +100,9 @@ module EM = {
 };
 
 exception Ealias = EM.E;
+
```
jordwalke: This mapping of all three is surprisingly clean.

src/syntax_util.ml
SanderSpies: This is an old diff, currently it says:

```
  "match", "switch";
  "method", "pub";
  "private", "pri";
```

formatTest/typeCheckedTests/expected_output/oo.re
```diff
@@ -90,101 +90,101 @@ and secondRecursiveClass init => {
...
 
 let constrainedAndCoerced = (
   [anonClosedObject, anonClosedObject]:
-    list < x : int, y : int > :>
-    list < x : int >
+    list {. x : int, y : int} :>
+    list {. x : int}
 );
 
 /* If one day, unparenthesized type constraints are allowed on the RHS of a
```
jordwalke: This comment no longer applies. yay.

formatTest/typeCheckedTests/expected_output/oo.re
```diff
@@ -90,101 +90,101 @@ and secondRecursiveClass init => {
...
 
 let coercedReturn = {
   let tmp = anonClosedObject;
-  (tmp :> < x : int >)
+  (tmp :> {. x : int})
 };
 
 let acceptsOpenAnonObjAsArg
-    (o: < x : int, y : int, .. >) =>
+    (o: {.. x : int, y : int}) =>
```
jordwalke: Thinking out loud: I'm concerned that `{.. x : int}` looks like `{..x, ` since there's no comma after the `..`. I think we can mitigate this by writing it as `{.. x: int}` (where the colon binds tight to the `x`.

formatTest/typeCheckedTests/expected_output/oo.re
```diff
@@ -90,101 +90,101 @@ and secondRecursiveClass init => {
...
 let res =
   acceptsClosedAnonObjAsArg {
-    method x = 0;
-    method y = 10
+    pub x = 0;
+    pub y = 10
   };
 
 /* TODO: Unify class constructor return values with function return values */
 class myClassWithAnnotatedReturnType
```
jordwalke: The one remaining syntax for objects that needs some love. (Class constructor return types). Another day!

formatTest/unit_tests/expected_output/wrappingTest.re
```diff
@@ -2051,16 +2051,16 @@ and mutuallyRecursiveTwo y => print_int y;
 /* let newMutualRecursionSyntax x => newMutuallyRecursiveTwo (x + x); */
 /* let newMutuallyRecursiveTwo y => print_int y; */
 /*  */
-type x = private int;
+type x = pri int;
 
-type myType 'a 'b 'c = private ('a, 'b, 'c);
+type myType 'a 'b 'c = pri ('a, 'b, 'c);
```
jordwalke: Do we have a test that proves that the following `Reason` code:

```rust
type private = (int, int);
```

Prints as ML:

```ocaml
type pri  = (int, int);
```

src/reason_pprint_ast.ml
```diff
@@ -5130,19 +5125,24 @@ class printer  ()= object(self:'self)
...
             let virtualName = [atom "virtual"; atom s.txt] in
             let openingTokens =
               (makeList ~postSpace:true ~inline:(false, true) ~break:IfNeed virtualName) in
-            label ~space:true (atom "method") openingTokens
+            label ~space:true (atom "pub") openingTokens
         in
         formatTypeConstraint opening (self#core_type ct)
       | Pcf_method (s, pf, Cfk_concrete (ovf, e)) ->
-        let methodText = if ovf == Override then "method!" else "method" in
+        let methodText = if ovf == Override then "method!" else
```
jordwalke: I think this is a bug! I believe we want to print `pri!` or `pub!`. There may be no test case that caught this? Could you at least add a small test case to catch this too?

formatTest/typeCheckedTests/expected_output/oo.re
SanderSpies: Seems like another `Good first task`.

formatTest/unit_tests/expected_output/wrappingTest.re
SanderSpies: I think this should be sufficient enough right?

https://github.com/SanderSpies/reason/blob/4d3b093c1631ff400c8d406a41c799a40d8eb6a9/formatTest/typeCheckedTests/expected_output/mlSyntax.re#L108

src/reason_pprint_ast.ml
SanderSpies: Ah cool! thanks for finding this one.

SanderSpies: Done.

jordwalke:
🎉

PR #768, opened 2016-09-15
[npm] dependencies with semver

yunxing:
We've pushed some of our dependencies to offical npm repo. This gives us the ability to specify ranges for a dependency that takes advantage of semver. 

WARNING: some existing project based on npm may break if this is merged. Those projects have to change their dependencies from a git url on npm-opam to the new semver version. Otherwise npm will try to dedup them.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,19 +46,20 @@
     },
     "name": "reason",
     "dependencies": {
-        "merlin": "https://github.com/npm-opam/merlin",
-        "re": "https://github.com/npm-opam/re",
-        "ocamlfind": "https://github.com/npm-opam/ocamlfind",
-        "BetterErrors": "https://github.com/npm-opam/BetterErrors",
-        "ocaml": "https://github.com/npm-opam/ocaml.git#npm-4.02.3",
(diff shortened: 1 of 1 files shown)


jordwalke:
Things to update:

- [`rebel`])(https://github.com/reasonml/rebel)
- Personal/internal projects.
- [`ExampleProject`](https://github.com/reasonml/ExampleProject)

Any others you can think of?

jordwalke:
We should probably also move `ocaml` to `opam-alpha` correct? (currently it's at `@reasonml/ocaml`). 


Also, `Reason` could be added to `@reasonml/reason`. Currently, we are all just pointing at the git URL but we'll want versioning soon. And do you think we could save the `npm` top level (non-scoped) `reason` package for when we're out of alpha stage?

yunxing:
everything should be under @opam-alpha now. reasonml was a deprecated scope that I just asked npm's support to remove.

jordwalke:
Maybe you can cancel that request? It would be nice to have our own scope (but maybe we can make a scope called reason?)

PR #771, opened 2016-09-16
AST+ (Part1)

yunxing:
This diff contains some first parts of AST+ work :
- Define an AST+ datatype by copying from latest ocaml trunk
- Change the parse to parse into AST+
- define conversion functions to convert AST+ to AST@(4.2.3). Unfortunately there is no magic to do it in a clean way, just some tedious work produced by some tediuos emacs scripts.

Work left to do:
- We've only finished the parser part (parse into AST+, convert to AST@4.2.3). We also want to do the symmatric part for printing (convert from AST@4.2.3 to AST+, print AST+).
- Need to do the same thing for 4.3.0 -- would be relatively easy with CPPO support I believe.
#706


yunxing:
This is fully functional now. Both parser and printer has support for the latest AST from OCaml/trunk, but can be fed into OCaml 4.2 compiler

The next step is to support OCaml 4.3 using CPPO

yunxing:
Rebased on top of jsx

yunxing:
Closing this in favor of https://github.com/let-def/migrate-parsetree, which is a more generic implementation that can benefit the whole community.

PR #775, opened 2016-09-20
<JSX> ☲  "Simplified" version of Sander's diff.

jordwalke:
Summary: This is a version of Sander's JSX Pull Request, but rebased on top of our recent large refactors, and modified to treat
JSX as "simple" constructs, meaning they are "clearly one unit", and do
not require parenthesis to be wrapped around them in order to be passed
as function arguments. That means you can treat <tag ...> as the
"opening paren" and "</tag>" as the closing paren.

This diff also supports:

   - Frags (which are simply lists that don't require commas separating
     contents).

   - Lists of JSX items

```
let myFunction =
   callSomeFunction
     <FirstItem>foo</FirstItem>
     <SecondItem>foo</SecondItem>;

let lst = [<Tag />, <AnotherTag />];

let lst =
  <>
    <Tag />
    <AnotherTag />
  </>;

```

Whats Not Implemented: Currently, the limitation of this approach is
that subsequent JSX elements must always have spaces separating them.
I have a good idea about how to solve that limitation, but we can start
using this right now, in an experimental mode.

```
let items = <Wrapper> <This /> <Must /> <Have /> <Spaces /> </Wrapper>
```

Test Plan:Test cases.

Reviewers:


formatTest/typeCheckedTests/expected_output/jsx.re
```diff
@@ -0,0 +1,368 @@
...
+     argument3=3
+     argument4=4
+     argument5=5
+     argument6="test">
+    <Namespace.Foo />
+  </LotsOfArguments>;
+
+let lowerCase = <div argument1=1 />;
+
+let a = <Foo  a>5 </Foo>;
```
bsansouci: Is this the string 5, or seen as `{5}`? (maybe this doesn't matter here actually...)

formatTest/typeCheckedTests/expected_output/jsx.re
```diff
@@ -0,0 +1,368 @@
...
+
+let a = <Foo  a>0.55 </Foo>;
```
bsansouci: `<Foo  a>` two spaces?

src/reason_parser.mly
```diff
@@ -2643,9 +2750,29 @@ _simple_expr:
...
       { mkexp (Pexp_variant ($1, None)) }
+  | jsx { $1 }
+  /*
+     Because [< is a special token, the <ident and <> won't be picked up as separate
+     tokens, when a list begins witha JSX tag. So we special case it.
+     (todo: pick totally different syntax for polymorphic variance types to avoid
+     the issue alltogether.
+     
+     first token
+     /     \
```
bsansouci: Should these be aligned with `<ident` and `args />`?

formatTest/typeCheckedTests/expected_output/jsx.re
jordwalke: Oh, that seems to be a mistake. Will fix.

formatTest/typeCheckedTests/expected_output/jsx.re
jordwalke: It's the number `5`. We may want to change how children are represented, or at least *printed*. Right now, the current JavaScript JSX conventions are a special case of what's supported. In `Reason`, the expression `{5}` is the same as `5` (verify for yourself - it's just a sequence expression with one item returned). So we could print it as `<Foo>{5}</Foo>.

src/reason_parser.mly
jordwalke: yes

jordwalke:
@bsansouci Want to start adapting this to Cheng Lou's needs?

jordwalke:
I'll commit this, and then if anyone has issues with it, I'll gladly fix them or revert it. Ben needs this in to rebase his changes off of.

PR #777, opened 2016-09-21
Part 2 <JSX> ☲ "Simplified" version of Sander's diff.

bsansouci:
Hey everyone, this is another PR related to JSX. Most of the work was done by @SanderSpies  and @jordwalke (see #775). 
This is building on top of them to allow pretty printing of calls to `React.createElement` to be pretty printed _back_ to JSX. This PR needs #775 to be in, because it'll probably cause merge conflicts that I'll have to solve by hand.

```rust
let c = React.createElement Foo.comp (Foo.props prop1::1 prop2::true ()) [|
  React.createElement Bar.comp Js.null [||]
|];
```

becomes

```js
let c = <Foo prop1=1 prop2=true> <Bar /> </Foo>
```

You can look at the tests for a couple more examples.

The reason to special cases those exact calls (with that signature) is that we can leverage Bucklescript's nice transpiling capabilities and transpile directly to normal react createElement calls in JS. This PR is only to allow to pretty print those calls back into a nicer JSX form. The current AST form that those JSX tags take are simpler and more universal, but for the sake of interop (for now), the form `React.createElement ComponentName.comp ...` is closer to JS React. We'll need some form of ppx to transform the current representation of those tags into one that BS will transpile better.


chenglou:
lucky number

jordwalke:
Before we get too far in comments, you should probably rebase on top of my diff that I just landed (not on Sander's branch which was outdated). Then you can force push to this PR.

bsansouci:
Done

chenglou:
Oooook so I'm gonna merge this. @bsansouci can you certify that this PR is legit and doesn't break Facebook.

chenglou:
Don't merge yet!

bsansouci:
Actually closing this. Most of this work was to support some workflow (converting from `React.createElement` to JSX) that isn't really needed in the end. It seems like @chenglou has a ppx that'll do a conversion from the `Foo.createElement prop1::1 ...` format to something that'll pretty print very well through Bucklescript ("very well" meaning, that interop with React.js). So we don't really need this PR anymore with @jordwalke and @SanderSpies work already in.

PR #778, opened 2016-09-21
Revert "[npm] dependencies with semver"

jordwalke:
Reverts facebook/reason#768

Reasons:

- All of the packages that are hosted on github npm-opam org aren't updated to point to versions on npm (instead they just point to master branch of npm-opam github).
- Even so, if you do `npm show @opam-alpha/ounit@2.0.0`, you'll see that version 2.0.0 points to a git tag github (which is how it should work) but that tag doesn't exist in this case.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,20 +46,19 @@
     },
     "name": "reason",
     "dependencies": {
-        "@opam-alpha/merlin": "^ 2.5.0",
-        "@opam-alpha/re": "^ 1.5.0",
-        "@opam-alpha/ocamlfind": "*",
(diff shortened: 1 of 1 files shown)


yunxing:
Did you run into any issues with the diff? 

The packages on github/npm-opam has master branch unchanged for backward compatibility since some people are still depending on them. 

Tags have already been generated to those repos last week. For example, see https://github.com/npm-opam/ounit/tree/2.0.0

jordwalke:
> The packages on github/npm-opam has master branch unchanged for backward compatibility since some people are still depending on them.

Okay, but feel free to break them. Very few people are depending on them right now.

PR #781, opened 2016-09-21
Basic versioning support

yunxing:
This PR provides basic versioning support in our opam file and package.json. 

After running "env version=x.y.z make release", package.json and opam will be regenerated with the new version number. Also, a new commit and a new tag will be generated and pushed to github. 

See https://github.com/facebook/reason/tree/1.2.3 for an example

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -70,15 +70,18 @@ update_error:
 
 .PHONY: build clean update_error compile_error
 
-VERSION      := $$(opam query --version)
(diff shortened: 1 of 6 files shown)


jordwalke:
This diff looks great! I think you just need to update the package.json dependencies to match your new `opam-alpha`, then we can merge.

jordwalke:
Feel free to merge whenever you like.

PR #783, opened 2016-09-22
Support attributes on class method names

bsansouci:
In OCaml when you do `type something = <foo : int [@bs.set]>`, the attribute actually gets set on the method `foo` instead of the type int, even if it looks like it's on the type. In Reason, we attach it on the type.
This seemed to be hard to fix because it required moving around a bunch of grammar rules (currently the attribute gets eaten up by a rule deep in the CFG, which ocaml fixed by having a whole other rule without those attributes called [poly_type_no_attr](https://github.com/ocaml/ocaml/blob/f872d67d4e97a6072f5891b5aef3c5f7e39e4a25/parsing/parser.mly#L2201)).
My fix is a simple one, simply make the syntax be `type something = <foo [@bs.set] : int>`. I think this is horribly confusing but it's a very quick temporary fix to unblock anyone trying to use those attributes. Buckle script won't work as intended without this fix because it won't see the attributes on the keys making `##` and `#=` basically useless.

### DIFF
--- a/formatTest/typeCheckedTests/input/attributes.re
+++ b/formatTest/typeCheckedTests/input/attributes.re
@@ -257,5 +257,10 @@ let module Js = {
 
 type classAttributesOnKeys = <
   key1 [@bs.set] : string,
-  key2 [@bs.get {null}] : Js.t int [@onType2]
+
+  /* The follow two are the same */
+  key2 [@bs.get {null}] : Js.t int [@onType2],
+  key3 [@bs.get {null}] : ((Js.t int) [@onType2]),
+
+  key4 : Js.t (int [@justOnInt])
 >;
--- a/formatTest/typeCheckedTests/input/attributes.re
+++ b/formatTest/typeCheckedTests/input/attributes.re
@@ -250,3 +250,12 @@ let myFun = fun ((X hello | Y hello) [@onOrPattern]) => hello;
 /* Bucklescript FFI item attributes */
 
 external imul : int => int => int = "Math.imul" [@@bs.val];
+
+let module Js = {
+  type t 'a;
+};
+
+type classAttributesOnKeys = <
+  key1 [@bs.set] : string,
+  key2 [@bs.get {null}] : Js.t int [@onType2]
+>;
--- a/src/reason_pprint_ast.ml
+++ b/src/reason_pprint_ast.ml
@@ -2706,7 +2706,11 @@ class printer  ()= object(self:'self)
               | [] -> label ~space:true
                       (label ~space:true (atom s) (atom ":"))
                       (self#core_type ct)
-              | _::_ -> makeList ~postSpace:true [atom s; (self#attributes attrs); atom ":"; self#core_type ct]
+              | _::_ ->
+                makeList
+                  ~postSpace:true
+                  ~break:IfNeed
+                  [atom s; (self#attributes attrs); atom ":"; self#core_type ct]
             )
           in
           let openness = match o with
--- a/src/reason_pprint_ast.ml
+++ b/src/reason_pprint_ast.ml
@@ -2701,13 +2701,13 @@ class printer  ()= object(self:'self)
             makeList ~wrap:("(",")") ~sep:"," ~postSpace:true ~break:IfNeed (List.map (self#core_type) l)
         | Ptyp_object (l, o) ->
           let core_field_type (s, attrs, ct) =
-            self#attach_std_attrs
-              attrs
-              (
-                label ~space:true
-                  (label ~space:true (atom s) (atom ":"))
-                  (self#core_type ct)
-              )
+            let l = extractStdAttrs attrs in
+            (match l with
+              | [] -> label ~space:true
+                      (label ~space:true (atom s) (atom ":"))
+                      (self#core_type ct)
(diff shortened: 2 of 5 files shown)


jordwalke:
And how do you add an attribute to the actual type (in Reason and OCaml)

bsansouci:
In OCaml:
```rust
/* the attributes is set on the whole method called bla */
type a = <bla : int [@bs.set]>;;

/* the attributes is set on the return type int of the method bla */
type a = <bla : (int [@bs.set])>;;

/* syntax error */
type a = <bla [@bs.set] : int>;;
```

In Reason:
```rust
/* the attributes is set on the return type int of the method bla */
type a = <bla : int [@bs.set]>;;

/* With this PR, the attributes is set on the whole method called bla.
 * Without this PR this is a syntax error.
 **/
type a = <bla [@bs.set] : int >;;

/* syntax error */
type a = <bla : (int [@bs.set])>;;
```

Basically, the ideal fix is to make OCaml and Reason behave the same way when adding `[@bs.set]` without parens, next to the return type (ie: it should be added on the method because that's more likely to be what you want AND it'll annoy newcomers who are trying BS because the manual's in OCaml).
My PR isn't a fix, it's just one way to get attributes on method so I can move forward on [tapknight](http://github.com/bsansouci/tapknight). I don't know enough of how grammars work to fix this I think, because the problem is [this rule](https://github.com/facebook/reason/blob/4964791c3ead9f12cc098d937d247d413a031167/src/reason_parser.mly#L4124) has an attribute which gets reduced before the [field](https://github.com/facebook/reason/blob/4964791c3ead9f12cc098d937d247d413a031167/src/reason_parser.mly#L4241) one does.

jordwalke:
```
/* the attributes is set on the whole method called bla */
type a = <bla : int [@bs.set]>;;
```

First you said it was set on the key only, but that comment makes it sounds like it's set on the entire "row". Can you clarify so I can make sense of the rest of the diff?

bsansouci:
Oh sorry I called the key `key` and I said `the method key`. Sorry for that confusion!

jordwalke:
I don't see why the following has to be a syntax error. Generally, whenever you have `term attr`, having the `attr` being applied to the term, you can put parens around the entire thing, and it doesn't change anything.

```
/* syntax error */
type a = <bla : (int [@bs.set])>;;
```

Like:

```
let x = (4 : int [@attr]);
let x = (4 : (int [@attr]);
```

Should be parsed the same.

formatTest/typeCheckedTests/input/attributes.re
```diff
@@ -250,3 +250,12 @@ let myFun = fun ((X hello | Y hello) [@onOrPattern]) => hello;
...
 
 external imul : int => int => int = "Math.imul" [@@bs.val];
+
+let module Js = {
+  type t 'a;
+};
+
+type classAttributesOnKeys = <
+  key1 [@bs.set] : string,
+  key2 [@bs.get {null}] : Js.t int [@onType2]
```
jordwalke: So `[@onType2]` applies to `Js.t int`.

How do we apply an attribute *only* to `int`? I think there should be a test case for it.

Also, as I suggested in the discussion, I think we should also be able to do `(Js.t int [@onType2])`.

jordwalke:
After looking at your diff, I actually think the `int [@attr]` will be parsed correctly and not be considered a syntax error as you previously suggested. Furthermore, `(int [@attr])` should also be equivalent. The only thing is that you need to add test cases confirming that it parses correctly and add test cases showing that two examples with and without parens, both render to the same output.

src/reason_pprint_ast.ml
```diff
@@ -2701,13 +2701,13 @@ class printer  ()= object(self:'self)
...
-                label ~space:true
-                  (label ~space:true (atom s) (atom ":"))
-                  (self#core_type ct)
-              )
+            let l = extractStdAttrs attrs in
+            (match l with
+              | [] -> label ~space:true
+                      (label ~space:true (atom s) (atom ":"))
+                      (self#core_type ct)
+              | _::_ -> makeList ~postSpace:true [atom s; (self#attributes attrs); atom ":"; self#core_type ct]
```
jordwalke: Should at least put a `~break:IfNeeded` here.

formatTest/typeCheckedTests/input/attributes.re
bsansouci: Added.

src/reason_pprint_ast.ml
bsansouci: Fixed.

formatTest/typeCheckedTests/input/attributes.re
```diff
@@ -250,3 +250,17 @@ let myFun = fun ((X hello | Y hello) [@onOrPattern]) => hello;
...
+let module Js = {
+  type t 'a;
+};
+
+type classAttributesOnKeys = <
+  key1 [@bs.set] : string,
+
+  /* The follow two are the same */
+  key2 [@bs.get {null}] : Js.t int [@onType2],
+  key3 [@bs.get {null}] : ((Js.t int) [@onType2]),
```
jordwalke: Yes, this is a good test case.

PR #789, opened 2016-09-23
Fix pretty printing jsx components with other attributes

### LINKED ISSUE #788: JSX formatting with extra attribute is wrong
Repro:

```
let asd = Bar.createElement foo::1 bar::2 ["a", "b"] [@JSX] [@foo];
```

Formats to:

```
let asd = Bar.createElement foo::1 bar::2 ["a", "b"] [@foo];
```

```sh
echo 'let asd = Bar.createElement foo::1 bar::2 ["a", "b"] [@JSX] [@foo];' | refmt -parse re -print re -use-stdin true -is-interface-pp false
```

@bsansouci @jordwalke

bsansouci:
This should fix #788.

Check the tests for examples.

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/jsx.re
+++ b/formatTest/typeCheckedTests/expected_output/jsx.re
@@ -399,6 +399,19 @@ let asd2 =
   [@JSX]
   [@foo];
 
+let span
+    test::(test: bool)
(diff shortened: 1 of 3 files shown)


chenglou:
Can you also add the equivalent tests for lower-case ones? Thx.

PR #791, opened 2016-09-24
Check in Reactjs ppx transform

chenglou:
This converts our ideal JSX `createElement` transform into one that works with the current ReactJS, compiled through BuckleScript (actual bindings coming soon! Wanted to stress test it a bit more beforehand).

The locations are wrong I think; I'll fix them later.


chenglou:
Merging. Will take retrospective feedback.

PR #792, opened 2016-09-24
Add shrinkwrap file for reason

yunxing:
Since reason only supports 4.2.3 for now, we need to make sure all dependencies installed with reason can support 4.2.3

### DIFF
--- a/npm-shrinkwrap.json
+++ b/npm-shrinkwrap.json
@@ -1,0 +1,8 @@
+{
+  "name": "reason",
+  "dependencies": {
+    "@opam-alpha/camlp4": {
+      "version": "4.4.27"
+    }
+  }
+}


yunxing:
Merging for now

jordwalke:
Why is a shrinkwrap file needed for this?
The specification in our `package.json` should over this:
```
"@opam-alpha/ocaml": "= 4.02.3",
```

jordwalke:
IIRC, the shrinkwrap is only intended to give more influence over *which* of many valid possible package versions to use, or to override it completely.

- I believe it should never be used to provide a correct implementation that would otherwise cause a breakage.
- I believe it won't be picked up by *dependers* of `reason`, so it is something you typically put at the top level of your app (pretty sure).

jordwalke:
After rebasing, a simple `npm install` inside of `reason` root directory gives the following result:

screen shot 2016-09-25 at 1 32 43 am

The package sync script might have picked up something that busted the constraints implied by the shrinkwrap.  I could see that happening if some other package has a "*" dependency and the sync script picked it up. Deleting the shrinkwrap made the issues go away.

Projects like `ExampleProject` aren't effected because shrinkwraps are only taken into account for top level projects.

I think we want to remove the shrinkwrap anyways right?

PR #793, opened 2016-09-24
Fix formatting in readme

carlosrocha:


jordwalke:
If the answer is "no", I'll just merge. Thanks for fixing!

carlosrocha:
It should be unindented with triple ticks, I will fix it.

PR #796, opened 2016-09-26
Change `let module Foo` into `module Foo`.

SanderSpies:
Related to #26.

Note that I chose to use `module` instead of the proposed `mod` keyword as I think it's more descriptive and clear. Of course if we want to use `mod` anyway I can still change it into that.

### DIFF
--- a/formatTest/unit_tests/expected_output/modules.re
+++ b/formatTest/unit_tests/expected_output/modules.re
@@ -586,3 +586,7 @@ open M;
 open M.Inner;
 
 open M;
+
+module OldModuleSyntax = {
+  module InnerOldModule = {};
+};
--- a/formatTest/unit_tests/expected_output/modules.re
+++ b/formatTest/unit_tests/expected_output/modules.re
@@ -24,7 +24,7 @@ let run () => TestUtils.printSection "Modules";
  * values. Modules, on the other hand do allow fields to reference other
  * fields. Appropriately, each field is introduced via the keyword `let`.
  */
-let module MyFirstModule = {
+module MyFirstModule = {
   let x = 0;
   let y = x + x;
 };
@@ -33,7 +33,7 @@ let result = MyFirstModule.x + MyFirstModule.y;
 
 
 /**
- * - A module is introduced with the `let module` phrase.
+ * - A module is introduced with the `module` phrase.
  * - A module *must* have a capital letter as its first character.
  * - The exported fields of a module must be listed within `{}` braces and each
  * exported value binding is specified via a `let` keyword.
@@ -43,7 +43,7 @@ let result = MyFirstModule.x + MyFirstModule.y;
  * Another way that modules are more powerful than records, is that they may
  * also export types.
  */
-let module MySecondModule = {
+module MySecondModule = {
   type someType = int;
   let x = 0;
   let y = x + x;
@@ -76,7 +76,7 @@ module type MySecondModuleType = {
  * written code that matches your understanding. For example, `MySecondModule`
  * could have been written as:
 
- let module MySecondModule: MySecondModuleType = {
+ module MySecondModule: MySecondModuleType = {
  type someType = int;
  let x = 0;
  let y = x + x;
@@ -98,12 +98,12 @@ module type MySecondModuleType = {
  * language.
  */
 let opensAModuleLocally = {
-  let module MyLocalModule = {
+  module MyLocalModule = {
     type i = int;
     let x: i = 10;
   };
   /* Notice how local modules names may be used twice and are shadowed */
-  let module MyLocalModule: MySecondModuleType = {
+  module MyLocalModule: MySecondModuleType = {
     type someType = int;
     let x: someType = 10;
     let y: someType = 20;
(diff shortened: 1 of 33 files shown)


jaredly:
why do we want `module` over `let module`?

chenglou:
Yeah, `module` over `mod`.

@jaredly I guess it's kinda how it's `type foo` rather than `let type foo`? Not much opinion on this though.

jaredly:
I guess it alights with `module type`

yunxing:
Looks good. 

Could you make this "let" work optional instead? This way we can have backward compatibility given that we don't have a good upgrade story for npm yet.

SanderSpies:
@yunxing Done, you can now use `let module` also.

formatTest/unit_tests/expected_output/modules.re
```diff
@@ -98,12 +98,12 @@ module type MySecondModuleType = {
  * language.
  */
 let opensAModuleLocally = {
-  let module MyLocalModule = {
+  module MyLocalModule = {
```
jordwalke: Great, was looking for this.

jordwalke:
Cool, I like this improvement. Nice job on the backwards compat too!

@jaredly:
> I guess it alights with module type too 

Good point!

SanderSpies:
@jordwalke Yeah, that would also be my preference - I think with some shuffling around of LET it should be possible. 

I don't think moving to different syntax should be a big issue, the current changes are not that different from the old situation.

SanderSpies:
Seems moving to LET UIDENT is even simple!

jordwalke:
Cool, glad to move forward quickly with this one especially since you've got the backwards compat. If noone objects, I'll merge first thing in the morning (or yunxing, feel free to when you see this).

the `let ModuleName` has an issue in that it conflicts with `let None = None`, but we can work that out way later. Really happy with this for now.

SanderSpies:
@jordwalke Guess I was a bit too optimistic - I now have a weird mixture of modules and functions. 

I think we need to do something similar as with the objects PR - which is creating a production to decide if it's a module or variable based on `LET UIDENT` and `LET LIDENT` (including all the variations).

freebroccolo:
> the `let ModuleName` has an issue in that it conflicts with `let None = None`, but we can work that out way later.

A disadvantage to this approach is that it makes accurate syntax highlighting harder. In my vscode extension, I highlight module names differently than data constructors in patterns. But with such a change there'd be no way to distinguish them syntactically in general without backtracking, which textmate style grammars do not support.

freebroccolo:
To give a bit more context to my previous comment, this is how I am currently highlighting things in the vscode extension:

highlights

So perhaps another point to consider is that with the current grammar there is a nice symmetry between `let` prefixed items corresponding roughly to definitions with some sort of runtime significance, e.g., functions, modules, objects, etc. The items that do not have a `let` prefix, like `{class, module} type` correspond more to items with static significance, e.g., types, module signatures, or classes.

Removing the `let` from `let module` would break this symmetry and there would also be no way to highlight the `module` keyword in `module M` and `module type T` differently in a robust way (e.g., works even with line breaks) due to the limitations of textmate grammars. Emacs probably has the same issue but it's been awhile since I wrote a higlighter for it so I don't remember exactly what it's capable of.

I'm not trying to suggest this current change shouldn't be made btw. I don't really have a strong opinion about it (especially as long as backward compatibility is there). But I do think that syntax changes should be made with the potential to highlight them effectively kept in mind, since there are a lot more limitations on what can be recognized there than in the actual language parser.

freebroccolo:
One other thing: I think the fact that we currently have `let module` for modules but `class type` and `module type` for classes and signatures kind of obscures the symmetry I was talking about. It would be more apparent if we also had `type class C = …` and `type module T = …`.

jordwalke:
This is a great PR. Backwards compat and everything.

@freebroccolo's point about taking it even further with `let ModuleName` is a good one. We can discuss that later (it would require a different syntax for `let None`, such as `let (None)` which is not out of the question - but we can discuss later).

jordwalke:
@freebroccolo's other point:

> there would also be no way to highlight the module keyword in module M and module type T differently in a robust way

Is this really the case? It seems highlighting is seriously underpowered!

freebroccolo:
> Is this really the case? It seems highlighting is seriously underpowered!

@jordwalke Unfortunately, yes. The problem stems from the fact that textmate grammars do the tokenization pass line-by-line for efficiency purposes (easy but crude) and textmate grammars do not support multi-line regexes. What this means is that you effectively only have backtracking up to what is possible within a single-line regex.

You can specify that a syntactic entity has a start and end pattern and that entity can span across multiple lines. But once a start pattern matches, that's it. You can't back up and try another if the end pattern fails to match. The highlighting will just fail for the rest of the document in that case.

So for accurate highlighting you always need to be able to determine which parsing contexts will come next before you finish the line, or just try to do some sort of approximate parsing.

So when I say "robust" what I mean is that you could highlight the `module` keyword in `module M` and `module type T` differently but only if `type` always occurs on the same line as `module`, otherwise the highlighting will just fail.

In practice it may not be a big deal but I personally find it pretty annoying when highlighting is fragile like that. So I've gone to great lengths to ensure almost all parts of the current syntax highlighting are stable even when the tokens for a syntactic entity are broken across lines.

chenglou:
(Notified in discord) Despite the increase in difficulty in highlighting, I think the syntax itself shouldn't be dragged down (to a certain degree) by it. I'm merging this; Apologies to @freebroccolo for the harder syntax highlighting, and thanks for your work so far. Really appreciate it.

Issue #794, opened 2016-09-25
Don't break line on jsx prop if prop list is short


chenglou:
```html
<div
  className=foo>
  ...
</div>
```

This should look like below instead:

```html
<div  className=foo>
  ...
</div>
```

Just like function calls; don't break line if the arguments are short enough.

Issue #787, opened 2016-09-22
[EASY] Formatting nested JSX is wonky.


jordwalke:
This is how it formats. Should be an easy fix.
```
    <>
      <Nav.Nav onNavTap=(delegate handleNavTap) navWidth=mainScreen.bounds.size.width navHeight=62.0 /> <StoryTableViewDisplay
                                                                    contentOffsetX=0.0
                                                                    contentInsetX=0.0
                                                                    contentOffsetY=0.0
                                                                    contentInsetY=contentInset
                                                                    height=mainScreen.bounds.size.height
                                                                    width=mainScreen.bounds.size.width
                                                                    rowWidth=mainScreen.bounds.size.width
                                                                    rowHeight=cardHeight
                                                                    rowSpacing=20.0
                                                                    rowDataSourceForFrameCount
                                                                    />
    </>
```

jordwalke:
It's a good one to save as a first task, but it's bothering me so I'll just fix it now.

chenglou:
@jordwalke is this one also related?

```
<div
  className=foo>
  ...
</div>
```

jordwalke:
I don't think that's related.

jordwalke:
what do you expect the formatting to be in that case?

chenglou:
<div className=foo>
  ...
</div>

jordwalke:
Sorry, what?

jordwalke:
Or I think that's the argument breaking logic that has arguments break if children do. We can tweak it.

chenglou:
Moved this into #794

PR #798, opened 2016-09-28
[Reason] JSX: Breaking Bad 🚌 

jordwalke:
[Reason] JSX: Breaking Bad 🚌
Summary: Fix the bad breaking of JSX tags.

Two improvements here:

- Attributes now break independently of children (fixes https://github.com/facebook/reason/issues/794)
- Fix wonky incorrect breaking (fixes https://github.com/facebook/reason/issues/787)

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/jsx.re
+++ b/formatTest/typeCheckedTests/expected_output/jsx.re
@@ -92,6 +92,7 @@ let module Namespace = {
   let module Foo = {
     let createElement
         intended::intended=?
+        anotherOptional::x=100
         children => {
       displayName: "test"
     };
@@ -167,7 +168,8 @@ let a3 = <So> <Much> <Nesting /> </Much> </So>;
 
 let a4 =
   <Sibling>
-    <One test=true foo=b /> <Two foo=b />
+    <One test=true foo=b />
+    <Two foo=b />
   </Sibling>;
 
 let a5 = <Foo> "testing a string here" </Foo>;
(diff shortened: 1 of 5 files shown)


formatTest/typeCheckedTests/expected_output/jsx.re
```diff
@@ -415,3 +417,67 @@ let asd2 = video test::false 10 [@JSX] [@foo];
...
+    <Namespace.Foo
+      intended=true
+      anotherOptional=200
+    />
+    <Namespace.Foo
+      intended=true
+      anotherOptional=200
+    />
+    <Namespace.Foo
+      intended=true anotherOptional=200>
```
jordwalke: @chenglou Only downside is that it will first try to wrap all the attributes like this, before breaking each attribute onto their own line. I can actually get used to this.

chenglou: Humm can this be fixed? Seeing this is a bit more satisfying:

```html
<div foo=bar1 />
<div foo=bar2 />
<div foo=bar3 />
```

Not an issue for now though.

jordwalke: They will typically render as your `div`s. It's just that for multiple attributes, they will first break as `Namespace.Foo` does above, and then if they still don't have space, they will quickly begin to break as:

```
<Namespace.Foo
  intended=true
  another=false>
  <child />
</Namespace.Foo>
```

There's just a chance that it breaks as:

```
<Namespace.Foo
  intended=true anotherOptional=200>
```

If it saves lines.

chenglou: Ah I misunderstood then, thanks.

PR #800, opened 2016-09-29
Add `of` keyword back as optional

yunxing:
This simple change here increases tons of backward compatibility to existing code.

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/basics.re
+++ b/formatTest/typeCheckedTests/expected_output/basics.re
@@ -6,6 +6,13 @@ type reasonXyz =
   | Q
   | R;
 
+type reasonXyzWithOf =
+  | X
+  | Y int int int
+  | Z int int
+  | Q
+  | R;
+
 let reasonBarAs =
   fun
   | ((Y _ | Z _) as t, _) => {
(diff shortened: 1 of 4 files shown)


bsansouci:
Hey @yunxing, would it be possible to run `upgradeSyntax.sh` instead of adding more optional syntax for backwards compatibility reasons? Maybe I'm misunderstanding the purpose of this PR, but I was under the impression that Reason was trying really hard to avoid the baggage of supporting legacy code by offering a good syntax upgrade path. Why is this case different?

yunxing:
@bsansouci You are totally right. That's the goal. It is just ... we don't have a good upgrade story yet for npm installation flow (it is being discussed though).

And also this case is special. Our initial version 0.0.6 uses "of" keyword, and there is a change that prevents upgrading from that version to the current version (the ast is still same, but our comment source map are a bit different now).  

All that being said, I think it is still nice to preserve some sort of backward compatibility when we could easily do (there are still tons of code in python2 although there is a tool to upgrade 2 to 3). But I definitely think we should encourage users to do more syntax upgrade when we are ready. Making this a habit for users will help us a lot in the long run.

jordwalke:
Yeah, if it's pretty much free, may as well!

PR #805, opened 2016-10-11
Add spaces after "<" and before ">" for obj type declaration.

bsansouci:
This is to avoid conflicting with the JSX syntax. It's a quick fix
intended to be temporary until the JSX stuff is more mature.

`refmt` is broken without this because `type a = < x : int >` becomes `type a = <x : int>` which then doesn't parse because `refmt` think it's some malformed JSX.

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/attributes.re
+++ b/formatTest/typeCheckedTests/expected_output/attributes.re
@@ -336,4 +336,4 @@ type classAttributesOnKeys = <
   key2 [@bs.get null] : (Js.t int) [@onType2],
   key3 [@bs.get null] : (Js.t int) [@onType2],
   key4 : Js.t (int [@justOnInt])
->;
+ >;
--- a/src/reason_pprint_ast.ml
+++ b/src/reason_pprint_ast.ml
@@ -2718,7 +2718,10 @@ class printer  ()= object(self:'self)
             | Open -> [atom ".."]
           in
           let rows = List.concat [(List.map core_field_type l); openness] in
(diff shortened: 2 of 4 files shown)


yunxing:
Looks good, thanks for fixing!

formatTest/typeCheckedTests/expected_output/attributes.re
```diff
@@ -336,4 +336,4 @@ type classAttributesOnKeys = <
   key2 [@bs.get null] : (Js.t int) [@onType2],
   key3 [@bs.get null] : (Js.t int) [@onType2],
   key4 : Js.t (int [@justOnInt])
->;
+ >;
```
jordwalke: This causes strange indenting though! But don't worry, @SanderSpies diff for new syntax for object types makes this obsolete anyways (and we can delete @yunxing's token splitting hack in the lexer/parser).

src/reason_pprint_ast.ml
```diff
@@ -2718,7 +2718,10 @@ class printer  ()= object(self:'self)
             | Open -> [atom ".."]
           in
           let rows = List.concat [(List.map core_field_type l); openness] in
-          makeList ~break:IfNeed ~postSpace:true ~wrap:("<", ">") ~sep:"," rows
+          if List.length rows = 0 then
+            atom "<>"
+          else
+            makeList ~break:IfNeed ~postSpace:true ~wrap:("< ", " >") ~sep:"," rows
```
jordwalke: There's a `~pad:(true, true)` argument which doesn't mess up formatting like my other comment pointed out.

PR #806, opened 2016-10-13
GraphQL

SanderSpies:
WORK IN PROGRESS.

This is just syntax for now - it doesn't work or anything.

I would love to get feedback on this already before I continue with implementing the other details.

### DIFF
--- a/formatTest/typeCheckedTests/input/graphql.re
+++ b/formatTest/typeCheckedTests/input/graphql.re
@@ -1,0 +1,59 @@
+/* TODO: add types for queries below here */
+
+/**
+* - the queries should type check against normal types.
+* - but should also result in well-formed strings
+* - bummer: we must supply the type of fields
+*/
+
+query basic => {
+    hero {
+        name: string
+    },
+    droid {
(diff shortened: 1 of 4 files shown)


kayceesrk:
A general question about such language extensions is whether they should be part of the core language syntax or ppx syntax extensions. Is there something in the proposed extension that would make it tedious to describe them as a ppx extension. Wondering if you had any thoughts on this?

SanderSpies:
@kayceesrk I share your concerns, and I'm not sure if this is the right idea. So far I'm leaning more towards staying close to the actual GraphQL query syntax (but still giving us the freedom to improve where possible).

PPX is definitely an interesting option:
```
let%query example = {
  hero: {
    name: string
  },
  droid: {
    name: string
  }
};
```
Although I'm not sure how we would add support for [fragments](http://graphql.org/learn/queries/#fragments) here.

kennetpostigo:
@SanderSpies this is something that I mentioned [here](https://github.com/andreas/ocaml-graphql-server/issues/8)

PR #810, opened 2016-10-16
[rebuild]use ocamlc.opt

yunxing:
ocamlc.opt is faster

cc @jordwalke


yunxing:
woah CI is so fast today.

Issue #811, opened 2016-10-17
adding a comment results in poor code formatting


jberdine:
When adding a comment to:

```
let hash {csu, mgd} => Hashtbl.hash (csu, mgd);
```

I would like to get:

```
let hash {csu, mgd} =>
  /* exclude the scope from the hash since it is ignored by equal */
  Hashtbl.hash (csu, mgd);
```

but currently (1.4.0) I get:

```
let hash {csu, mgd} =>
  /* exclude the scope from the hash since it is ignored by equal */
  Hashtbl.hash (
  csu,
  mgd
);
```

jordwalke:
Good find.

yunxing:
Good find. Taking a look

yunxing:
![image](https://cloud.githubusercontent.com/assets/1017689/19457027/4f9afd22-9479-11e6-9594-11311c158a34.png)

Looks like the highlighted area should be merged as direct siblings in the format tree.

chenglou:
Solved by ^

PR #812, opened 2016-10-18
Put function name and arguments into the same tuple

yunxing:
#811

Currently in a function application, the function name and the arguments are not in the same label (see screenshot I put up in #811). This creates weird formatting, especially when we add comments on top of it. This diff put them under the same label. 

cc @jordwalke

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/oo.re
+++ b/formatTest/typeCheckedTests/expected_output/oo.re
@@ -157,21 +157,24 @@ let acceptsClosedAnonObjAsArg
     (o: < x : int, y : int >) =>
   o#x + o#y;
 
-let res = acceptsOpenAnonObjAsArg {
-  method x = 0;
-  method y = 10
-};
+let res =
+  acceptsOpenAnonObjAsArg {
+    method x = 0;
+    method y = 10
+  };
 
-let res = acceptsOpenAnonObjAsArg {
-  method x = 0;
-  method y = 10;
-  method z = 10
-};
+let res =
+  acceptsOpenAnonObjAsArg {
+    method x = 0;
+    method y = 10;
+    method z = 10
+  };
 
-let res = acceptsClosedAnonObjAsArg {
-  method x = 0;
-  method y = 10
-};
+let res =
+  acceptsClosedAnonObjAsArg {
+    method x = 0;
+    method y = 10
+  };
 
 /* TODO: Unify class constructor return values with function return values */
 class myClassWithAnnotatedReturnType
@@ -284,12 +287,8 @@ let module HasTupleClasses: {
 
 class intTuples = class tupleClass int int;
 
-class intTuplesHardcoded = (
-  class tupleClass int int
-) (
-  8,
-  8
-);
+class intTuplesHardcoded =
+  (class tupleClass int int) (8, 8);
 
 
 /**
--- a/formatTest/unit_tests/expected_output/basicStructures.re
+++ b/formatTest/unit_tests/expected_output/basicStructures.re
@@ -1,5 +1,6 @@
 /* Copyright (c) 2015-present, Facebook, Inc. All rights reserved. */
-let run () => TestUtils.printSection "Basic Structures";
+let run () =>
+  TestUtils.printSection "Basic Structures";
 
 while something {
   print_string "You're in a while loop";
@@ -312,7 +313,8 @@ let annotatingFuncApplication = (
 );
 
 /* Pretty printer might stick the [int] at the label. */
-let annotatingSingleFuncApplication: int = _dummyFunc "a";
+let annotatingSingleFuncApplication: int =
+  _dummyFunc "a";
 
 /* So lets try a place where it won't */
 let annotatingSingleFuncApplication = {
@@ -604,10 +606,11 @@ let anotherRecord = {
 };
 
 let anotherRecord = {
-  ...SomeReally.longFunctionCall {
-    passingRecordField: 0,
-    andThisOtherRecordField: 10
-  },
+  ...
+    SomeReally.longFunctionCall {
+      passingRecordField: 0,
+      andThisOtherRecordField: 10
+    },
   name: "joe++",
   age: testRecord.age + 10
 };
--- a/formatTest/unit_tests/expected_output/modules.re
+++ b/formatTest/unit_tests/expected_output/modules.re
@@ -188,13 +188,15 @@ let module MyFunctor (M: HasTT) => {
    bottom of this file. [Actually, forgiving the trailing SEMI might not be
    such a great idea].
    */
-let module MyFunctorResult = MyFunctor {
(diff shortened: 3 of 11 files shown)


src/reason_pprint_ast.ml
```diff
@@ -2218,10 +2218,9 @@ let formatAttachmentApplication finalWrapping (attachTo: (bool * layoutNode) opt
           | (hd::tl, Some (useSpace, toThis)) ->
             (* TODO: Can't attach location to this - maybe rewrite anyways *)
             let attachedArgs = makeAppList attachedList in
-            label
-              ~space:true
-              (label ~space:useSpace toThis attachedArgs)
-              wrappedListy
+              (label ~space:useSpace toThis (label
```
yunxing: @jordwalke I'm not sure if this is the right solution, you probably have more context than me in this code.

formatTest/unit_tests/expected_output/basicStructures.re
```diff
@@ -1,5 +1,6 @@
 /* Copyright (c) 2015-present, Facebook, Inc. All rights reserved. */
-let run () => TestUtils.printSection "Basic Structures";
+let run () =>
```
jordwalke: This seems right because there was no breaking opportunity before, so this diff also fixed something that wasn't working as intended.

formatTest/typeCheckedTests/expected_output/oo.re
```diff
@@ -157,21 +157,24 @@ let acceptsClosedAnonObjAsArg
...
-  method x = 0;
-  method y = 10
-};
+let res =
+  acceptsOpenAnonObjAsArg {
+    method x = 0;
+    method y = 10
+  };
 
-let res = acceptsOpenAnonObjAsArg {
```
jordwalke: I think the old version is more conventional (FB's conventions at least).

```rust
let myComponent = React.CreateComponent {

};
```

Because it's under `50` characters, it shouldn't have to break to the next line.

jordwalke: There is special logic to handle this, in fact and this diff probably adds something that makes that logic be ignored.

formatTest/unit_tests/expected_output/modules.re
```diff
@@ -482,12 +483,13 @@ Printf.printf
  </div>;
  };
  */
-include YourLib.CreateComponent {
```
jordwalke: And I think this is better before too.

formatTest/unit_tests/expected_output/syntax.re
```diff
@@ -592,10 +591,10 @@ let result =
...
 let result =
   myRecordWithFunctions.addThreeNumbersTupled (
-  10,
-  20,
-  30
-);
+    10,
+    20,
+    30
+  );
```
jordwalke: I *like* the new version better, but I think this is pretty rare in practice.

formatTest/unit_tests/expected_output/wrappingTest.re
```diff
@@ -546,15 +548,16 @@ let reallyLongFunctionNameThatJustConcats a =>
 
 let seeHowLongValuesWrap = {
   age: 30,
-  name: reallyLongFunctionNameThatJustConcats [
```
jordwalke: I think the old one is better here.

src/reason_pprint_ast.ml
jordwalke: Some context: `wrappedListy` is for when we apply the special formatting logic I mentioned, to detect that a final function argument is list-like in nature, and therefore can "dock" to the entire binding etc (which matches common calling convention):

```rust
let x = myFunction {

};
```

jordwalke: I wonder if the real underlying cause of the comment formatting was due to a lack of `SourceMap` node that was needed.

formatTest/typeCheckedTests/expected_output/oo.re
jberdine: FWIW, I prefer the new version.  I still like the rule that anything that does not fit on a single line should have only whitespace to the left of its start.  If it was up to me, I would not dock anything at the end of a line except a brace, paren, or bracket.

formatTest/unit_tests/expected_output/modules.re
```diff
@@ -445,9 +447,8 @@ let module Id (X: Type) => X;
 let module Compose
            (F: Type => Type)
            (G: Type => Type)
-           (X: Type) => F (
-  G X
-);
+           (X: Type) =>
+  F (G X);
```
jberdine: If function names and their args are in one tuple, I think that functor names and their arguments should be too.

src/reason_pprint_ast.ml
yunxing: The problem here is in the original code, we are making a label of 
```
Label
  left:
    Label 
      left: "let x = " 
      right: "myFunction"
  right:
    "{"
    ...
```
Once we add a comment like this: 
```
let x = 
  /* blahblahblahblahblahblahblahblahblahblah */
  myFunction {
    abc
};
```
The format tree becomes:

```
Label
  left:
    Label 
      left: "let x = " 
      right: 
        Sequence
           "/* blahblahblahblahblahblahblahblahblahblah */"
           "myFunction"
  right:
    "{"
    ...
```

In this case pretty print just formats the tree "break" everything, even when they can fit into one line:
```
let x =>
  /* exclude the scope from the hash since it is ignored by equal */
  myFunction {
    abc
  };
```

formatTest/typeCheckedTests/expected_output/oo.re
jordwalke: In this case, isn't this a brace that is docked to the end of the line?

formatTest/unit_tests/expected_output/modules.re
jordwalke: I agree - they should be consistent. Right now, they are consistent, correct? Were you suggesting that whatever we do with function args, we do with functor args?

formatTest/typeCheckedTests/expected_output/oo.re
jberdine: True, this docks a brace at the end of a line, but it also docks `acceptsOpenAnonObjAsArg`, which is what I would rather it didn't do.  Instead I'd prefer a line break before `acceptsOpenAnonObjAsArg`, and then dock the brace at the end of that line:
```
let res =
  acceptsOpenAnonObjAsArg {
```

formatTest/unit_tests/expected_output/modules.re
jberdine: I was suggesting to treat functor args the same as function args, but as far as I understand this code isn't formatted the same as it would be as a function instead of a functor.  In particular, wouldn't function args be indented only 2 spaces, rather than being aligned under the name?

Issue #817, opened 2016-10-19
function optional arg type syntax error


chenglou:
```
external foo : label::(<foo : int>)? => unit = "bla";
```

prints to

```
external foo : label::<foo : int>? => unit = "bla";
```

Which then syntax errors.

chenglou:
Object syntax changed. No longer encounter this case.

PR #820, opened 2016-10-20
Fix issue #817.

bsansouci:
I'm pretty certain that objects are the only thing you can inline in the type signature (the rest is identifiers :( ). If someone can think of an edge case please tell me.

### DIFF
--- a/formatTest/unit_tests/input/syntax.re
+++ b/formatTest/unit_tests/input/syntax.re
@@ -864,6 +864,10 @@ let A | B | C = X;
  */
 external f : int => int = "foo";
 
+external foo : label::(<foo : int>)? => unit = "bla";
+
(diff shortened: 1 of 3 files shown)


formatTest/unit_tests/input/syntax.re
```diff
@@ -864,6 +864,10 @@ let A | B | C = X;
  */
 external f : int => int = "foo";
 
+external foo : label::(<foo : int>)? => unit = "bla";
```
jordwalke: I think the correct fix is to actually allow `<..>` as a "simple type" which should allow it to be parsed after the `::`. Did you try that?

chenglou:
Is this still relevant now that we've got the new obj syntax?

chenglou:
Guess not ¯\_(ツ)_/¯

PR #822, opened 2016-10-22
Expose reason.cma

yunxing:
so that people can just add "-package reason" in their ocamlfind


jordwalke:
Nice feel free to merge when you're ready.

Issue #827, opened 2016-10-26
[atom-reason] New highlight color for _name confusing


chenglou:
```
let _foo = 5;
```

This highlights as if it was a comment:

![unspecified](https://cloud.githubusercontent.com/assets/1909539/19714348/41724dc4-9b02-11e6-86e7-fb86294bd3d0.png)

@freebroccolo

freebroccolo:
@chenglou So highlighting names prefixed with underscores this way was an intentional design decision but maybe I should change it if people feel it is confusing.

The reason I did it that way is because when I'm writing a piece of code where the variable will be unused, I prefix it with an underscore to silence merlin warning about unused variables. At this point, I view it as being "ignored", hence the comment style.

I also think the visual distinction is helpful when scanning a block of code where there's a function defined with some unused arguments:

![unused](https://cloud.githubusercontent.com/assets/151197/19736997/e71946bc-9b6e-11e6-9973-59a3d3a85a08.png)

So one option could be to leave `_` prefixed names the way it is for function arguments but change it for other `let` bindings.

Another approach would be to use a more specific scope name so that the current comment-style highlighting could be applied as a user style customization. This is something I've been planning to do anyway with most of the other rules but it will take more work.

freebroccolo:
I should also point out that the `_` highlighting was intended to be useful for patterns in general though, not just for function arguments. I think that's why I originally did it that way:

![patterns](https://cloud.githubusercontent.com/assets/151197/19737681/9def9c4a-9b71-11e6-96da-6d29de294eba.png)

hcarty:
FWIW I like the "nothing to see here, move along" look of the `_`-prefixed values.  It makes the intent visually explicit.

PR #828, opened 2016-10-26
[atom-reason] fix several highlighting issues

freebroccolo:
Note this doesn't address #827 yet which I will respond to later.

* handle more attributes/extension nodes
* fix value highlighting in attributes
* fix conditional operators in parens
* move record paths to atomic expressions
* allow prefix operators for function names
* end record paths on operators with /
* highlight raise
* fix commas in record fields
* fix tuple patterns in let bindings
* improve highlighting for array indexing
* allow `...Module.field` for record fields
* don't terminate case arm on `||`

### DIFF
--- a/editorSupport/language-reason/grammars/reason.json
+++ b/editorSupport/language-reason/grammars/reason.json
@@ -23,26 +23,7 @@
             { "include": "#attribute-identifier" }
           ]
         },
-        {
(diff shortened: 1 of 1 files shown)


chenglou:
If only we could generate these from our existing tokenizer...

Issue #824, opened 2016-10-23
Special case keywords to be allowed as names for arguments


bsansouci:
It seems like I can't use `type` as a name for an argument because it's not an identifier. I'd like to allow those by adding special cases for the keywords we deem useful as named args.
`to` is another one (`let doThing from::0 to::10` dies of course).

let-def:
Ideally, I would not be against this idea, however it would break compatibility with OCaml (because in OCaml you really cannot use reserved words), and the story with punning will be messy (punning would be disabled in these cases, which is quite ad-hoc).

PR #833, opened 2016-10-31
Give a better error message when using a reserved keyword.

SanderSpies:


SanderSpies:
Ensures that: 
```
let x val => ();
```
will result in a proper error message.

bsansouci:
Does this solve #824 ?

SanderSpies:
@bsansouci No, this is just a prettier error message.

PR #841, opened 2016-11-12
[atom-reason] highlight `as`; jsx improvements

freebroccolo:
This PR fixes a bug with `as` in type expressions and improves highlighting for JSX. @jordwalke @chenglou 

Based on the discussion on discord, I investigated some alternatives for highlighting the JSX. There is a [react plugin](https://orktes.github.io/atom-react/) that shows highlighting using the "atom dark" theme where the tag delimiters and the tag name have the same colors. However, this is because the colors for those tokens for that theme happen to coincide. For other themes it can look differently.

I put together a comparison of some of the more common themes to show how JSX is highlighted with that plugin:

comparison

So what I've done is gone through and changed the highlighting for this plugin to use the usual theme tag highlighting scopes for lowercase names as well as the tag delimiters. For module paths, the module prefix is still highlighted as a module but if the last path component is a lowercase identifier, it is highlighted with the scope for tag names.

I think this is a okay trade off for a few reasons. For one, with the default "One Dark theme", if the module path prefixes use the theme tag scope, the paths will consist of light red next to dark red, which is something I've tried to minimize with the most recent changes @chenglou requested. I also think it's generally a good idea to try and maintain the invariant that syntactic entities have (for the most part) the same color consistently throughout the document.

But also for practical purposes, for a few themes it ends up looking worse than others to use the same scope for the tag name and the delimiters. For the "Dracula" theme (the one I personally use the most), for instance, it would either be the keyword pink (for lowercase) or the italic, underlined green (for module names), both of which look kind of bad for large blocks.

Currently I'm trying to get pretty good results across Dracula, Monokai, Dark+ [on vscode], One Dark, and Oceanic Next. So it's also a bit tricky to make adjustments to help one theme (e.g., Dracula, as above) without negatively impact the others too much.

~~But one other change I made is to lighten the background for JSX in the One Dark and a few others that have a similar scope. It's a bit subtle but maybe it helps a little bit.~~

We could try to improve things further still and maybe look at using some custom styles in the plugin itself but in the meantime this should be a little better.

Here are some previews:

#### Atom Dark
atom-dark
#### One Dark
dark-one
#### Oceanic Next
oceanic
#### Base16 Tomorrow Night
tomorrow

### DIFF
--- a/editorSupport/language-reason/grammars/reason.json
+++ b/editorSupport/language-reason/grammars/reason.json
@@ -162,25 +162,30 @@
         {
           "begin": "\\b([[:lower:]][[:word:]]*)\\b[[:space:]]*(=)",
           "end": "(?<![=])(?=[/>[:lower:]])",
+          "comment": "meta.separator",
           "beginCaptures": {
-            "1": { "name": "constant.language support.property-value entity.name.filename variable.interpolation" },
-            "2": { "name": "markup.inserted keyword.control.less message.error" }
+            "1": { "name": "constant.language support.property-value entity.name.filename markup.inserted" },
+            "2": { "name": "variable.other.class.js variable.interpolation keyword.operator keyword.control message.error" }
           },
           "patterns": [
             { "include": "#value-expression-atomic-with-paths" }
           ]
         },
         {
-          "match": "\\b([[:lower:]][[:word:]]*)\\b",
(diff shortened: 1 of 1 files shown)


freebroccolo:
I am happy to remove the lightening of the background for JSX blocks on themes that support it if people find the effect to jarring.

jordwalke:
Awesome thank you!

Issue #11, opened 2015-12-04
Any objection to converting the pretty printer to Reason?


jordwalke:

jordwalke:
I thought this would be difficult to self-host because of the challenges in building Reason from source, when you don't have an available Reason. But then I realized we can use a simple trick. The OCaml compiler (supposedly) supports loading files that are simply serialized ASTs. (The same output that `-pp` accepts, and the same output that `refmt` spits out by default).

So we can simply check in the serialized AST for `reason_pprint_ast` and friends. Our build scripts for Reason can simply disregard the actual `.re` files and only ever build the serialized AST files. The result is that it successfully builds a version of Reason that is a binary, even if Reason was not originally installed.

yunxing:
@jordwalke Interesting idea. If we are able to do that, I imagine we can even write parsers in reason as well (need to add a -pp option in Menhir first).

chenglou:
Stale, will keep it in ocaml for now

PR #863, opened 2016-11-25
(#11) Converting pretty printer to Reason.

SanderSpies:
Thanks to @jordwalke for helping out once again.

This correctly builds now. 

I guess we now also need a way to generate the binary AST when modifying the pretty printer. Any thoughts on that?


src/reason_pprint_ast.re
```diff
@@ -0,0 +1,6694 @@
...
+    let rec append space::space=false txt =>
+      fun
+      | SourceMap loc sub [@implicit_arity] =>
+        SourceMap loc (append space::space txt sub) [@implicit_arity]
+      | Sequence config l [@implicit_arity] when snd config.wrap != "" => {
+          let sep = if space {" "} else {""};
+          Sequence {...config, wrap: (fst config.wrap, snd config.wrap ^ sep ^ txt)} l
+          [@implicit_arity]
+        }
+      | Sequence conf
```
jordwalke: I think when converting you should get rid of these. You do that by passing `-assume-explicit-arity`, and then fixing any compiler errors by manually fixing constructors in the `.re` file.

src/reason_pprint_ast.re
```diff
@@ -0,0 +1,6694 @@
+/*
```
jordwalke: If you put this as the comment block of code at the top:
/*
 * Syntax and fallback syntax.
 * vim: set ft=rust:
 * vim: set ft=reason:
 */

Then it will highlight correctly.

SanderSpies:
If we want we can revisit this later.

PR #866, opened 2016-11-28
add merlin pin commands before `make`

knowbody:
So I followed the "contribution guide" but got stuck on this error, while trying to `make` the forked version of Reason:

Running these two commands solved the problem:

```
opam pin add -y merlin 'https://github.com/the-lambda-church/merlin.git'
opam pin add -y merlin_extend 'https://github.com/let-def/merlin-extend.git'
```

Also the docs [here](http://facebook.github.io/reason/#community) mention the `make` step which is missing here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,8 +104,11 @@ opam init
 opam update
 opam switch 4.02.3
 eval $(opam config env)
+opam pin add -y merlin 'https://github.com/the-lambda-church/merlin.git'
+opam pin add -y merlin_extend 'https://github.com/let-def/merlin-extend.git'
 git clone git@github.com:facebook/reason.git
 cd reason
+make
 opam pin add -y reason .
 ```
 


chenglou:
Humm that's weird. @yunxing removed it in https://github.com/facebook/reason/commit/d1db42b00e46f1420a7482317171c78d0121a25c. @jordwalke did you know what happened?

chenglou:
@knowbody ok, this is removed because we've already added the dependencies correctly in the opam meta file. Can you try `opam install` instead of pinning them manually? If that works, can you change the PR to that line instead? Thanks.

SanderSpies:
@chenglou this command does nothing for me 

(note I'm on OCaml 4.04 and need to get everything working again)

chenglou:
ccing @jordwalke again

jordwalke:
I think the right contributing instructions for `opam` are to clone and then do `opam install --deps-only`. But we should probably just include the `npm` instructions too.

jordwalke:
If the `--deps-only` flag works for you then we should add it to the docs, but I've updated the main docs with instructions on contributing via `npm` which is much easier:

https://github.com/facebook/reason/commit/e895a62c2978b188a46215f89546f39718f07894

SanderSpies:
I want to point out that I had to use the `opam pin` commands again to get everything working.

chenglou:
Stale. Should be good now

PR #871, opened 2016-12-02
Minor doc fix "argument clearly one unit" -> "argument is clearly one unit"

latentflip:
Spotted this when reading docs.


chenglou:
Thanks!

PR #880, opened 2016-12-07
Punning support

SanderSpies:
Unfortunately the wrappingTest is failing due to idempotent issues with comments.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -969,7 +969,7 @@ let four = increment 4;
 
 ```
 
-It's also possible to use punning for labelled curied arguments.
+It's also possible to use punning for labelled curried arguments.
 
 ```reason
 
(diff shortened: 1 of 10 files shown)


yunxing:
@SanderSpies patching your diff to see what's causing the comment issue.

yunxing:
image

Trying to print out the AST of a smiple test, looks like the location for the pattern variable "a" is missing (see highlighted area)?

SanderSpies:
Thanks for taking a look. I was looking in this direction, but forgot about this trick.

SanderSpies:
@yunxing Thanks again. It should work now 👍

yunxing:
Looks good! Let's merge this after CI check.

docs/index.html
```diff
@@ -428,7 +428,7 @@
 let thirteen = increment (double 6);
 ```
 
-####
```
jordwalke: Oops! These need to stay here btw.

Issue #879, opened 2016-12-07
JSX support for explicit passing of optional argument


chenglou:
https://realworldocaml.org/v1/en/html/variables-and-functions.html#explicit-passing-of-an-optional-argument

This is very, very useful for react. We'd be able to have `<Foo bar=?props.bar />` when `props.bar` is `option 'a`.

(For those who are confused by this, `<Foo bar=props.bar />` requires `props.bar` to be non-nullable if you defined foo's `bar` as an optional argument and actually use it)

cc @jordwalke @SanderSpies maybe even more important than punning ¯\_(ツ)_/¯

SanderSpies:
Close please.

chenglou:
Great stuff. Time to tweak the API.

PR #888, opened 2016-12-10
Explicit passing of optional JSX argument

SanderSpies:
Related to #879

### DIFF
--- a/formatTest/typeCheckedTests/input/jsx.re
+++ b/formatTest/typeCheckedTests/input/jsx.re
@@ -59,6 +59,15 @@ module Namespace = {
     };
 };
 
+module Optional = {
+    let createElement ::optional children => {
+        switch optional {
+            | Some a => {displayName: a}
+            | None => {displayName: "nope"}
+        };
+     };
+};
+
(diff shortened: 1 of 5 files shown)


formatTest/typeCheckedTests/input/jsx.re
```diff
@@ -59,6 +59,15 @@ module Namespace = {
     };
 };
 
+module Optional = {
+    let createElement ::optional children => {
```
jordwalke: I'm surprised you don't get a warning

> Warning 43: the label optional is not optional.

Since it's not marked as optional `::optional=?` at definition time.

You might want to add a couple of test cases of these forms

```
let createElement ::notActuallyOptional children =>
let createElement ::optional=? children =>
let createElement ::optional="DefaultValue" children =>
```

SanderSpies: From what I see it works correctly, as this is a required option value.

I did add the extra tests as there is a slight issue with partial application in the JSX tests.

jordwalke:
Great, this will make the React bindings much more pleasant.

PR #890, opened 2016-12-12
Temporarily remove BetterErrors

chenglou:
This cuts down install time. For the users of BetterErrors, I'll find a
better intergration very soon, instead of baking it into reason.


chenglou:
Both opam and npm local install succeeded. Will wait for CI.

PR #891, opened 2016-12-12
JSX documentation

SanderSpies:
Feedback very welcome!

I'm not sure if we should add fragments already or the link to the bindings.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -1110,6 +1110,64 @@ let obj: tesla {. drive: int => int, x: int => int} = {
 };
 ```
 
+JSX
+----------------------------------
+Reason has support for JSX syntax with some slight differences compared to the one found
+in JavaScript.  JSX tags translate to function calls as shown in these examples:
+
+Capitalized tag:
+
+```reason
+<MyComponent foo=bar />
+```
+
+becomes
+
+```reason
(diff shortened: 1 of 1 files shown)


bsansouci:
Hey do we still want to document JSX that way? @chenglou. 

If Yeah, you should update the PR to use the new format https://github.com/facebook/reason/pull/924

kennetpostigo:
JSX is really great. I really want to take advantage of it since its in Reason. It would be nice if the documentation could show how to bind jsx to calls other than react, because a friend of mine and I really want to use it with Vulkan.

chenglou:
@kennetpostigo the secret repo has the current transform & docs on how it works. For ppx in general we've explicitly refrained from documenting it too much because ppx is such a double-edged sword.

This PR is stale but as soon as we open up the secret repo we'll document it better. In the meantime, there's really nothing special about how this ppx is done. [This](https://github.com/facebook/reason/blob/3eea36bfcc4744a1974f269db84616e3a31fcf5a/src/reactjs_jsx_ppx.ml) is the single file and here's a guide: https://whitequark.org/blog/2014/04/16/a-guide-to-extension-points-in-ocaml/

chenglou:
Bah let's just get this in first =)

PR #895, opened 2016-12-13
Remove dependency on Re

chenglou:
We only use it in a few places where normal string utils would do. This PR implements these string helpers.

### DIFF
--- a/src/syntax_util.ml
+++ b/src/syntax_util.ml
@@ -49,6 +49,37 @@ let replace_string old_str new_str str =
   replace_string_ old_str new_str 0 str buffer;
   Buffer.contents buffer
 
+let split str ~by =
+  let str = ref str in
+  let accum = ref [] in
+  try
+    while true do
+      let strVal = !str in
+      let foundIdx = String.index_from strVal 0 by in
+      let subStr = String.sub strVal 0 foundIdx in
+      str := String.sub strVal (foundIdx + 1) (String.length strVal - foundIdx - 1);
+      accum := (subStr :: !accum);
+    done;
+    assert false
+  with Not_found ->
+    List.rev (!str :: !accum)
+
+let rec trim_right_idx str idx =
+  if idx = -1 then 0
+  else
+    let ch = String.get str idx in
+    if ch = '\t' || ch = ' ' || ch == '\n' || ch == '\r' then trim_right_idx str (idx - 1)
+    else idx + 1
+
+let trim_right str =
+  let length = String.length str in
+  if length = 0 then "" else String.sub str 0 (trim_right_idx str (length - 1))
+
+let strip_trailing_whitespace str =
+  split str ~by:'\n'
+  |> List.map trim_right
+  |> String.concat "\n"
+  |> String.trim
 
 module StringMap = Map.Make (String)
 
--- a/src/syntax_util.ml
+++ b/src/syntax_util.ml
@@ -49,34 +49,49 @@ let replace_string old_str new_str str =
   replace_string_ old_str new_str 0 str buffer;
   Buffer.contents buffer
 
-let split str ~by =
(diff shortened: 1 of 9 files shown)


chenglou:
The new strip trailing spaces logic uses a dumb string split that makes a 6k lines file refmt go from 600ms to 1000ms. Not sure if we should go forward with this. Slowdown negligible for files with hundreds of lines.

src/syntax_util.ml
```diff
@@ -49,6 +49,37 @@ let replace_string old_str new_str str =
...
 
+let split str ~by =
+  let str = ref str in
+  let accum = ref [] in
+  try
+    while true do
+      let strVal = !str in
+      let foundIdx = String.index_from strVal 0 by in
+      let subStr = String.sub strVal 0 foundIdx in
+      str := String.sub strVal (foundIdx + 1) (String.length strVal - foundIdx - 1);
```
hhugo: There is not need to allocate this intermediate string. It makes the function quadratic.

src/syntax_util.ml
```diff
@@ -49,6 +49,37 @@ let replace_string old_str new_str str =
...
+let rec trim_right_idx str idx =
+  if idx = -1 then 0
+  else
+    let ch = String.get str idx in
+    if ch = '\t' || ch = ' ' || ch == '\n' || ch == '\r' then trim_right_idx str (idx - 1)
+    else idx + 1
+
+let trim_right str =
+  let length = String.length str in
+  if length = 0 then "" else String.sub str 0 (trim_right_idx str (length - 1))
```
hhugo: no need to call `String.sub` when trim_right_idx returns `length` or `0`.

Also it would be less confusing to make`trim_right_idx` return an index instead of the length

src/syntax_util.ml
```diff
@@ -49,6 +49,37 @@ let replace_string old_str new_str str =
...
+
+let strip_trailing_whitespace str =
+  split str ~by:'\n'
+  |> List.map trim_right
+  |> String.concat "\n"
+  |> String.trim
```
hhugo: This does more than just stripping trailing whitespace. It also strip leading and trailing newlines

src/syntax_util.ml
```diff
@@ -49,6 +49,35 @@ let replace_string old_str new_str str =
   replace_string_ old_str new_str 0 str buffer;
   Buffer.contents buffer
 
+let rec split_ str ~by accum =
+  try
+    let foundIdx = String.index_from str 0 by in
+    let subStr = String.sub str 0 foundIdx in
+    split_
```
jordwalke: Thank you @samoht for the catch!

src/syntax_util.ml
chenglou: I do return the index. Also, this turned out to be the optimization that brings the perf back completely. Thanks!

src/syntax_util.ml
chenglou: Yeah this is intended. We remove all leading and trailing stuff, then add a single trailing line break back.

chenglou:
Thanks so much for the review. Turned out it was the unneeded string allocation from `sub` that made the perf worse. It's now better than `Re` in this particular use-case (ran it on reason_pprint_ast.ml itself, and a small ml file).

bobzhang:
@chenglou yes, of course. In general regex is slower than string manipulation, great to see reason has less dependencies

yunxing:
Could you also update package.json.in and opam.in ? those are two templates files that we use to generate opam and package.json.

yunxing:
Looks good to me now!

chenglou:
Works and builds. Merging!

Thanks again @hhugo @bobzhang @samoht =D. Reason now has one less dependency without perf tradeoffs.

Issue #874, opened 2016-12-04
JSX punning doesn't report correct location


chenglou:
Repro:

```
<Foo bar/>
```

If bar's type is wrong, the location isn't attached to it. Merlin then sets it to the default 0.

chenglou:
Fixed by ^

PR #897, opened 2016-12-13
JSX location fixes

SanderSpies:
To fix:

"chenglou: Sander Spies.re trying out the new foo=?bla jsx syntax. When a type error is thrown there the location is 0"

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/jsx.re
+++ b/formatTest/typeCheckedTests/expected_output/jsx.re
@@ -95,8 +95,8 @@ module Namespace = {
 };
 
 module Optional1 = {
-  let createElement ::optional children =>
-    switch optional {
+  let createElement ::requiredOpt children =>
+    switch requiredOpt {
     | Some a => {displayName: a}
     | None => {displayName: "nope"}
     };
@@ -506,10 +506,10 @@ let myFun () =>
 /*   <Bar /> */
 /* </Foo>; */
 /* let res = <Foo a=10 b=(<Foo a=200 />) />; */
-let zzz = "oh hai";
+let zzz = Some "oh hai";
 
 let optionalCallSite =
-  <Optional1 optional=?zzz />;
+  <Optional1 requiredOpt=?zzz />;
 
 fakeRender optionalCallSite;
 
--- a/formatTest/typeCheckedTests/expected_output/jsx.re
+++ b/formatTest/typeCheckedTests/expected_output/jsx.re
@@ -95,8 +95,8 @@ module Namespace = {
 };
 
 module Optional1 = {
-  let createElement ::requiredOpt children =>
-    switch requiredOpt {
+  let createElement ::required children =>
+    switch required {
     | Some a => {displayName: a}
     | None => {displayName: "nope"}
     };
@@ -112,9 +112,9 @@ module Optional2 = {
 
 module DefaultArg = {
   let createElement
-      ::optional=(Some "foo")
+      ::default=(Some "foo")
       children =>
-    switch optional {
+    switch default {
     | Some a => {displayName: a}
     | None => {displayName: "nope"}
     };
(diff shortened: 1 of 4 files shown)


SanderSpies:
Should also fix #874.

jordwalke:
How does this fix 898?

SanderSpies:
@jordwalke I had the impression the location info would help here. At least on my system I couldn't reproduce the issue anymore. Perhaps I missed something.

jordwalke:
That issue was about a totally different parse tree being constructed. It looks unrelated to me.

SanderSpies:
@jordwalke You are right, my bad. I made some more more fixes for #898. Could you check again?

formatTest/typeCheckedTests/expected_output/jsx.re
```diff
@@ -95,8 +95,8 @@ module Namespace = {
 };
 
 module Optional1 = {
-  let createElement ::optional children =>
-    switch optional {
+  let createElement ::requiredOpt children =>
```
jordwalke: Is my understanding correct that this is not *technically* an "optional" labeled argument right? It's just a regular labeled argument that happens to be of type `option string`.

formatTest/typeCheckedTests/expected_output/jsx.re
```diff
@@ -506,10 +506,10 @@ let myFun () =>
 /*   <Bar /> */
 /* </Foo>; */
 /* let res = <Foo a=10 b=(<Foo a=200 />) />; */
-let zzz = "oh hai";
+let zzz = Some "oh hai";
 
 let optionalCallSite =
-  <Optional1 optional=?zzz />;
+  <Optional1 requiredOpt=?zzz />;
```
jordwalke: Curious how this type checked before.

formatTest/typeCheckedTests/input/jsx.re
```diff
@@ -332,8 +332,8 @@ let myFun () => {
 /* </Foo>; */
 
 /* let res = <Foo a=10 b=(<Foo a=200 />) />; */
-let zzz = "oh hai";
-let optionalCallSite = <Optional1 optional=?zzz />;
+let zzz = Some "oh hai";
+let optionalCallSite = <Optional1 requiredOpt=?zzz />;
 fakeRender optionalCallSite;
 let optionalArgument = <Optional2 />;
```
jordwalke: Since `Optional2` is an *actual* "optional labeled" argument, I feel we should also add a test where we explicitly pass `=?`.

jordwalke: Also for `DefaultArg` below as well. To make sure that it's parsed correctly - we should force passing an `option` datatype via `<DefaultArg optional=?someIdentifier />`

jordwalke: Is there support for punned explicitly passed optionals yet? (No worries if we haven't done it yet, it can come later).

formatTest/typeCheckedTests/expected_output/jsx.re
SanderSpies: I was doing the wrong thing by wrapping the value in Some for the user.

formatTest/typeCheckedTests/expected_output/jsx.re
SanderSpies: Yeah, that was the idea behind the `requiredOpt` naming. I'll rename it to `required` to avoid any confusion.

jordwalke:
Play looks good. Now the only question is which syntax you like more attr=?foo or attr?=foo. I could see arguments in favor of either.

SanderSpies:
I prefer the former as it seems more in sync with `::attr=?`.

chenglou:
sooooo we good?

SanderSpies:
Yeah, this should be good. Punning is something we could fix in another Good First Task PR.

chenglou:
Ooook merging this then! Will ping you if it breaks >=)
Thanks!

PR #902, opened 2016-12-14
Add filetype directive

mjackson:


jordwalke:
Thanks! Will merge once CI passes

mjackson:
FWIW syntax highlighting works correctly when I put the VimReason snippet *before* the Vundle stuff in my `.vimrc`.

chenglou:
Seems like Jordan added something that made this diverge

mjackson:
Ya, we don't need to have this. It was an issue with the way Vim loads plugins. vim-reason-loader should address this.

PR #903, opened 2016-12-14
Github Highlighting for tests and table formatting 

kennetpostigo:
I added to all test files so that syntax highlighting on github works:
```
/*
 * vim: set ft=rust:
 * vim: set ft=reason:
 */
```

Also fixed formatting of tables in the OCaml Reason comparison html. Some tables weren't appearing on the opposite side of the description text.

### DIFF
--- a/docs/mlCompared.html
+++ b/docs/mlCompared.html
@@ -156,27 +156,27 @@ optional. `{}` braces are only needed if you have more than one item to chain
 together via `;`.
 
 > <table>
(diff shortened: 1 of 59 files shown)


vramana:
Considering this https://github.com/github/linguist/pull/3336 will be merged soon. Are highlighting coments necessary?

kennetpostigo:
@vramana I could always remove it later. Was going through test for docs and it was really bothering me.

PR #905, opened 2016-12-15
New JSX PPX logic

chenglou:
This changes `Foo.createElement bar::1 [][@JSX]` from the previous logic (a no-op when the JSX element is upper-cased) to `Foo.createElement bar::1 [] ()`. This way we don't have to forward `ref` and `key` for each component's boilerplate `createElement` call. For more info, see the super duper secret library.


chenglou:
Merging because I got swag

PR #909, opened 2016-12-16
Fix 861

let-def:
The escaping scheme is now simply that all occurrences of `/*` and `*/` should be replaced by `/\*` and `*\/`.

The patch also replaces the implementation of `replace_string`. It no longer allocates any memory on the fast path (no match). Previous version would allocate ~8 words per character.

### DIFF
--- a/src/syntax_util.ml
+++ b/src/syntax_util.ml
@@ -142,29 +142,25 @@ let identifier_mapper f =
 
 (** unescape_stars_slashes_mapper unescapes all stars and slashes in an AST *)
 let unescape_stars_slashes_mapper =
-  let unescape_stars_slashes str =
-    let len = String.length str in
-    if len < 2 then
-      str
-    else
-      let ending = String.sub str 1 (len - 1) in
-    String.sub str 0 1 ^
+  let unescape_stars_slashes = function
+    | str when String.length str <= 1 -> str
+    | "/\\*" -> "/*"
+    | "*\\/" -> "*/"
+    | str ->
       replace_string "\\*" "*"
-        (replace_string ("\\/") "/" ending)
+        (replace_string "\\/" "/" str)
   in
   identifier_mapper unescape_stars_slashes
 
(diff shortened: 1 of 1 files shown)


yunxing:
Looks good, thank you!

jordwalke:
We should make sure to bump the version of Reason before publishing this (because it breaks existing code that uses the unescaped versions IIRC - a version bump conveys that breaking change).

For example, this line in Rebel breaks https://github.com/reasonml/rebel/blob/master/src/rebel.re#L449

Rebel shouldn't be depending on master reason, it should have a specific version locked down - but in this case unless we bump the version of Reason, even locking down a version won't be good enough to prevent breakages of Rebel (or other packages).

jordwalke:
It's somewhat of a shame that something like ppx can break something. Imagine if we committed the raw AST when publishing packages. We wouldn't have that problem at all. I just don't know if the locations for debugging will work if not parsed on the host that we compile on.

let-def:
I don't understand your point about ppx breakage?

jordwalke:
Sorry, probably because I already fixed the issue in master. It previously had `^\/` which now must be `^/` (since you improved the lexing).

jordwalke:
But basically, my point is that since the AST itself hasn't changed and all source files can be parsed before they are even sent to the package user, it's a shame that parse time changes can break things.

let-def:
In one case you are sensitive to concrete syntax changes, in the other to abstract syntax (AST) changes.
Not sure which is better, especially if you want your package to work with multiple major versions
(as this is related to versioning problem with PPX which is bothering many people, there is hope to find a solution satisfying everyone).

jordwalke:
They are two problems, and it seems the concrete syntax problem is much easier to solve. In fact, before publishing you can just check in the binary ast with a `.ml`/`.mli` extension and the compiler will just automatically understand it. The cross-version issue is different but we have a pretty good proposal that is in alignment with what ocamllabs is thinking.

PR #912, opened 2016-12-19
Fix stale vim plugin path

fortes:
The hyperlink in the paragraph above points to the right repo, fixed here.


jordwalke:
Thank you!

PR #913, opened 2016-12-21
Revert "Fix 861 (#909)"

jordwalke:
This reverts commit 9fe606dc03b057197dce536c773aff0e45693acb.

cc @let-def


jordwalke:
There was some issues with how this was escaping things. Fred is on it.

PR #917, opened 2016-12-25
JSX PPX: make children labelled

chenglou:
We need this to partially apply children. We have a unit at the end to
catch all things anyway.

**DON'T MERGE**. Possibly Being superseded by #924


jordwalke:
Should this be the default for children? Instead of just in the JS ppx?

jordwalke:
I noticed you haven't merged this yet - are you depending on it anywhere yet?

chenglou:
I'd love the transform to stay as close to the non-ppexed version as much as possible. I thought you needed the non-transformed version yourself? If not I can change the original one instead of doing it in the ppx.

Not using it yet. Soon.

chenglou:
DON'T MERGE. Possibly Being superseded by #924

chenglou:
Killed by ^

PR #919, opened 2016-12-27
Improve auto-mode-alist configuration

syohex:
`autoload` cookie affects only one s-expression. So original code should be written as bellow.

```lisp
;;;###autoload
(progn
  (add-to-list 'auto-mode-alist '("\\.re\\'" . reason-mode))
  (add-to-list 'auto-mode-alist '("\\.rei\\'" . reason-mode)))
```

However this is enough to improve regular expression of file extension.


yunxing:
Didn't know that. Thanks for fixing!

Issue #918, opened 2016-12-26
Extensible variants reformat to Ocaml


IwanKaramazow:
```
module Attr = {
    type t = int;
};
type tag 'props = ..;
type ttype = Type;
type description 'props = {
  ttype: ttype,
  props: 'props
};
type element = Native (tag 'props) (description 'props): element;

type attrProps = { attr: Attr.t };

type tag 'props += Attr: tag attrProps;

```

```
module Attr = {
  type t = int;
};

type tag 'props = ..;

type ttype =
  | Type;

type description 'props = {ttype: ttype, props: 'props};

type attrProps = {attr: Attr.t};

type 'props tag +=
  | Attr: attrProps tag;
```
Notice how the extension of the variant gets reformatted to something halfway Ocaml, halfway Reason.

jordwalke:
Whoever fixes this, please mark yourself as the assignee. Please add test cases as well.

[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=40407524)](https://www.bountysource.com/issues/40407524-extensible-variants-reformat-to-ocaml?utm_source=40407524&utm_medium=shield&utm_campaign=ISSUE_BADGE)

jordwalke:
@IwanKaramazow are you on master? If not, which version.

IwanKaramazow:
I'm on master. My package.json includes `"reason": "facebook/reason"`.
Will fix this myself, is blocking me too much on RED

jordwalke:
Okay thank you. I'll assign it to you.

chenglou:
This is done. @jordwalke cough up the money! lol

jordwalke:
Let me know if bounty source doesn't do its thing correctly.

PR #921, opened 2016-12-28
Fix printing of extensible variants

IwanKaramazow:
The printer defaulted to the default (Ocaml) printer when pretty printing extensible variants.
See https://github.com/facebook/reason/issues/918

This PR tries to fix the problem & intends to make Extensible Variants great again.

### DIFF
--- a/src/reason_pprint_ast.ml
+++ b/src/reason_pprint_ast.ml
@@ -2446,6 +2446,62 @@ class printer  ()= object(self:'self)
     in
     (SourceMap (ptype_loc, everything))
 
+  method formatOneTypeExt prepend name assignToken te =
+    let equalInitiatedSegments =
+      let segments = List.map self#type_extension_binding_segments te.ptyext_constructors in
+      [makeList ~break:Always_rec ~postSpace:true ~inline:(true, true) segments] in
+    let formattedTypeParams = List.map self#type_param te.ptyext_params in
+    let binding = makeList ~postSpace:true (prepend::name::[]) in
+    let labelWithParams = match formattedTypeParams with
+        [] -> binding
+      | phd::ptl -> label ~space:true binding (makeList ~postSpace:true ~break:IfNeed ~inline:(true, true) (phd::ptl)) in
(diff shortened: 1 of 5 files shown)


src/reason_pprint_ast.ml
```diff
@@ -2446,6 +2446,62 @@ class printer  ()= object(self:'self)
...
+      | Pext_rebind rebind ->
+        let r = self#longident_loc rebind in
+        let prepend = (atom "=") in
+        ([makeList ~postSpace:true [prepend; r]], None)
+    in
+    let everything = match resolved with
+      | ([], None) -> barName
+      | ([], Some res) -> add_bar sourceMappedName res
+      | (_::_ as args, None) -> add_bar nameOf (normalize args)
+      | (_::_ as args, Some res) -> add_bar nameOf (normalize (args@[res]))
```
yunxing: I'm a bit confused here. What's an example of this case?

IwanKaramazow: The first element of the tuple represents constructor arguments, the second an optional formatted gadt.

```ocaml
(* Case 1: No constructor arguments, neither a gadt *)
type attr = ..;
type attr += | Str

(* Case 2: No constructor arguments, is a gadt  *)
type attr = ..;
type attr += | Str :attr

(* Case 3: Has Constructor args, not a gadt  *)
type attr  = ..;
type attr += | Str string;
type attr += | Point int int;

(* Case 4: Has Constructor args & is a gadt *)
type attr  = ..;
type attr += | Point int int :attr;
```

jordwalke: That would make an excellent comment!

SanderSpies:
Within the [404](https://github.com/facebook/reason/pull/907/files#diff-7dce44be99776d19af8c2d7b0d14e2c7R5843) branch the method `type_extension` is also implemented. I'm not sure if it solves this problem, but might be worth a look.

yunxing:
Looks good to me, merging.

Issue #889, opened 2016-12-12
Support trailing commas in tuples.


jordwalke:
We can easily parse these to avoid accidental errors. When printing it will be very easy to omit them. Ideally, we would even print the trailing comma, but *only* when breaking on multiple lines, and otherwise not (but that advanced printing requires deeper hooks into our formatting library).

jordwalke:
This is a great first task. You could look at how we do this for trailing commas in records.

IwanKaramazow:
Merged, can be closed.

yunxing:
@IwanKaramazow thanks!

PR #923, opened 2016-12-29
Add support for trailing commas in tuples

IwanKaramazow:
Implements https://github.com/facebook/reason/issues/889

This does not print trailing commas when line breaks are needed.
I think this requires an enhancement in the EasyFormat lib.

### DIFF
--- a/formatTest/unit_tests/expected_output/basicStructures.re
+++ b/formatTest/unit_tests/expected_output/basicStructures.re
@@ -345,6 +345,11 @@ let (tupleItem, withOutsideTypeConstraint): (
   20
 );
 
+/* Trailing commas */
+let trailingCommaAccepted = (1, 2);
+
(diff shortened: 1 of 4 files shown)


SanderSpies:
LGTM

yunxing:
Yes, we need more options in EasyFormat in order to do this. cc @pieterv who has the same requirement.

yunxing:
Thanks!

pieterv:
@yunxing yes i believe this requires a change to EasyFormat (see https://github.com/mjambon/easy-format/issues/11). I did manage to implement a hack (as suggested by @jordwalke) where we insert a token in places we potental want trailing commas and then after printing from EasyFormat to string we find the instances of the token followed by a "\n" and replace with ",\n" otherwise strip it. Its working well so far although can mess with the indentation a little since you remove chars after doing all the breaking.

PR #924, opened 2016-12-29
JSX: Foo.createElement ::bar [] to Foo.createElement ::bar children::[] ()

chenglou:
This changes the JSX to desugar to children being just another labeled argument (motivation hidden for now =P). Since we can't use it to finish applying a function, we add a unit at the end.

This is #917 but at the syntax level. We don't really need to have it at the ppx level.

### DIFF
--- a/miscTests/jsxPpxTest.sh
+++ b/miscTests/jsxPpxTest.sh
@@ -20,8 +20,7 @@ do
   else
     echo "Wrong"
     # show the error
-    # diff -u $testPath/expected$i.txt $testPath/actual$i.txt
-    icdiff $testPath/expected$i.txt $testPath/actual$i.txt
+    diff -u $testPath/expected$i.txt $testPath/actual$i.txt
     exit 1
   fi
 done
--- a/miscTests/jsxPpxTest.sh
+++ b/miscTests/jsxPpxTest.sh
@@ -2,7 +2,7 @@ echo "Testing reactjs @JSX ppx..."
 
 testPath="miscTests/reactjs_jsx_ppx_tests"
 
-for i in 1 2 3 4 5 6
+for i in 1 2 3 4 5
 do
   test="$testPath/test$i.re"
 
@@ -20,7 +20,8 @@ do
   else
     echo "Wrong"
     # show the error
-    diff -u $testPath/expected$i.txt $testPath/actual$i.txt
+    # diff -u $testPath/expected$i.txt $testPath/actual$i.txt
+    icdiff $testPath/expected$i.txt $testPath/actual$i.txt
     exit 1
   fi
 done
--- a/src/reason_parser.mly
+++ b/src/reason_parser.mly
@@ -346,6 +346,9 @@ let mkuplus name arg =
 let mkexp_cons consloc args loc =
   mkexp ~loc (Pexp_construct(mkloc (Lident "::") consloc, Some args))
 
+let mkexp_constructor_unit consloc loc =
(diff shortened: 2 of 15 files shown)


src/reason_parser.mly
```diff
@@ -2648,7 +2651,10 @@ jsx:
   | jsx_start_tag_and_args SLASHGREATER {
     let (component, _) = $1 in
     let loc = mklocation $symbolstartpos $endpos in
-    component [("", mktailexp_extension loc [] None)] loc
+    component [
+      ("children", mktailexp_extension loc [] None);
+      ("", mkexp_constructor_unit loc loc)
```
chenglou: I'm likely not using whatever location stuff correctly here.

src/reason_pprint_ast.ml
```diff
@@ -857,17 +857,18 @@ let default = new Pprintast.printer ()
 type funcReturnStyle =
   | ReturnValOnSameLine
 
-let rec detectJSXComponent e attributes l =
-  match (e, attributes) with
+let rec detectJSXComponent e attributes l = None
```
chenglou: What's happening here? Simply having this change (without the rest) doesn't change break any unit test?

src/reason_parser.mly
SanderSpies: LGTM

src/reason_pprint_ast.ml
jordwalke: There's a note where this `detectJSXComponent` is called:

```
        (* This case will happen when there is something like

             Bar.createElement a::1 b::2 [] [@bla] [@JSX]

           At this point the bla will be stripped (because it's a visible
           attribute) but the JSX will still be there.
         *)
```

So it could just be that our unit tests aren't testing it?

chenglou: It's not stripped, just checked

miscTests/jsxPpxTest.sh
```diff
@@ -20,7 +20,8 @@ do
   else
     echo "Wrong"
     # show the error
-    diff -u $testPath/expected$i.txt $testPath/actual$i.txt
+    # diff -u $testPath/expected$i.txt $testPath/actual$i.txt
+    icdiff $testPath/expected$i.txt $testPath/actual$i.txt
```
yunxing: what's icdiff? ****

chenglou: ops, forgot to change it back, thanks. It's the diff I wish was on every platform. Check it out!

src/reason_pprint_ast.ml
```diff
@@ -3578,10 +3548,12 @@ class printer  ()= object(self:'self)
   method formatJSXComponent componentName args =
     let rec processArguments arguments processedAttrs children =
       match arguments with
-      | ("", {pexp_desc = Pexp_construct (_, None)}) :: tail ->
+      | ("children", {pexp_desc = Pexp_construct (_, None)}) :: tail ->
```
yunxing: I haven't been following jsx for a while so my understanding maybe lagging, but is this logic going to make the parser locked to reactjs ("children")? Should we somehow move it to reactjs_jsx_ppx?

chenglou: It's not really locked. Your suggestion is implemented in #917.
<Foo>bar</Foo> desugaring to `Foo.createElement children::[bar] () [@JSX]` rather than `Foo.createElement [bar][@JSX]` isn't really more prescriptive. I'd say on the contrary, it allows children to be partially applied too rather than making the assumption that it can never be.

PR #925, opened 2016-12-30
[docs] Fix old NeoBundle instruction.

zacanger:
From `jordwalke` to `reasonml`.


yunxing:
Looks good. Thanks man!

PR #928, opened 2017-01-01
JSX: type dom elements

chenglou:
See the new `expected1.re`.

**Don't merge yet**. Testing in codebase internally.


chenglou:
Tested, works.

Issue #930, opened 2017-01-03
refmt incorrect in array access on ref


Schmavery:
When reformatting access to an array ref, refmt messes up the precedence and reformats to uncompilable code.
The input:
```reason
let a = ref [|0|];
let foo = (!a).(0);
let foo = Array.get !a 0;
```
reformats to:
```reason
let a = ref [|0|];
let foo = !a.(0);
let foo = !a.(0);
```

Compiling this throws the error:
```
Error: The types don't match.
This is: int array ref
Wanted:  'a array
```

The precedence of ! is seems too low for this transformation to work?
Edit: this is also true of character access in a string `(!a).[0]`.  Maybe some problem in the logic [here](https://github.com/facebook/reason/blob/ec1560da5314986aa63667806bb152c90ad54b40/src/reason_pprint_ast.ml#L3203)?
Edit2: Been doing some experimenting and it seems like `simple_enough_to_be_lhs_dot_send` might be more appropriate than `simplifyUnparseExpr` here.

chenglou:
Closed by ^

PR #931, opened 2017-01-03
Fix reformat of ref access (#930)

Schmavery:
Before this, refmt converts
```reason
let foo = (!a).(0);
let foo = Array.get !a 0;
```
to:
```reason
let foo = !a.(0);
let foo = !a.(0);
```

This doesn't compile because of the precedence of `!`.
Now it should reformat to:
```reason
let foo = (!a).(0);
let foo = (!a).(0);
```

I also did my best to fix the problem where it occurred elsewhere (string access, Bigarray, setter sugar)


jordwalke:
Thanks @Schmavery !

PR #932, opened 2017-01-04
Remove re.rtop in toplevel

yunxing:
@yutongp @chenglou


yutongp:
remove `re.str`?

chenglou:
Yeah, it's not used anymore.

yunxing:
@yutongp Good point, changed the title.

PR #933, opened 2017-01-04
Documentation markdown update

jasonrose:
I've made a small update to the docs to make sure I can go through the whole workflow for contributing to this.


docs/index.html
```diff
@@ -442,7 +442,7 @@
 let thirteen = increment (double 6);
 ```
 
-####
```
SanderSpies: These should not be removed, so I was told before.

chenglou:
Thanks!

PR #935, opened 2017-01-04
Deprecating single hyphen refmt args

jasonrose:
This aligns `refmt` parameters with GNU's recommendations of two hyphens (`--`) for long-form parameters.  We also additionally support two short-form parameters, `-h` and `-i` for `--help` and `--interface`.

This also defaults `--interface` to false; the biggest change here is that piping stdin from the terminal now assumes that it is not an interface unless explicitly told so.  `refmt` would fail out in this situation before this PR.

I have left support in for the old flags but now log deprecation messages to stderr.  This might cause some regressions if people are piping stdout to a file and not handling stderr.

### DIFF
--- a/src/refmt_impl.ml
+++ b/src/refmt_impl.ml
@@ -54,13 +54,13 @@ let usage = {|Reason: Meta Language Utility
 [Usage]: refmt [options] some-file.[re|ml|rei|mli]
 
    // translate ocaml to reason on stdin
-   echo 'let _ = ()' | refmt -print re -parse ml -use-stdin true -is-interface-pp false
+   echo 'let _ = ()' | refmt --print re --parse ml --use-stdin true
 
    // print the AST of a file
-   refmt -parse re -print ast some-file.re
+   refmt --parse re --print ast some-file.re
 
    // reformat a file
-   refmt -parse re -print re some-file.re
+   refmt --parse re --print re some-file.re
 
 [Options]:
 |}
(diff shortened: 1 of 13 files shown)


src/refmt_impl.ml
```diff
@@ -81,16 +81,26 @@ let () =
   let print_width = ref None in
   let print_help = ref false in
   let options = [
-    "-is-interface-pp", Arg.Bool (fun x -> intf := Some x), "<interface>, parse AST as <interface> (either true or false)";
-    "-use-stdin", Arg.Bool (fun x -> use_stdin := x), "<use_stdin>, parse AST from <use_stdin> (either true, false). You still must provide a file name even if using stdin for errors to be reported";
-    "-recoverable", Arg.Bool (fun x -> recoverable := x), "Enable recoverable parser";
-    "-assume-explicit-arity", Arg.Unit (fun () -> assumeExplicitArity := true), "If a constructor's argument is a tuple, always interpret it as multiple arguments";
-    "-parse", Arg.String (fun x -> prse := Some x), "<parse>, parse AST as <parse> (either 'ml', 're', 'binary_reason(for interchange between Reason versions)', 'binary (from the ocaml compiler)')";
+    "-is-interface-pp", Arg.Bool (fun x -> prerr_endline "-is-interface-pp is deprecated; use -i or --interface instead"; intf := Some x), "";
```
jordwalke: Nice - supporting the old form too.

vjeux:
Thank you for doing that!

jordwalke:
Wonderful!

yunxing:
@jasonrose Thanks!

chenglou:
@jasonrose late, but thanks for the upgrade path. This ensured that nuclide isn't suddenly broken.

Issue #547, opened 2016-05-27
Better syntax error for variant tags that are not upper-cased


chenglou:
cc @yunxing (related: https://github.com/chenglou/BetterErrors/issues/25)

Can't fix this with regex

chenglou:
Closed by ^

PR #937, opened 2017-01-06
Provide a better error message when a variant constructor is accidentally lowercase

SanderSpies:
#547


src/reason_parser.messages
```diff
@@ -12068,7 +12068,7 @@ implementation: TYPE LIDENT EQUAL UNDERSCORE EQUAL BAR WITH
 ## BAR
 ##
 
-<SYNTAX ERROR>
+Variant constructors begin with an uppercase character
```
chenglou: Can you add a "need to" here? "Variant constructors need to begin with..."

Issue #856, opened 2016-11-23
foo#=<bar /> doesn't parse correctly


chenglou:
cc @bsansouci

Schmavery:
If I had to guess, I would say that it might be parsing `#=<` as being the name of the infix operator.

jordwalke:
I think this is acceptable for all the reasons Schmavery mentioned. As long as it parses with the space inserted and prints correctly in all cases we should be fine I think.

Issue #849, opened 2016-11-21
Better syntax for ## and #=


chenglou:
cc @bobzhang

Currently, for BuckleScript (and hopefully jsoo) we have `foo##bar` for dot access, and `foo##bar #= 1` for object field assignment.

The feedback we've gotten is that people are scared of "hash hash" ¯\_(ツ)_/¯

So we can special-case hash hash and hash equal to something else in Reason. Proposal:

1. `foo..bar .= baz`
2. `foo->bar <- baz`
3. `foo."bar" = baz`

Third one might or might not work.

bsansouci:
Yes I will confirm that `##` is too heavy for me https://github.com/bsansouci/tapknight/blob/6c3d7371a5c67209d75ee6bc309db755d8cf4081/src/client.re#L57, especially formatted like this ^.

If spacing was added as a special case, life would be a tiny bit nicer. 

Could we get `foo->bar = baz` to work? otherwise I'd prefer 2nd.

Getting sidetracked:
It'd be interesting if we could open a syntax locally to a scope. For example, it'd be cool to locally override `{}` and `foor.bar = bas` to be `[%bs.obj {}]` and `foo##bar #= baz`. That overriding could maybe look like
```
let obj = JS.(
  bar.baz = 10; /* bar##baz #= 10; */
  let a = {}; /* let a = [%bs.raw "{}"] */
 print_endline bar.baz; /* console.log(bar##baz); */
 bar
);

obj.baz /* error, this is of type Js.obj and not a record with a field bas */
```

It would be like you open a module. It would probably become a bit strange if you need to use records inside a `JS.(...)` block.

chenglou:
> It'd be interesting if we could open a syntax locally to a scope 

Might be dangerous if you refactor code and pull e.g. a function out.

The thing about `foo->bar = baz` and `foo."bar" = baz` and `foo..bar = baz` to infer what to desugar `=` to, is that I'm not sure if this is possible: `myRefToFooBar = baz`. In which case this'd fail. Others know better.

bobzhang:
I like proposal 1, another important thing is to make merlin easy to auto-complete
cc @let-def

chenglou:
Adding a 4th proposal that @bobzhang mentioned.

freebroccolo:
Out of these options I would prefer (3) or (4). I don't like (1) much because I feel like having more instances of `.` and `..` in the syntax is just going to look confusing. I already kind of wish we would only have `...`.

[...]

PR #941, opened 2017-01-08
Improved BuckleScript method call syntax

SanderSpies:
#849

Changes `Foo##bar` into `Foo."bar"`

Note that `##` still needs to used for scenario's like: `Foo##(something."else").`

Also `#=` has gotten an extra post space during printing as it otherwise will fail to be idempotent.


chenglou:
Context: we're sticking with `#=` for the moment.

cc spam @bsansouci @Schmavery @jordwalke @yunxing, adding a space after `#=` ok? I recall some precedence problems around it?

> Note that ## still needs to used for scenario's like: Foo##(something."else").

I don't think @bobzhang made such thing work. This compiles to something weird: https://bloomberg.github.io/bucklescript/js-demo/?gist=0d4169c80f571302b5a163ee01556a7e

SanderSpies:
@jordwalke mentioned a better form: `Foo["bar"]["more"]`. I'll update this PR unless there are objections.

chenglou:
That'd be a bit more tedious than `##` and might mislead people into doing `foo[stringRef]` (though we can warn against that)

SanderSpies:
It requires also some more work, so let's keep it as is.

bobzhang:
@chenglou your example probably is a bug on BuckleScript, it should fail with a nice error msg

bobzhang:
How about using '.=' instead of '#='

SanderSpies:
@bobzhang could easily be added (did a quick experiment).

formatTest/unit_tests/expected_output/bucklescript.re
```diff
@@ -2,70 +2,70 @@ bla#=10;
...
+test."var"#=(Some (-10));
 
 obj##.prop;
 
 obj##.prod := exp;
 
-preview##style##border#=(
-                          Js.string "1px black dashed"
-                        );
+preview."style"."border"#= (
```
SanderSpies: Something goes wrong here.

chenglou: @bsansouci

chenglou: @bsansouci @jordwalke

bsansouci: What's wrong exactly?

SanderSpies: @bsansouci 
https://github.com/facebook/reason/pull/941/files#diff-239206428e68c12d0f82323cb9944abaL17

SanderSpies: Nevermind, I was looking wrong.

Schmavery:
Adding a space (before and) after #= shouldn't result in any parsing issues.  In fact, it should fix #856

SanderSpies:
Added support for `.=`.

chenglou:
Is this ready?

SanderSpies:
Yes, ready for review

chenglou:
Fack. cc @yunxing is this good to go? If it is, @bobzhang everything's ready on BS' side right?

bobzhang:
this is good, would you do a sanity check on your side that no `#` character in method name

SanderSpies:
Updated with extra check for hash within method name.

yunxing:
The parser && printer part looks good to me. 

@chenglou Feel free to merge this if the syntax also looks good to you.

chenglou:
This PR works but completely wrecks editor highlighting.

screenshot 2017-02-01 01 01 20
screenshot 2017-02-01 01 02 00

I can fix this for atom (and I guess vscode by extension), @jordwalke what about vim? cc @yunxing for emacs

chenglou:
Will work on making the highlighting work then merge this tomorrow.

Issue #940, opened 2017-01-08
refmt produces untypable output for exception match cases


jberdine:
refmt on
```
switch (f ()) {
| x => Some x
| exception (Sys_error _ as exc) => raise exc
};
```
yields
```
switch (f ()) {
| x => Some x
| exception Sys_error _ as exc => raise exc
};
```
which does not type-check with `Error: Unbound value exc`.  Warning 11 (unused match case) is also triggered.

It seems that the parens in the exception match case are required for correct parsing, but refmt removes them anyway.

jordwalke:
Thanks, I'll get to this next.

IwanKaramazow:
Made a pr with a fix above.

chenglou:
Pretty good response rate we're getting =D

PR #942, opened 2017-01-08
Fix wrapping of exception alias patterns, issue 940

IwanKaramazow:
Fix for https://github.com/facebook/reason/issues/940

The wrapping around `Ppat_exception`'s pattern with aliases was non-existent before. 
What I did was parameterizing `method pattern` to add a wrapping whenever a `Ppat_exception` occured with an alias.

### DIFF
--- a/formatTest/unit_tests/input/wrappingTest.re
+++ b/formatTest/unit_tests/input/wrappingTest.re
@@ -1946,16 +1946,3 @@ let tryingTheSameInLocalScope = {
   let blah = fun a b => a;       /* Done */
   let blah a b => a;             /* Done (almost) */
 };
-
-
-/**
-   Issue 940: https://github.com/facebook/reason/issues/940
-   The parens in the exception match case are required for correct parsing
-   i.e. (Sys_error _ as exc) instead of Sys_error _ as exc
-   The latter doesn't type-check with Error: Unbound value exc.
-   Warning 11 (unused match case) is also triggered.
- */
-switch (f ()) {
-| x => Some x
-| exception (Sys_error _ as exc) => raise exc
-};
--- a/formatTest/unit_tests/input/wrappingTest.re
+++ b/formatTest/unit_tests/input/wrappingTest.re
@@ -1946,3 +1946,16 @@ let tryingTheSameInLocalScope = {
   let blah = fun a b => a;       /* Done */
   let blah a b => a;             /* Done (almost) */
 };
+
+
+/**
+   Issue 940: https://github.com/facebook/reason/issues/940
+   The parens in the exception match case are required for correct parsing
+   i.e. (Sys_error _ as exc) instead of Sys_error _ as exc
+   The latter doesn't type-check with Error: Unbound value exc.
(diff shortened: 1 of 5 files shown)


formatTest/unit_tests/input/wrappingTest.re
```diff
@@ -1946,3 +1946,16 @@ let tryingTheSameInLocalScope = {
...
 };
+
+
+/**
+   Issue 940: https://github.com/facebook/reason/issues/940
+   The parens in the exception match case are required for correct parsing
+   i.e. (Sys_error _ as exc) instead of Sys_error _ as exc
+   The latter doesn't type-check with Error: Unbound value exc.
+   Warning 11 (unused match case) is also triggered.
+ */
```
jordwalke: It would be nice to see a *non* aliased exception match as well. This is also something that should probably done in `typeCheckedTests` to doubly verify that we are parsing/printing correctly.

src/reason_pprint_ast.ml
```diff
@@ -3089,7 +3096,12 @@ class printer  ()= object(self:'self)
           | Ppat_lazy p ->formatPrecedence (label ~space:true (atom "lazy") (self#simple_pattern p))
           | Ppat_extension e -> self#extension e
           | Ppat_exception p ->
-              makeList ~postSpace:true [atom "exception"; self#pattern p]
+              (*
+                An exception pattern should be wrapped in (...)
+                Example: (Sys_error _ as exc)
+                  results in incorrect parsing with type error otherwise.
+              *)
+              makeList ~postSpace:true [atom "exception"; self#pattern ~wrap:true p]
```
jordwalke: Oh, we already have a helper for generating a "wrapped" pattern. It's called `self#simple_pattern p`.

jordwalke: So, could this be:

```
 makeList ~postSpace:true [atom "exception"; self#simple_pattern p]
```

src/reason_pprint_ast.ml
```diff
@@ -3089,7 +3096,12 @@ class printer  ()= object(self:'self)
           | Ppat_lazy p ->formatPrecedence (label ~space:true (atom "lazy") (self#simple_pattern p))
           | Ppat_extension e -> self#extension e
           | Ppat_exception p ->
-              makeList ~postSpace:true [atom "exception"; self#pattern p]
+              (*
+                An exception pattern should be wrapped in (...)
```
jordwalke: Also, at *parse* time we are a little more liberal - we accept things like:

```
| exception IntAndString i s =>...
```

It's just that the rules for what goes to the right of the exception are a little (too) nuanced. It accepts "non simple" parameters, *except* in the case of `as`. So with your diff, it consistently applies "simplification" to the `exception` argument, and that seems fine to me (I think as I mentioned we should use the `simple_pattern` method though).

IwanKaramazow:
Comments & suggestions are integrated

yunxing:
Looks perfect, thanks @IwanKaramazow!

jordwalke:
Thanks again, @IwanKaramazow !

PR #943, opened 2017-01-09
Deprecating explicit --use-stdin

jasonrose:
This PR makes `refmt` behave more like other command line tools like `python`, `ocaml`, and others that can both take a pipe as stdin or specify a file on the command line.

In a situation where the user specifies both `--use-stdin` and a filename, such as`echo 'let _ = ()' | refmt --print re --use-stdin thisdoesntexist.ml`, `refmt` would try to infer the source type from the filename, making the `--parse` arg optional.

This PR changes the behavior to always prefer a filename if it is present, ignoring the piped stdin if it is present.  This also means that both `--print` and `--parse` are always required when using stdin.

### DIFF
--- a/editorSupport/VimReason/plugin/reason.vim
+++ b/editorSupport/VimReason/plugin/reason.vim
@@ -23,7 +23,7 @@ endif
 " From auto-format plugin:
 " https://github.com/Chiel92/vim-autoformat/blob/master/plugin/autoformat.vim
 let g:vimreason_reason = "refmt"
-let g:vimreason_args_expr_reason = '"--use-stdin true --print re --interface " .  (match(expand("%"), "\\.rei$") == -1 ? "false " : "true ")  . expand("%")'
+let g:vimreason_args_expr_reason = '"--print re --interface " .  (match(expand("%"), "\\.rei$") == -1 ? "false " : "true ") . " --parse " . expand("%:e")'
 
 function! Strip(input_string)
   return substitute(a:input_string, '\s*$', '\1', '')
--- a/editorSupport/sublime-reason/Reason.py
+++ b/editorSupport/sublime-reason/Reason.py
@@ -28,7 +28,6 @@ def reason_command_line(file_name):
 
     return [
         refmt,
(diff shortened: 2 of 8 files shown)


editorSupport/VimReason/plugin/reason.vim
```diff
@@ -23,7 +23,7 @@ endif
 " From auto-format plugin:
 " https://github.com/Chiel92/vim-autoformat/blob/master/plugin/autoformat.vim
 let g:vimreason_reason = "refmt"
-let g:vimreason_args_expr_reason = '"--use-stdin true --print re --interface " .  (match(expand("%"), "\\.rei$") == -1 ? "false " : "true ")  . expand("%")'
+let g:vimreason_args_expr_reason = '"--print re --interface " .  (match(expand("%"), "\\.rei$") == -1 ? "false " : "true ") . " --parse " . expand("%:e")'
```
jasonrose: The vim plugin is the only situation in the codebase ATM where the source file type was being inferred from the current buffer's file extension; I had to make it explicit so since we are deprecating this exact behavior.

editorSupport/sublime-reason/Reason.py
```diff
@@ -28,7 +28,6 @@ def reason_command_line(file_name):
 
     return [
         refmt,
-        '--use-stdin', 'true',
         '--parse', 're',
         '--interface', is_interface(file_name),
```
jasonrose: I need to fix it here since it exhibits the same situation as the vim extension above.

yunxing: Interesting. I think you don't need to fix it for emacs since it always copies the buffer into a temporary file before formatting.

yunxing:
Thanks, diff looks good. Let's merge when CI is green.

editorSupport/sublime-reason/Reason.py
jasonrose: Nevermind, it's reformatting as a binary which is what the method claims to be doing and so this works.

jordwalke:
Thanks!

PR #944, opened 2017-01-10
Attributes in class type field method

ngzhian:
Partially address #710

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/attributes.re
+++ b/formatTest/typeCheckedTests/expected_output/attributes.re
@@ -284,6 +284,10 @@ and anotherClassType = {
 }
 [@@structureItem];
 
+class type _y = {
+  pub height: int [@@bs.set]
+};
+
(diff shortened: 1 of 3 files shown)


yunxing:
@ngzhian Thanks for the fix! The diff looks very good and I like the additional test. 

Why do you call this a partial fix? Is it because we are still missing the class attributes? 

In #710 you also mentioned some indentation issues, is that fixed?

ngzhian:
This is partial because it only handles attributes for class type field, but not for the entire class type.

e.g.
```
class type _y = object 
  method height : int [@@bs.set] <-- this is converted
  (* [height] is setter only *)
end [@bs] <-- this is not converted
```
This PR does not have indentation issues because those issues only affect class type attributes. I intend to tackle that in a separate PR.

yunxing:
@ngzhian Makes much more sense now. Thank you!

Issue #945, opened 2017-01-10
Invokation of refmt without arguments produces unfriendly error message


andreypopp:
That's what I got:
```
% ./_install/bin/refmt                                                                                                                                                         
Fatal error: exception Refmt_impl.Invalid_config("Cannot determine default implementation parser for filename ''.")
```
I understand the origin of this error now but I thought that compiled version of refmt is broken at first.

chenglou:
Doesn't it display the help page now? Which version are you on?

andreypopp:
Tried this branch which is a port of 52c3b0215c89d82422322e31db4098fd0f8ca7e3

jasonrose:
I caused this regression with my latest PR so I'll fix it.

yunxing:
Should be fixed now, thanks!

PR #946, opened 2017-01-10
Fixing regression with no-args refmt invocation

jasonrose:
Addresses #945


yunxing:
Looks good. Thanks for the quick fix!

yunxing:
CI finally passes!

PR #947, opened 2017-01-11
Clarify Reason npm package is lowercase

AlexTes:
In the 'Install Via `npm` block Reason is referred to in a code block. Reading on it becomes clear the Reason project refers to Reason in a code block always, having nothing to do with the names of different Reason related things such as the npm package. Which is fine, even good I'd say. But without this knowledge seeing `Reason` everywhere and no mention of the package name had me guess it to be 'Reason' which is wrong and unfortunately results in a cryptic npm error: `npm ERR! Cannot read property 'path' of null`. Additionally, since these are code blocks I and I think many others honor the capitalization in them.

We can avoid letting the user guess and avoid possibly steering them in the wrong direction (by writing `Reason` in the block) by mentioning the package name. The project does the same in other places in the README where a name for something Reason related is required; I suggest we do the same for the npm block :].

I will also see if I can help improve this experience on the npm side ^_^ .

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Reason: Build Systems Rapidly [![Build Status](https://travis-ci.org/facebook/re
 
 Install Via `npm`
 ----------------
-Installing `Reason` via `npm` is the easiest way to get started. (`npm` version > 3.0 is required - [install here](https://nodejs.org/en/download/current/)).
+Installing `Reason` via `npm` is the easiest way to get started. Like so: `npm install reason`. (`npm` version > 3.0 is required - [install here](https://nodejs.org/en/download/current/)).
 
 #### Reason Project
 


chenglou:
Nice touch =D Thanks?
Npm disabled upper-case letters in package name afaik

AlexTes:
omg its @chenglou :O ! (loved your talk on abstraction!)
I just noticed I got this completely wrong. So let me close the PR and explain.
EDIT: Just got merged 😓. Your speed is amazing! but timing unfortunate.

After reading again and playing close attention to the headers I noticed installing Reason with npm was never the intention of the block. The intention is to clone the ReasonProject starter kit and install it's dependencies through npm. I'd just remove the first line in that case which in my reading gives a different instruction than the nearly identical line following it in the next sub-section. I'm talking about removing 'Installing Reason via npm is the easiest way to get started.'.

Now that my addition has ironically in my own opinion been wrongfully merged, how do you want to proceed @chenglou ? I can open a PR to remove the mentioned line and my addition which would result in users cloning the starter kit and installing its dependencies with npm, which I think the 'Install Via `npm`' block always intended 😅 .

chenglou:
Hello!
Sorry, I didn't check the surround context before merging. Is there anything else to edit beside a revert of the diff? If not, I can just revert.

AlexTes:
@chenglou Yes. The suggested diff in my previous comment is my addition plus the sentence prior to it that confused me twice now 😅 

ORIGINAL:
```
## Install Via npm

Installing Reason via npm is the easiest way to get started. Like so: npm install reason. (npm version > 3.0 is required - install here).

### Reason Project

Installing the ReasonProject starter kit using npm is the easiest way to get started with Reason. [...]
```

SUGGESTED:
```
## Install Via npm

(npm version > 3.0 is required - install here).

### Reason Project

Installing the ReasonProject starter kit using npm is the easiest way to get started with Reason. [...]
```

chenglou:
Thanks for the suggestion! Removed the parentheses too =).
Keep these coming! We like these kind of polishes.

AlexTes:
Thanks, will do, this is awesome, and you are awesome.
Have a great day!

PR #950, opened 2017-01-12
Fix #914.

### LINKED ISSUE #914: bs.obj sugar key wrapped for no reason
Repro:

`{"/foo": bar}` gets formatted to `{"(/foo)": bar};`

@bsansouci

bsansouci:
This is a pretty simple fix.
I use the existent parameter `withStringKeys` to when to format the key
as an ident, or as a simple string.

For now we don't support `Ldot` and `Lapply` in the string keys. I'm not sure what the behavior should be in this case.

cc @chenglou

### DIFF
--- a/formatTest/unit_tests/expected_output/bucklescript.re
+++ b/formatTest/unit_tests/expected_output/bucklescript.re
@@ -86,3 +86,5 @@ let d = {
     "a": fun arg1 arg2 => arg1 + arg2
   }
 };
+
+let a = {"/foo": 10};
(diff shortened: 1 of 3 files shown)


chenglou:
cc @yunxing

chenglou:
every weird string key should be supported though; Is that hard? I believe the same will apply for #941

bsansouci:
@chenglou I think `Ldot` and `Lapply` are kinds of identifier. In this case, we parse the key as an `Lident` https://github.com/facebook/reason/blob/c2661cabc1405b373a84047a61afffd0e121aef2/src/reason_parser.mly#L3483. So there's no way (right?) to have anything else in those case. That'll support _any_ string (because we parse any string).

yunxing:
Accidentally lost track on this, looking into it now.

yunxing:
Very sorry for the delay. Diff looks good. Let's merge this!

PR #951, opened 2017-01-12
Give a nicer error when using a wrong module name

SanderSpies:
`module foo = {`

will now give the error: 

`A module's name needs to begin with a upper-case letter`


yunxing:
Nice :D

PR #952, opened 2017-01-12
Allow `module Foo {` as module syntax.

SanderSpies:


yunxing:
That's fast, thanks. 

I have a few questions:

1. Does this work with module with types? Something like `module A : B = {}`. 

2. Does this work with functors? 

I feel it would be confusing to the user, if the syntactic sugar only works in a limited use cases.. What do you think?

SanderSpies:
Agreed, let's polish it a bit more.

yawaramin:
Is the intention to remove `module A = { ... };` syntax, or keep that as well? Personally I prefer a single syntax for each thing.

chenglou:
@SanderSpies are you still tripping on this? Kinda agree with @yawaramin but I'm a bit torn. It _is_ nice to format to the correct syntax without erroring.

SanderSpies:
@chenglou Currently not working on this. I'm focused on attributes at the moment. Closing this for now. We can revisit later if needed.

Issue #637, opened 2016-07-13
Postfix syntax for variants


chenglou:
Most probably not gonna work out because of the sheer number of parsing conflicts it'll create. But I'm filing this anyway. The reason I'm wondering about this is for simulating units, e.g. `1Px`, `50Percent`, `1min`, etc. They'll be desugared to e.g. `Px 1` and you'd prevent accidental mis-addition of two incompatible units together.

jordwalke:
It actually wouldn't create many parsing conflicts I believe because these would in theory be extracted in the tokenizer. The question is just a matter of specifying *which* variants should be lexed/parsed/printed this way.

yunxing:
One trick I was thinking is to add an attribute to the node, annotating how it should be printed. 

If we eventually have our own ocaml printer, we can just ignore that special attribute.

jordwalke:
Yeah, that would work. Is there any precedence for postfix sugar in other languages? tbph, the set of code that I'd want unit of measurement tracking are exactly the cases where I would not want to pay for the extra boxing of a variant tag.

Schmavery:
Hopefully wouldn't conflict with anything trying to use [extension literals](http://caml.inria.fr/pub/docs/manual-ocaml/extn.html#sec253)?

saschatimme:
This would be really neat for a CSS solution like elm-css!

cullophid:
I just starter working on https://github.com/SentiaAnalytics/bs-css.
This would be majs it much nicer

PR #953, opened 2017-01-12
Implement postfix parsing

SanderSpies:
Work in progress.

Changes:

`50Cats` 

to 

`Cats 50`.

Printing still needs to be done to finish this.

### DIFF
--- a/src/reason_lexer.mll
+++ b/src/reason_lexer.mll
@@ -229,6 +229,31 @@ let remove_underscores s =
       |  c  -> Bytes.set b dst c; remove (src + 1) (dst + 1)
   in remove 0 0
 
+let remove_numbers s =
+  let l = String.length s in
+  let b = Bytes.create l in
+  let rec remove src dst =
+    if src >= l then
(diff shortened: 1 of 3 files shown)


src/reason_lexer.mll
```diff
@@ -359,6 +386,11 @@ rule token = parse
   | "=?"
       (* Need special label extractor? *)
       { OPTIONAL_NO_DEFAULT }
+  | postfix_chars
+      {
+        let l = Lexing.lexeme lexbuf in
+        POSTFIX (remove_numbers l, int_of_string (remove_chars l))
```
yunxing: Nit: instead of using two functions, is it possible to just do one pass and return both number part and string part?

SanderSpies: Will do

chenglou:
Thanks for the change! Btw #637 is pretty early. Let's see if people still want postfix...

SanderSpies:
Added printer support also and tests.

SanderSpies:
@chenglou This PR will be waiting patiently

cullophid:
:+1

PR #954, opened 2017-01-12
Format class type instance attributes

### LINKED ISSUE #710: Class type ppx attributes disappear

ngzhian:
This is an attempt to fix #710 while maintaining an acceptable pretty print format.
Here's what this PR achieves, given: `class type_z = { pub height: int } [@@bs];`, it formats it to
```
class type _z = {
  pub height: int
}
[@@bs];
```

This is similar to how a virtual class is formatted [here](https://github.com/facebook/reason/blob/master/formatTest/typeCheckedTests/expected_output/oo.re#L34-L55).

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/attributes.re
+++ b/formatTest/typeCheckedTests/expected_output/attributes.re
@@ -288,6 +288,11 @@ class type _y = {
   pub height: int [@@bs.set]
 };
 
+class type _z = {
+  pub height: int
+}
+[@@bs];
+
 module NestedModule = {
   [@@@floatingNestedStructureItem hello];
 };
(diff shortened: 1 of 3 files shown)


yunxing:
Looks nice, thanks. I haven't touched this part of the codebase yet, so not sure what kind of challenges are there to correct indent this. Maybe @jordwalke has more contexts. 

The diff is definitely good enough for now though. Thanks so much for the fix!

PR #955, opened 2017-01-12
Don't break attributes over multiple lines.

IwanKaramazow:
Improves attribute formatting:
* doesn't break attributes on multiple lines, when it isn't necessary.
* provides better identation when breaking the declaration over multiple lines

Example rehydrate, before:
```rust
external createElement : reactClass => props::Js.t {..}? => array reactElement => reactElement = "createElement" [@@bs.splice] [@@bs.val] [@@bs.module
                                                                    "react"
                                                                    ];

external createClassInternalHack : Js.t 'classSpec => reactClass = "createClass" [@@bs.val] [@@bs.module
                                                                    "react"
                                                                    ];

external createCompositeElementInternalHack : reactClass =>
                                              Js.t {.. reasonProps : 'props} =>
                                              array reactElement =>
                                              reactElement = "createElement" [@@bs.val] [@@bs.module
                                                                    "react"
                                                                    ] [@@bs.splice];
```

After:
```rust
external createElement : reactClass => props::Js.t {..}? => array reactElement => reactElement =
  "createElement" [@@bs.splice] [@@bs.val] [@@bs.module "react"];

external createClassInternalHack : Js.t 'classSpec => reactClass =

type reactElementthisIsasupersupersupersupserupserupserupseruplongtype;

  reactClass =>
  Js.t {.. reasonProps : 'props} =>
  array reactElementthisIsasupersupersupersupserupserupserupseruplongtype =>
  reactElement =
  "createElement" [@@bs.val] [@@bs.module "react"] [@@bs.splice];

  reactClass => Js.t {.. reasonProps : 'props} => array reactElement => reactElement =
  "createElement" [@@bs.val] [@@bs.module "react"] [@@bs.splice];
```

### DIFF
--- a/src/reason_pprint_ast.ml
+++ b/src/reason_pprint_ast.ml
@@ -5065,6 +5065,36 @@ class printer  ()= object(self:'self)
     | Pstr_attribute (s, _) -> (not (s.txt = "ocaml.text") && not (s.txt = "ocaml.doc"))
     | _ -> true
 
+  (*
+    external render : reactElement => element => unit =   (* frstHalf *)
+      "render" [@@bs.val] [@@bs.module "react-dom"];      (* sndHalf *)
+
+    To improve the formatting with breaking & indentation:
+      * consider the part before the '=' as a label
+      * combine that label with '=' in a list
+      * consider the part after the '=' as a list
+      * combine both parts as a label
+  *)
+  method primitive_declaration vd =
+    let attrs = List.map (fun x -> self#item_attribute x) vd.pval_attributes in
+    let lblBefore =
+      label ~space:true
+        (makeList ~postSpace:true
+          [atom "external"; protectIdentifier vd.pval_name.txt; atom ":"])
+        (self#core_type vd.pval_type)
+    in
+    let frstHalf = makeList ~postSpace:true [lblBefore; atom "="] in
+    let string_literals = makeSpacedBreakableInlineList (List.map self#constant_string vd.pval_prim) in
+    (* when there aren't any attrs, add a final semi here *)
+    let prim = if (List.length attrs) == 0
+      then appendSep false ";" string_literals
+      else string_literals
+    in
+    let sndHalf =
+      makeList ~inline:(true, true) ~postSpace:true
+        [prim; makeList ~postSpace:true attrs]
+    in
+    label ~space:true frstHalf sndHalf
 
   method class_instance_type x = match x.pcty_desc with
     | Pcty_signature cs ->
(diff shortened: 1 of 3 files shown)


yunxing:
Looks good. Let's wait for the CI.

chenglou:
Should these attributes break on many lines? They can contain arbitrary payloads. Seems like they should be arranged like function arguments currently?

IwanKaramazow:
I think this is ok for now, `#structure_item` will handle the output with breaking from more complex structures.

yunxing:
@chenglou I think putting them into a single line makes some sense, as in we always want to view the attributes with the node it's decorating? 

Altnative is to format the example as 

```
external createClassInternalHack : Js.t 'classSpec => reactClass = "createClass" [@@bs.val]          
                                                                            [@@bs.module "react"];
```
which I think is also acceptable.

IwanKaramazow:
How should I proceed?

chenglou:
What if they're long? 3 Tags for BS isn't _too_ uncommon. They'd break the column limit hard. Why not

```
external createClassInternalHack : Js.t 'classSpec => reactClass = "createClass" 
  [@@bs.val] [@@bs.module "react"];
```

IwanKaramazow:
I kinda like the one liner more, but agreed it does make sense to break the attributes.

@SanderSpies, @jordwalke  What do you think?

chenglou:
No hard opinion on this fwiw. Though maybe this'd work better than the previous example:

```
external createClassInternalHack : Js.t 'classSpec => reactClass = 
```

yunxing:
@chenglou I like your example better.

@IwanKaramazow Let us know if it is hard to implement this (you would need to make a Label before and after the "=", which could be not that trivial). If it is hard, let's just push what we have currently.

IwanKaramazow:
@yunxing I think I made @chenglou 's version work.
See latest commit.

Example of refmt on rehydrate:
```ocaml
external createElement : reactClass => props::Js.t {..}? => array reactElement => reactElement =
  "createElement" [@@bs.splice] [@@bs.val] [@@bs.module "react"];

external createClassInternalHack : Js.t 'classSpec => reactClass =

type reactElementthisIsasupersupersupersupserupserupserupseruplongtype;

  reactClass =>
  Js.t {.. reasonProps : 'props} =>
  array reactElementthisIsasupersupersupersupserupserupserupseruplongtype =>
  reactElement =
  "createElement" [@@bs.val] [@@bs.module "react"] [@@bs.splice];

  reactClass => Js.t {.. reasonProps : 'props} => array reactElement => reactElement =
  "createElement" [@@bs.val] [@@bs.module "react"] [@@bs.splice];
```

yunxing:
@IwanKaramazow Thanks, looking into this now.

IwanKaramazow:
@yunxing any remarks?

src/reason_pprint_ast.ml
```diff
@@ -5065,6 +5065,36 @@ class printer  ()= object(self:'self)
...
+      label ~space:true
+        (makeList ~postSpace:true
+          [atom "external"; protectIdentifier vd.pval_name.txt; atom ":"])
+        (self#core_type vd.pval_type)
+    in
+    let frstHalf = makeList ~postSpace:true [lblBefore; atom "="] in
+    let string_literals = makeSpacedBreakableInlineList (List.map self#constant_string vd.pval_prim) in
+    (* when there aren't any attrs, add a final semi here *)
+    let prim = if (List.length attrs) == 0
+      then appendSep false ";" string_literals
```
yunxing: Took me a while to understand why you add a ";" here.  I think it is probably due to some tweaks we put into comment interleaving logic?

yunxing:
@IwanKaramazow Sorry for the delay. Diff looks good. I think I should review the usage of appendSep logic a bit.

PR #956, opened 2017-01-12
Simple whitespace cleanup

tekknolagi:
This makes the spacing in the file more consistent

### DIFF
--- a/src/refmt_impl.ml
+++ b/src/refmt_impl.ml
@@ -161,7 +161,8 @@ let () =
       in
       let _ =
         if not parsedAsInterface then
-          raise (Invalid_config ("The file parsed does not appear to be an interface file.")) in
+          raise (Invalid_config ("The file parsed does not appear to be an interface file."))
+      in
       let _ = Reason_pprint_ast.configure
           ~width: print_width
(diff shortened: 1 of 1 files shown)


src/refmt_impl.ml
```diff
@@ -209,8 +209,9 @@ let () =
           raise (Invalid_config ("Invalid --parse setting for interface '" ^ s ^ "'."))
         )
       in
-      let _ = if parsedAsInterface then
-          raise (Invalid_config ("The file parsed does not appear to be an implementation file.")) in
+      let _ =
+          if parsedAsInterface then
+              raise (Invalid_config ("The file parsed does not appear to be an implementation file.")) in
```
yunxing: Maybe move 'in' to the next line so it is more consistent? 

(This is why we need reason to automatically make the file consistent!)

chenglou:
Made the change myself in the online editor. Github magic.

chenglou:
Aaaaaand I forgot a spot. Forget about that magic.

Pushed a new change.

tekknolagi:
Oh alright! I will go ahead and change the other instance.

PR #957, opened 2017-01-12
Docs nitpick: Operations aren't primitives

quicksnap:
Brought up in discord.. discuss!


yunxing:
Thanks. I think the original doc meant to say "Primitives types, primitive operations, primitive containers" in a board sense. But it causes confusion we should fix that.

PR #960, opened 2017-01-13
Try to use replay(reason playground) in reason docs

yutongp:
@yunxing Try to use replay to embed some code snippet in reason docs

screen shot 2017-01-13 at 1 57 29 am
screen shot 2017-01-13 at 1 58 14 am

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -184,7 +184,16 @@ evaluates to the `integer` `3`.  The `{}` block ensures that `one` and `two`
 are not accessible after `three` is computed.
 
 <pre>
-<p data-id="-P8x86Mg6w" data-host="replay-154206.appspot.com" data-height="256" data-font-size="12px" class="replay" ></p>
+<p data-id="-P8x86Mg6w" data-host="replay-154206.appspot.com" data-height="256" data-font-size="12px" class="replay-embed" >
+<!--- Snippet Generated Through Replay
+let three = {
+  let one = 1;
+  let two = 2;
+  one + two
+};
+print_int three
+-->
+</p>
 </pre>
 
 > Note that the last item may optionally include a terminating semicolon.
--- a/docs/index.html
+++ b/docs/index.html
@@ -191,7 +191,9 @@ let three = {
   let two = 2;
   one + two
 };
-print_int three
+/* one and two not accessible here! */
+
+print_int three;
 -->
 </p>
 </pre>
--- a/docs/index.html
+++ b/docs/index.html
@@ -22,6 +22,7 @@
   <script src='theme-white/script.js'></script>
   <link  href='support/theme.css' rel='stylesheet'>
   <script src='support/theme.js'></script>
+  <script async src="https://replay-154206.appspot.com/static/ei.js"></script>
 
 
   <script id="markdown" type="text/markdown" src="index.html">
@@ -181,14 +182,10 @@ to the last item in the sequence. In this example, the binding `three`
 evaluates to the `integer` `3`.  The `{}` block ensures that `one` and `two`
 are not accessible after `three` is computed.
 
-```reason
-let three = {
-  let one = 1;
-  let two = 2;
-  one + two
-};
-/* one and two not accessible here! */
-```
+<pre>
+<p data-id="-P8x86Mg6w" data-host="replay-154206.appspot.com" data-height="256" data-font-size="12px" class="replay" ></p>
+</pre>
+
 > Note that the last item may optionally include a terminating semicolon.
 
 #### Constant bindings
--- a/docs/index.html
+++ b/docs/index.html
@@ -22,8 +22,9 @@
   <script src='theme-white/script.js'></script>
   <link  href='support/theme.css' rel='stylesheet'>
   <script src='support/theme.js'></script>
-  <script async src="https://replay-154206.appspot.com/static/ei.js"></script>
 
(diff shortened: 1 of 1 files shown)


yunxing:
@yutongp This looks awesome! Could you help add some missing feature before we merge this? (otherwise it would be confusing to other people and maintainers)

- [ ] Could you provide a link to the replay server repo with some documentation about how to set up? 
- [ ] (See comment) Is it possible to change the api a bit so that we can still have the "code" in the API call, as opposed to just a hash value, which is not maintainable?

docs/index.html
```diff
@@ -181,14 +183,10 @@
...
-```reason
-let three = {
-  let one = 1;
-  let two = 2;
-  one + two
-};
-/* one and two not accessible here! */
-```
+<pre>
+<p data-id="-P8x86Mg6w" data-host="replay-154206.appspot.com" data-height="256" data-font-size="12px" class="replay" ></p>
```
yunxing: This api call has only a hash value. Is it possible to preserve the original code as part of the api call so it is more maintainable? 

As an example, see Golang's plugin code:
image

yutongp: IMO This is not easy. This is the trade off of an embed mode. With embed we achieve:
1. simple dependency
2. async load
3. no cross origin concern
4. very easy to use 

we may lose some of these:
1. super flexible css styling 
2. change code snippet directly on file

The example you paste have all playground required js and css files included in the html file directly and can post to golang.org/compile to run code.
in order to do that with replay we need:
1. include all replay dependencies:
```
<link href='https://fonts.googleapis.com/css?family=Montserrat:400,700' rel='stylesheet' type='text/css'>
<script src="https://cdnjs.cloudflare.com/ajax/libs/codemirror/5.15.2/codemirror.js"></script>
<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/codemirror/5.15.2/codemirror.css">
<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/codemirror/5.17.0/theme/monokai.min.css">
<link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/codemirror/5.17.0/theme/solarized.min.css">
<script src="https://cdnjs.cloudflare.com/ajax/libs/codemirror/5.15.2/mode/mllike/mllike.min.js"></script>
<script src="https://cdnjs.cloudflare.com/ajax/libs/codemirror/5.15.2/mode/javascript/javascript.min.js"></script>
<script src="https://cdnjs.cloudflare.com/ajax/libs/codemirror/5.15.2/addon/edit/matchbrackets.min.js"></script>
<script src="https://cdnjs.cloudflare.com/ajax/libs/codemirror/5.17.0/addon/edit/closebrackets.min.js"></script>
<script src="//ajax.googleapis.com/ajax/libs/jquery/1.8.2/jquery.min.js"></script>
<script src="/static/url.min.js"></script>
<script src="/static/playground.js"></script>
<link rel="stylesheet" href="/static/style.css">
```
2. add HTTP access control in replay to support cross origin request

I feel like this is too heavy for now compare to embed model before replay become a default code snippet editor in reason docs.

yutongp: Not sure how often you may change the code snippet, but if you just want to see the code content directly in the html file, we can add the code in the <p> tag replay generated like this:
```
<p data-id="-P8x86Mg6w" data-host="replay-154206.appspot.com" data-height="256" data-font-size="12px" class="replay" >
<code>
let three = {
  let one = 1;
  let two = 2;
  one + two
};
print_int three
</code>
</p>
```
but changing the code here will not change the code in output embedded content

yutongp:
@yichengq could you work on replay documentation?

docs/index.html
yunxing: @yutongp Thanks, this makes sense. Adding original code snippet in the comment looks good. It's a bit more effort to maintain all the "dataid" here, but I think in the short term, we can always require contributors who changes the comment here to change the "dataid" as well. 

In the long term, we should put the comments into separate files, and have a script to automatically generate the code snippet id and master html file based on those snippet files.

yunxing:
Diff looks good. Do you mind make the project public so people can help maintain it?

yunxing:
@yutongp @yichengq

kennetpostigo:
If project is open on github I would integrate it with the introduction to reason guide I wrote as well as build something similar to [A Tour of Go](https://tour.golang.org/welcome/1) .

yutongp:
https://github.com/yichengq/replay
replay is open now

kennetpostigo:
@yutongp thanks so much! This is amazing, I look forward to integrating it into resources I make for reason.

yunxing:
@kennetpostigo It would definitely be a great use case and what @yutongp and @yichengq envisioned at first, I think you can start using it today. 

In the meanwhile, we are also getting a proper domain for this.

kennetpostigo:
@yunxing awesome! some docs for this could be useful for myself and others about features this has. Going to look at this PR changes to get a handle of it. I can also wrap the embed in a react-component similar to what I did for facebook login component.

chenglou:
Ping! This is great, we should have it. If not, we can always do like http://bloomberg.github.io/bucklescript/reason-demo/ and bring the whole BuckleScript compiler (compiled to JS). This way it we don't need a server and for our examples we don't use bindings or anything so it's as if it's ocamlc.

PR #963, opened 2017-01-13
Add --output and --in-place options

### LINKED ISSUE #886: in-place transformation for refmt
Some switch that allows the transformation of a file in-place, since `cat foo.re | refmt ... > foo.re` doesn't work because of how unix pipes work.

tekknolagi:
`--output`/`-o` allow the user to specify an output file for the results of refmt.
`--in-place` sets the output the file to the input file. Fixes #886

This is currently a pretty gross solution. Looking into fixing that now.


tekknolagi:
@yunxing I think this is ready.

src/printer_maker.ml
```diff
@@ -0,0 +1,52 @@
...
+
+        val makePrinter : string option ->
+                          string ->
+                          bool ->
+                          out_channel ->
+                          Format.formatter ->
+                          ((t * Reason_pprint_ast.commentWithCategory) -> unit)
+    end
+
+let prepare_output_file output_file =
```
yunxing: let prepare_output_file = function 
| Some name -> open_out name
| None -> ()

tekknolagi: good point.

src/refmt_impl.ml
```diff
@@ -138,118 +114,48 @@ let () =
...
-        | Some "none" -> (fun (ast, comments) -> ())
-        | Some "ml" -> Reason_toolchain.ML.print_canonical_implementation_with_comments
-        | Some "re" -> Reason_toolchain.JS.print_canonical_implementation_with_comments
-        | Some s -> (
-          raise (Invalid_config ("Invalid --print setting for implementation '" ^ s ^ "'."))
-        )
-      in
-      thePrinter (ast, comments)
+    let (ast, parsedAsML) = Printer.parse !prse use_stdin filename in
+    (* As it turns out, calling formatter_of_out_channel on stdout results in
```
yunxing: Is this still true? I see later on you are calling formatter_of_out_channel on output_chan and it could be stdout.

tekknolagi: Ah, whoops. Did not mean to leave this comment in. I think I mislabeled the true culprit.

yunxing:
Looks good. Could you also help fix the conflicts?

tekknolagi:
Sure. I can try rebasing master locally

tekknolagi:
Fixed!

yunxing:
Thanks! Let's wait for CI and we can push.

tekknolagi:
How has CI not started 8 hours later?

yunxing:
@tekknolagi It has been an issue for us -- all projects in FB share the same worker pool.

Issue #962, opened 2017-01-13
Overeager refmt punning


rickyvetter:
```
module Foo = {
  type bar = {foo:Baz.foo};
}
```

goes to:

```
module Foo = {
  type bar = {foo};
};
```

IwanKaramazow:
Looking into this.

IwanKaramazow:
@rickyvetter Pull request with fix is ready for review & CI.

yunxing:
Thanks for the quick fix, @IwanKaramazow

PR #964, opened 2017-01-13
Fix overeager punning in record types (#962).

IwanKaramazow:
Fix for https://github.com/facebook/reason/issues/962
```ocaml
module Foo = {
  type bar = {foo: Baz.foo};
};
```
should not become
```ocaml
module Foo = {
  type bar = {foo};
};
```

The latter will look for a `type foo` inside `module Foo`, which isn't what we want. 
When determining if a record row should be punned,
there is now a check in place which takes types with longer identifiers - `Baz.foo` - into account.

### DIFF
--- a/src/reason_pprint_ast.ml
+++ b/src/reason_pprint_ast.ml
@@ -2265,8 +2265,10 @@ let recordRowIsPunned pld =
         | { ptyp_desc = (Ptyp_constr ({ txt; _ }, args)); _}
             when
             (Longident.last txt = name
+              (* Don't pun types from other modules, e.g. type bar = {foo: Baz.foo}; *)
+              && List.length (Longident.flatten txt) == 1
               (* don't pun parameterized types, e.g. {tag: tag 'props} *)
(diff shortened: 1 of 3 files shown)


src/reason_pprint_ast.ml
```diff
@@ -2265,8 +2265,10 @@ let recordRowIsPunned pld =
         | { ptyp_desc = (Ptyp_constr ({ txt; _ }, args)); _}
             when
             (Longident.last txt = name
+              (* Don't pun types from other modules, e.g. type bar = {foo: Baz.foo}; *)
+              && List.length (Longident.flatten txt) == 1
```
chenglou: Is it possible to pattern match on this instead of flattening?

IwanKaramazow: Sure, is fixed in the next commit.

chenglou:
Great! Will merge after green. Thanks

chenglou:
Eh no I don't have the patience for this. Merging lol

PR #967, opened 2017-01-14
Add doc about using fun w/ immediate pattern match

ckknight:
This adds information about using `fun` without an explicit argument and instead using pattern matching immediately.


yunxing:
Thank you!

Issue #966, opened 2017-01-14
Menhir error messages in Merlin


SanderSpies:
I'm picking this up. 

As discussed on Discord with @jordwalke:

> So, Menhir has two modes: Recover from parse failures + bad error messages, or no recover from parse fail but display really great error for first parse failure.
> 
> For merlin, we pass the flag --recover or whatever, which disables the great error messages.
> 
> We should find a way to join them.
> 
> So when passing --recover, first we parse not recovering, just to get the really great error message/location. Then if we caught a parse error with a really great error message, we then parse in recovering mode, and join that really great error message into the recovered parse tree.
> 
> Recovered regions (bad parses) are replaced by ppx extension nodes like [%parseError "hey this doesn't parse"]
> 
> 
> That's why merlin integration is so smooth with Reason even if you have a bad parse - we replace bad regions with valid ocaml AST (an extension node like that is valid in parts of the AST).
> 
> That happens to contain a helpful error message
> 
> So, we need to just inject one of those extension nodes into that recovered tree, that contains the really good error message.
> 
> I would try this to start:
> 1. Start changing the "--recoverable" mode which is what we use with merlin.
> 2. It currently uses Menhir's "recoverable mode", but try swapping it out with the non recoverable parsing mode so that it fails on the first error, but with a better error.
> 3. See how structured you can extract that error. How can you get the line number/column out?
> 4. If there was no error in step 3, I think we're done.
> 5. If there was an error in step 3, then we rerun the entire parser again in Menhir's "recover mode" as we do today with --recoverable. That recovered AST has portions of the tree substituted with extension nodes like [%Syntax.error blah]. So we take the resulting AST (which has recovered portions in it), and we inject literally anywhere an an extra extension node that has the best error from step 3.
> 
> This approach will ensure that it works in all editors.
> 
> Merging in that additional node at the end should be really easy - you can just "lie" about the location to match step three, and put it at the very top of the AST.
> 

cc @let-def

SanderSpies:
Fixed.

PR #969, opened 2017-01-15
Provide Menhir errors to Merlin

SanderSpies:
#966 

Work in progress.

### DIFF
--- a/src/reason_toolchain.ml
+++ b/src/reason_toolchain.ml
@@ -79,24 +79,41 @@ module S = MenhirLib.General (* Streams *)
 
 let invalidLex = "invalidCharacter.orComment.orString"
 let syntax_error_str err loc =
-  if !Reason_config.recoverable then
-    [
-      Ast_helper.Str.mk ~loc:loc (Parsetree.Pstr_extension (Syntax_util.syntax_error_extension_node loc invalidLex, []))
-    ]
-  else
-    raise err
+    match err with
+    | Location.Error err ->
+      [
+        Ast_helper.Str.mk ~loc:err.loc (Parsetree.Pstr_extension (Syntax_util.syntax_error_extension_node err.loc err.msg, []))
+      ]
+    | _ ->
+      let menhirError = Syntax_util.findMenhirErrorMessage loc in
+      match menhirError with
+        | Syntax_util.MenhirMessagesError errMessage ->
+            [Ast_helper.Str.mk ~loc:errMessage.Syntax_util.loc (Parsetree.Pstr_extension (Syntax_util.syntax_error_extension_node errMessage.Syntax_util.loc errMessage.Syntax_util.msg, []))]
+        | _ ->
+          if !Reason_config.recoverable then
+              [Ast_helper.Str.mk ~loc:loc (Parsetree.Pstr_extension (Syntax_util.syntax_error_extension_node loc invalidLex, []))]
+            else
+              raise err
 
 let syntax_error_core_type err loc =
-  if !Reason_config.recoverable then
-    Ast_helper.Typ.mk ~loc:loc (Parsetree.Ptyp_extension (Syntax_util.syntax_error_extension_node loc invalidLex))
-  else
-    raise err
+  match err with
+  | Location.Error err ->
+      Ast_helper.Typ.mk ~loc:err.loc (Parsetree.Ptyp_extension (Syntax_util.syntax_error_extension_node err.loc err.msg))
+  | _ ->
+    if !Reason_config.recoverable then
+      Ast_helper.Typ.mk ~loc:loc (Parsetree.Ptyp_extension (Syntax_util.syntax_error_extension_node loc invalidLex))
+    else
+      raise err
 
 let syntax_error_sig err loc =
-  if !Reason_config.recoverable then
-    [Ast_helper.Sig.mk ~loc:loc (Parsetree.Psig_extension (Syntax_util.syntax_error_extension_node loc invalidLex, []))]
-  else
-    raise err
+  match err with
+  | Location.Error err ->
+    [Ast_helper.Sig.mk ~loc:err.loc (Parsetree.Psig_extension (Syntax_util.syntax_error_extension_node err.loc err.msg, []))]
+  | _ ->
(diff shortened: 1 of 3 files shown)


SanderSpies:
I notice that I need to do more work here. This only works for a subset of the Menhir errors.

SanderSpies:
The AST is starting to look a lot better, however it seems that type based errors overwrite the parser ones.

For example, within my AST I can have: 
```
 Pstr_extension "merlin.syntax-error"
    [
      structure_item (_none_[1,0+-1]..[1,0+-1]) ghost
        Pstr_eval
        expression (_none_[1,0+-1]..[1,0+-1]) ghost
          Pexp_constant Const_string("A module's name needs to begin with a upper-case letter\n",None)
    ]
```

but Merlin will show a different message instead:

```
Error: unbound instance variable z
```

yunxing:
I haven't dived into this part of Merlin's codebase yet. But I think Merlin somehow reinterpret the error messages... Maybe @let-def has more contexts of this.

SanderSpies:
Some interesting findings:
```
module foo = {}
```
will generate the typed error, but
```
module Container = {
  module bar = {
  };
};
```
will generate the menhir based error.

In the AST for both situations the menhir based error shows.

src/reason_toolchain.ml
```diff
@@ -487,9 +510,14 @@ module JS_syntax = struct
        loop_handle_yacc supplier in_error checkpoint
     | I.HandlingError env ->
        if !Reason_config.recoverable then
+         (
+         let loc = last_token_loc supplier in
+         let state = state checkpoint in
```
SanderSpies: It would be nice if we could call this without side-effects.

Without this: all the errors are visible, but not the nice Menhir ones.
With this: sometimes only shows 1 Menhir messages error

yunxing: This has been a while so my memory may be blended -- I remember that Menhir functions are mostly side-effects free ? (e.g., checkpoint won't change after this function). I can take a deeper look to validate.

SanderSpies: Thanks, that would be great.

jordwalke:
About your problem with type errors overwriting menhir ones. If the AST we give to Merlin has no reference to `z` (because that entire region has been entirely replaced by the extension point), then how can `Merlin` show an error about `z`? Are you sure you're feeding the right AST to merlin?

src/reason_toolchain.ml
SanderSpies: Example test case that goes wrong:

```
module Y = {
  type z = int;

  module Y = {


    module Z = {

    };
  };

  let a b = b + 1;

  let z a = a + 2;

  let o b = > b + 2;

};

```

SanderSpies:
@jordwalke Apologies, this is the complete AST that I'm feeding Merlin:

```
[
  structure_item (test.re[1,0+7]..[1,0+8])
    Pstr_extension "merlin.syntax-error"
    [
      structure_item (_none_[1,0+-1]..[1,0+-1]) ghost
        Pstr_eval
        expression (_none_[1,0+-1]..[1,0+-1]) ghost
          Pexp_constant Const_string("A module's name needs to begin with a upper-case letter\n",None)
    ]
  structure_item (test.re[1,0+7]..[3,14+2])
    Pstr_eval
    expression (test.re[1,0+7]..[3,14+1])
      Pexp_setinstvar "y" (test.re[1,0+7]..[1,0+8])
      expression (test.re[1,0+11]..[3,14+1])
        Pexp_object
        class_structure
          pattern (test.re[1,0+12]..[1,0+12])
            Ppat_var "this" (test.re[1,0+12]..[1,0+12])
          []
]
```

Strange thing is that this does work if we wrap it another correct module.

SanderSpies:
Btw. in some cases we do get a rather crappy AST. For example:
```
module Y = {
  module Y = {
    let a b => b + 1;
    let a b => b + 1;
  };
  module Z = {
   module M = {
   };
   let o = switch x {
   };
  };
};
```
will error on the last line. This is also currently the case. 

The actual important error however is the wrongly formatted switch statement of which we actually do get an error, but the AST swallows this error. So there is probably room for improvement here.

akotulski:
If this is what I think it is, thank you for doing this!
I've been seeing useless compilation errors in Atom, while ones from command line are much better. I'd love to have them in editor as well - I'm not super familiar with reason syntax and I make bunch of syntax errors when trying to use it.

src/reason_toolchain.ml
SanderSpies: Might have a fix for this.

SanderSpies:
Asked @yunxing for a review of this PR. 

I think it's an improvement upon the current situation even if menhir error messages sometimes get overwritten by typed errors. This is also the case in the current situation btw.

yunxing:
This is very smart. You deserve a medal!

yunxing:
@SanderSpies What are the remaining issues? Do you have a repro that we can create issue with?

SanderSpies:
@yunxing The main example is:
```
module foo = {
};
```

this results in the following AST on our part:

```
[
  structure_item (test.re[1,0+7]..[1,0+10])
    Pstr_extension "merlin.syntax-error"
    [
      structure_item (_none_[1,0+-1]..[1,0+-1]) ghost
        Pstr_eval
        expression (_none_[1,0+-1]..[1,0+-1]) ghost
          Pexp_constant Const_string("A module's name needs to begin with a upper-case letter\n",None)
    ]
  structure_item (test.re[1,0+7]..[3,16+2])
    Pstr_eval
    expression (test.re[1,0+7]..[3,16+1])
      Pexp_setinstvar "foo" (test.re[1,0+7]..[1,0+10])
      expression (test.re[1,0+13]..[3,16+1])
        Pexp_object
        class_structure
          pattern (test.re[1,0+14]..[1,0+14])
            Ppat_var "this" (test.re[1,0+14]..[1,0+14])
          []
]
```

However Merlin changes this error message into: 
```
Error: unbound instance variable foo
```
which probably comes from here:
https://github.com/ocaml/merlin/blob/master/orig/ocaml_402/typing/typeclass.ml#L1774

PR #985, opened 2017-01-19
I don't know why these were changed in my commit... but fixed

tekknolagi:
These were changed in https://github.com/facebook/reason/commit/a3fa9358a74298ebde7944059a631d3c55af5b2b for not sure what reason


yunxing:
Thanks for the quick fix.

Issue #976, opened 2017-01-17
Put reason version in `refmt -v`


chenglou:
Good for debugging.

yunxing:
On my list!

tekknolagi:
I can tackle this, I think.

chenglou:
Would be nice if doing master `refmt -v` shows the git hash or something, and not `1.4.0`. That was the context for this issue. Asking about someone's refmt installed version.

yunxing:
Should be doable.

tekknolagi:
Oh! That is different. I had already completed the version number. I think that can also be tackled, though.

PR #986, opened 2017-01-19
Add version number to refmt --version output

tekknolagi:
Fix https://github.com/facebook/reason/issues/976

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,13 +76,20 @@ update_error:
 ROOT_DIR:=$(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
 SUBSTS:=$(ROOT_DIR)/pkg/substs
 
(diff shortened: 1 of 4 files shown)


yunxing:
Looks pretty good. Thanks for doing this so quickly! 

@chenglou Now I feel you should get ice cream for Max instead for me.

tekknolagi:
@yunxing there was ice cream involved??

tekknolagi:
Should I have committed `package.ml`?

yunxing:
@tekknolagi Yes, you need to commit package.ml.

yunxing:
I feel it is pointless to wait 10 hours for CI :( We should look into Circle CI. 

Merging this first. If it breaks I believe we can fix them under 10 hours.

Issue #988, opened 2017-01-19
Float formatting shouldn't have parentheses


chenglou:
`2. *. -60.` formats to `2. *. (-60.)`

bsansouci:
@jordwalke Any reason why we wrap unary minus with parens?

chenglou:
@yunxing is fixing this right now afaik

yunxing:
@bsansouci It's a legacy from ocaml https://github.com/ocaml/ocaml/blob/4.02.3/parsing/pprintast.ml#L183

yunxing:
With the current function application syntax, it is actually hard to implement this properly as explained in #989. We can revisit if we move the the new function application syntax.

chenglou:
¯\_(ツ)_/¯

PR #989, opened 2017-01-20
Remove parens for negative int and float literals

yunxing:
This diff removes parens for negative int and float literals by lexing them as one token directly instead of rely on the parser (otherwise it won't be lookahead 1).


yunxing:
@jordwalke @let-def Need some help in review here to make sure I'm not doing anything crazy.

yunxing:
#988

tekknolagi:
How does this work with, say, `myFunc 3 -4`, those being separate arguments? One argument?

yunxing:
@tekknolagi I think you need a myFunc (3 -4) to represent a single expression as argument.

yunxing:
Actually, I think this doesn't work. It makes people confuse about `foo -4` (are you applying `foo` over `-4`, or are you subtracting `foo` by `4`)? 

We can always say `-4` means negative 4 instead of subtracting 4 from something, but I would rather not do that here as it is counterintuitive for people coming from other languages. 

Note that we won't have this issue if we move to the parens function application syntax, where we can say "foo(-4)" to represent applying `foo` over `-4`.

tekknolagi:
I think this is vaguely reminiscent of the C parsing weirdness, where if you have something like this:

```c
foo * myvar;
```

the parser does not know what kind of expression that is unless it knows the kind of symbol `foo` is. Is it a type? Another variable?

PR #991, opened 2017-01-20
Add more docs to reason_parser.messages

yunxing:
People occasionally get this error when changing the parser (this is a good thing, meaning we are reducing possible error space). Document it here so we know how to deal with it.


SanderSpies:
In some cases the developer should change the tokens instead.

yunxing:
@SanderSpies the new token list should be automatically generated, no?

SanderSpies:
Ah, yes you are right.  I've never done that before though - wasn't that really time consuming?

PR #992, opened 2017-01-20
[Reason] Fix removed "module type of" feature.

jordwalke:
Summary:This was originally supported, and must have gotten removed in
the middle of a refactor.

Test Plan:Test case.

Reviewers:chenglou, yunxing

CC:

### DIFF
--- a/src/reason_parser.mly
+++ b/src/reason_parser.mly
@@ -1617,8 +1617,8 @@ nonlocal_module_binding_details:
 non_arrowed_module_type: mark_position_mty(_non_arrowed_module_type) {$1}
 _non_arrowed_module_type:
   | simple_module_type {$1}
-  | MODULE TYPE module_expr %prec below_LBRACKETAT
-      { mkmty(Pmty_typeof $3) }
+  | MODULE TYPE OF module_expr %prec below_LBRACKETAT
+      { mkmty(Pmty_typeof $4) }
   | module_type attribute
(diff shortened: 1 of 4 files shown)


src/reason_parser.mly
```diff
@@ -1617,8 +1617,8 @@ nonlocal_module_binding_details:
 non_arrowed_module_type: mark_position_mty(_non_arrowed_module_type) {$1}
 _non_arrowed_module_type:
   | simple_module_type {$1}
-  | MODULE TYPE module_expr %prec below_LBRACKETAT
-      { mkmty(Pmty_typeof $3) }
+  | MODULE TYPE OF module_expr %prec below_LBRACKETAT
```
yunxing: Should we use 
```
| MODULE TYPE OF? module_expr..
```
to make this optional? 
Is there any code out there is written as "include module type Sting" ?

jordwalke: I'm not sure but I doubt it because even if converting from `.ml`, the printer printed it as `module type of`, which wouldn't have parsed. Likewise, even if someone wrote it in `.re`, the printer would be printing it in a form that wouldn't have parsed.

yunxing: @jordwalke fair point.

chenglou:
yolo

SanderSpies:
Probably my mistake, apologies!

Issue #990, opened 2017-01-20
interface external format strips attributes


chenglou:
Repro:

```
external createClassInternalHack : Js.t 'classSpec => reactClass = "createClass" [@@bs.val] [@@bs.module
                                                                    "React"
                                                                    ];
```

Formats to

```
external createClassInternalHack: Js.t 'classSpec => reactClass = "createClass";
```

**In an interface file**

IwanKaramazow:
Looking into this one after I finished my coffee. (Still waking up)

chenglou:
That was fast, thanks =D

IwanKaramazow:
Pr ready for review.

PR #994, opened 2017-01-20
Fix printing of attributes in interface files

IwanKaramazow:
https://github.com/facebook/reason/issues/990

Printing of attributes in interface files was non-existent.
This pr implements the printing of those attributes.

Example:
```ocaml
let x: int [@@bs.val];

type t 'a;

type reactClass;

type reactElement;

external createClassInternalHack :
  t 'classSpec => reactClass =
  "createClass" [@@bs.val] [@@bs.module "React"];

external map : ('a => 'b) [@bs] => array 'b =
  "" [@@bs.send.pipe : array 'a];

external createClassInternalHack :
  t 'classSpec => reactClass =

  reactClass =>
  t {.. reasonProps : 'props} =>
  array reactElement =>
  reactElement =
  "createElement" [@@bs.val] [@@bs.module "react"] [@@bs.splice];
```

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/attributes.rei
+++ b/formatTest/typeCheckedTests/expected_output/attributes.rei
@@ -1,7 +1,38 @@
 /* Copyright (c) 2015-present, Facebook, Inc. All rights reserved. */
 let test: int;
+
+
 /**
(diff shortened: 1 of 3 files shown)


yunxing:
Looking good.

PR #995, opened 2017-01-20
Use cmdliner (round 2)

tekknolagi:
I have adapted Cmdliner to our current `master` branch. This was previously discussed here: https://github.com/facebook/reason/pull/395

### DIFF
--- a/.merlin
+++ b/.merlin
@@ -1,3 +1,3 @@
-PKG utop merlin_extend compiler-libs easy-format MenhirLib
+PKG utop merlin_extend compiler-libs easy-format MenhirLib str cmdliner
 B _build/src
 S src
--- a/.merlin
+++ b/.merlin
@@ -1,3 +1,3 @@
-PKG utop merlin_extend compiler-libs easy-format MenhirLib str cmdliner
+PKG utop merlin_extend compiler-libs easy-format MenhirLib cmdliner
 B _build/src
 S src
--- a/opam
+++ b/opam
@@ -23,6 +23,8 @@ build-test: [
 depends: [
   "easy-format"  {>= "1.2.0"}
   "ocamlfind"    {build}
+  "ounit"        {test}
+  "cmdliner"     {>= "0.9.8"}
   "utop"         {>= "1.17"}
   "menhir"       {>= "20160303"}
   "merlin-extend"{>= "0.3"}
--- a/opam
+++ b/opam
@@ -1,7 +1,7 @@
 
 opam-version: "1.2"
 name: "reason"
-version: "1.4.0"
+version: "1.5.0"
 maintainer: "Jordan Walke <jordojw@gmail.com>"
 authors: [ "Jordan Walke <jordojw@gmail.com>" ]
 license: "BSD"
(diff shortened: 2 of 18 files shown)


tekknolagi:
If somebody has suggestions for an alternative to wrapping the entirety of `refmt` in a `try`, please let me know! :)

src/refmt_impl.ml
```diff
@@ -6,167 +6,138 @@ open Reason_implementation_printer
...
+      thePrinter ast;
+      (* Also closes all open boxes. *)
+      Format.pp_print_flush output_formatter ();
+      flush output_chan;
+      Printer_maker.close_output_file output_file output_chan;
+      `Ok ()
+  )
+
+(* TODO: gross solution. fix. *)
+let refmt_wrapper a b c d e f g h i j k l =
```
yunxing: I don't find a good way to fix this. Maybe others have some ideas?

yunxing:
@tekknolagi Thank you for picking this up. Diff looks very good, especially for the use of the API. Let's rebase against master, where BetterErrors is not needed anymore.

src/refmt_impl.ml
tekknolagi: I think this is not working at the moment. I will investigate today.

tekknolagi:
@yunxing Sounds great. I want to make the behavior as close to possible as it currently is first.

src/refmt_impl.ml
tekknolagi: Fixed. Still gross.

tekknolagi:
Rebased. Tests pass.

opam
```diff
@@ -23,6 +23,8 @@ build-test: [
 depends: [
   "easy-format"  {>= "1.2.0"}
   "ocamlfind"    {build}
+  "ounit"        {test}
```
yunxing: do we use this?

opam
```diff
@@ -23,6 +23,8 @@ build-test: [
...
+  "cmdliner"     {>= "0.9.8"}
```
yunxing: please add this in package.json.in and opam.in and make a new version (dryrun)

opam
chenglou: nooooooo

tekknolagi: Done.

opam
tekknolagi: Done.

.merlin
```diff
@@ -1,3 +1,3 @@
-PKG utop merlin_extend compiler-libs easy-format MenhirLib
+PKG utop merlin_extend compiler-libs easy-format MenhirLib str cmdliner
```
SanderSpies: Are we re-introducing `str` here? 

Related: https://github.com/facebook/reason/pull/659

tekknolagi: My bad. Was blindly including some things from the old Cmdliner PR.

src/cmdliner_shim.ml
```diff
@@ -0,0 +1,40 @@
+open Cmdliner
```
yunxing: This can definitely be in a type theory problem set :D

yunxing:
Looks great, thanks for doing this!

src/cmdliner_shim.ml
tekknolagi: I'll propose it to my PL prof

Issue #997, opened 2017-01-21
Extension points should line break and not be on the same line when it's too long


chenglou:
Repro:

```
let isArrayPolyfill: (Obj.t => Js.boolean) [@bs] = [%bs.raw "function(a) {return Object.prototype.toString.call(a) === '[object Array]'}"];
```

This is the new behavior after some attribute line break changes I think. Extensions should be legit expressions here.

IwanKaramazow:
This isn't just extensions though: 

```
let x = "some super long string some super long stringsome super long stringsome super longstringsome super long string some super long string some super long string";
```

We can't break strings over multiple lines? Or what do you wanna see in this case?

chenglou:
Before it was

```
let isArrayPolyfill: (Obj.t => Js.boolean) [@bs] = [%bs.raw 
  "function(a) {return Object.prototype.toString.call(a) === '[object Array]'}
"];
```

extension points aren't quite like attributes so maybe they should format like a function application or something.

IwanKaramazow:
_Opens MacVim, starts debugging_

chenglou:
cc @jordwalke to assess what I just said.

IwanKaramazow:
No need for anything crazy, surgical insertion of `~break:IfNeed` in the right places gives us the best of both worlds. PR ready for review.

chenglou:
Done! Thanks.

PR #999, opened 2017-01-21
Break payloads of extension nodes IfNeed

IwanKaramazow:
https://github.com/facebook/reason/issues/997

It was just a matter of `~break:IfNeed` in the right places to get everything printed out correctly.
Had to introduce a small change on how the following is printed in the tests:
```
  reactClass =>
  Js.t {.. reasonProps : 'props} =>
  array reactElement =>
  reactElement =
  "createElement"
  [@@bs.val] [@@bs.module "react"] [@@bs.splice];
```
` [@@bs.val] [@@bs.module "react"] [@@bs.splice]` now breaks on a seperate line when necessary. In reality this will appear after `"createElement"` because the tests have a pretty short print-width of `50`

Example Rehydrate:
```
let isArrayPolyfill: (Obj.t => Js.boolean) [@bs] = [%bs.raw
  "function(a) {return Object.prototype.toString.call(a) === '[object Array]'}"
];

external createClassInternalHack : Js.t 'classSpec => reactClass =
```

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/attributes.re
+++ b/formatTest/typeCheckedTests/expected_output/attributes.re
@@ -383,7 +383,8 @@ external createCompositeElementInternalHack :
   Js.t {.. reasonProps : 'props} =>
   array reactElement =>
   reactElement =
-  "createElement" [@@bs.val] [@@bs.module "react"] [@@bs.splice];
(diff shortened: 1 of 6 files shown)


chenglou:
m-m-merrrged!

Issue #1002, opened 2017-01-24
Merlin outcome printer still prints object syntax as <..>


chenglou:
Repro:

```reason
type foo = {. bar: int};
```

<img width="229" alt="screenshot 2017-01-24 01 15 06" src="https://cloud.githubusercontent.com/assets/1909539/22241047/91cf0886-e1d2-11e6-8fbb-1176a5a28a37.png">

SanderSpies:
I have a fix for this in a few minutes.

SanderSpies:
In general `reason_oprint.ml` seems to need a lot of love.

chenglou:
Works, awesome!

PR #1003, opened 2017-01-24
Print objects as {..} with outcome printer #1002.

SanderSpies:
Also some other minor fixes regarding `pub`, `pri` and `as`.


yunxing:
Looks great.

PR #1006, opened 2017-01-25
Remove --output flag

tekknolagi:
Following my discussion with @nojb on https://github.com/dbuenzli/cmdliner/pull/17 and some thought, I think that introducing `--output` was a bad idea:

1. Shell redirection exists for a reason. It is trivial to write: `refmt --parse ml --print re file.ml > file.re`.
2. I introduced `--in-place` because writing to the file that we read from is a special case, and requires some shell shenanigans to work (see https://github.com/facebook/reason/issues/886). But at the same time introduced `--output`, which is wholly unnecessary and also brings with it some annoying command-line logic (mutual exclusion of `--output`, `--in-place`, etc).

I have removed `--output` in this PR ~that (at the moment) builds upon https://github.com/facebook/reason/pull/995~.

/cc @chenglou @yunxing

### DIFF
--- a/src/refmt_args.ml
+++ b/src/refmt_args.ml
@@ -52,11 +52,6 @@ let heuristics_file =
   in
   Arg.(value & opt (some file) None & info ["h"; "heuristics-file"] ~doc)
 
-let output =
-  let docv = "FILENAME" in
-  let doc = "target file for output; default [stdout]" in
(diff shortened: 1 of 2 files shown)


yunxing:
I think your diff contains content from the other PR?

tekknolagi:
@yunxing fixed.

yunxing:
Looks good. Thanks.

PR #1007, opened 2017-01-26
Make --print default to Reason

tekknolagi:
Per @chenglou's request, `refmt` should default to printing Reason code, not binary. This is a saner default for people trying out the `refmt`.

/cc @yunxing

### DIFF
--- a/miscTests/jsxPpxTest.sh
+++ b/miscTests/jsxPpxTest.sh
@@ -9,7 +9,7 @@ do
 
   expected=`cat $testPath/expected$i.re`
 
-  ocamlc -dsource -ppx ./reactjs_jsx_ppx.native -pp ./refmt_impl.native -impl $test \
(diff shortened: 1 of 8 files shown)


yunxing:
Looks nice. Thanks! 

I like how cmdliner makes this much easier.

tekknolagi:
Yep! Looking into the failing tests now.

miscTests/jsxPpxTest.sh
```diff
@@ -9,7 +9,7 @@ do
 
   expected=`cat $testPath/expected$i.re`
 
-  ocamlc -dsource -ppx ./reactjs_jsx_ppx.native -pp ./refmt_impl.native -impl $test \
+  ocamlc -dsource -ppx ./reactjs_jsx_ppx.native -pp "./refmt_impl.native --print binary" -impl $test \
```
yunxing: I think you need to do the same for some rules in rebuild as well.

yunxing:
This diff looks good to me. 

@jordwalke Do you mind help to also review this?

PR #1010, opened 2017-01-26
Vendor cmdliner in

yunxing:
RIP package management


yunxing:
@bobzhang ^

chenglou:
We'll do whatever that makes @bobzhang's bspack workflow a bit easier. But for the sake of sharing cached artifacts in the future we should probably pull these out again or something...

bobzhang:
bspack is done in dev time, git submodules will be good enough

chenglou:
Is it possible for bspack to work on npm installed cmdliner?

PR #1011, opened 2017-01-26
Add tmfer

tekknolagi:
This is a small command that allows formatting of OCaml->OCaml.


chenglou:
Great! But I think refmt --parse ml --print ml would work if we make ocaml comment interleaving better!

PR #1012, opened 2017-01-27
Inline the attributes list in primitive_declaration

### LINKED ISSUE #1009: refmt oddity with externals and ppx strings
```reason
external foo : bool => bool = "" [@@bs.module "Bar"] [@@ocaml.deprecated "Use bar instead. It's a much cooler function. This string needs to be a little long"];
```
refmts to:
```reason
external foo : bool => bool =
  ""

    [@@bs.module "Bar"]
    [@@ocaml.deprecated
      "Use bar instead. It's a much cooler function. This string needs to be a little long"
    ];
```
which has two issues in my mind:
1. The blank line between the empty string and the ppx should not be there.
2. Both ppx are indented two levels. They should be indented just one level.

yunxing:
Fix #1009

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/attributes.re
+++ b/formatTest/typeCheckedTests/expected_output/attributes.re
@@ -388,3 +388,10 @@ external createCompositeElementInternalHack :
 
 external add_nat : int => int => int =
   "add_nat_bytecode" "add_nat_native";
+
+external foo : bool => bool =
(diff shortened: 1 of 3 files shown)


yunxing:
@rickyvetter ^^

jordwalke:
Wow is something blocking CI? Amazingly long wait times

PR #1025, opened 2017-01-30
Remove `of` keyword in oprint.

SanderSpies:
This ensures `of` is not displayed within editors that use Merlin.

Eg. `type foo = | Foo of int` now correctly becomes `type foo = | Foo int`.


yunxing:
I was just thinking about it this morning. 

Thank you!

PR #1030, opened 2017-01-31
Properly build toploop

yunxing:
Easy_format is not a dependency anymore. This fixes master.


yunxing:
Merging this to fix master :( I will set up circleCI soon so that CI can be faster.

PR #1033, opened 2017-01-31
Generate all the possible error states for Menhir. 

SanderSpies:
Handy for when an error state is not present in the current `reason_parser.messages` file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -63,6 +63,12 @@ clean: clean-tests
 compile_error: update_error
 	menhir --explain --strict --unused-tokens src/reason_parser.mly --compile-errors src/reason_parser.messages > src/reason_parser_message.ml
 
+all_errors:
+	@ echo "Regenerate all the possible error states for Menhir."
+	@ echo "Warning: This will take a while and use a lot of CPU and memory."
+	@ echo "---"
+	menhir --explain --strict --unused-tokens src/reason_parser.mly --list-errors > src/reason_parser.all.messages
+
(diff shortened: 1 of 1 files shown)


yunxing:
hmm, how do we use "src/reason_parser.all.messages"?

SanderSpies:
Perhaps there's a better solution, but the idea is that you can lookup error states that are not present in the current reason_parser.messages.

yunxing:
@SanderSpies, which use case are you thinking about? 

My understanding is that sometimes we may run into an error state that is not in reason_parser.messages and we would get a runtime error. We can't update the existing reason_parser.messages database to automatically include that error state due to some bugs in Menhir. That's why in this diff you want to regenerate all possible error states so you can look it up. But what do you do with it? Do you copy the entire state over from "reason_parser.all.messages" to "reason_parser.messages"?

SanderSpies:
For example when working on https://github.com/facebook/reason/pull/1018 there was an error code, but I could not find the error state within the .messages file. This could also happen to others and I want to have a relatively easy way to handle this. The idea is to also add this I want to the new wiki page.

SanderSpies:
So not copy all the states, but just the specific one you need.

yunxing:
@SanderSpies I see, thanks for explanation.

SanderSpies:
Btw. it might be an idea to just add the file to the repo to lessen the
friction.

chenglou:
Good stuff! I'll add the lines to wiki

PR #1036, opened 2017-02-02
Experimental metaquot test case

SanderSpies:
cc @chenglou


chenglou:
Not sure if we're exposing this correctly; merging this anyways. Let's see if we can use it soon!

PR #1041, opened 2017-02-06
Automatically generate show functions for user-defined types

tekknolagi:
Currently this only works when converting from Reason code to something else. Additionally, it only works on type declarations not composed with other declarations with `and`. Last, it only works on types defined in the file that is passed to `refmt`.

I vendored in ppx_deriving, slimmed down slightly to accommodate our need for only the `show` plugin.

TODO:
- [x] Cleanly vendor ppx_deriving
  - [x] Fix dependency issue with ppx_tools.metaquot
- [x] Add a command-line flag to enable; opt-in
- [x] ~Add a command-line option for location of the runtime (required)~ Prepend the runtime to the AST
- [x] Support recursive (compound?) types
- [x] Make types `nonrec`
- [x] Determine if lack of idempotence is an issue
- [x] Make tests


tekknolagi:
@yunxing what is the best way to bundle in a file so that it can be spit out at an arbitrary location of the client's choice?

vendor/ppx_deriving/ppx_deriving_show.ml
```diff
@@ -0,0 +1,5087 @@
+(*  This file is part of the ppx_tools package.  It is released  *)
```
yunxing: In the future, it would be nice if we can have a script to generate this file from ppx version for Reason developers. But I'm not sure how to do that gracefully yet.

src/reason_config.ml
```diff
@@ -3,7 +3,9 @@
  *)
 
 let recoverable = ref false
+let add_printers = ref false
```
yunxing: In the long run, once the feature is stable, we should probably turn on the default behavior here. 

We will need create an option in build systems to turn on this feature (so that people using rebuild can have a way to try this out). Let's file an issue once this is merged.

_tags
```diff
@@ -4,8 +4,9 @@ true: warn(@5@8@10@11@12@14@23-24@26@29@40), bin_annot, safe_string, debug
 <node_modules/**>: -traverse
 "formatTest": -traverse
 "src": include
-<src/*>: package(menhirLib unix compiler-libs.common ocamlbuild findlib merlin_extend)
+<src/*>: package(menhirLib unix compiler-libs.common ocamlbuild findlib merlin_extend ppx_tools.metaquot result dynlink findlib.dynload)
```
SanderSpies: I explicitly removed this from my ppx PR per @chenglou's request.

tekknolagi: As far as I can tell right now, yes. Looking into how to remove it.

tekknolagi:
@yunxing ppx runtime is now auto-generated in AST form and conditionally prepended to the beginning of the file.

vendor/ppx_deriving/ppx_deriving_show.ml
tekknolagi: I can write a script in the future but it's messy. An issue for post PR?

yunxing:
Discussed offline -- This diff adds a new option 'add_printers', which allows the printer to generate printers for each type and necessary runtime at the top of the file. The "add_printers" should mostly be used with binary mode, where we feed the data into OCaml compiler.

vendor/ppx_deriving/ppx_deriving_runtime.ml
```diff
@@ -0,0 +1,40 @@
+type nonrec int = int
```
yunxing: Nice use of nonrec here, thanks @hcarty!

_tags
yunxing: We are looking into how to remove ppx_tools.metaquot.

@chenglou Could you help make sure the rest of the new dependencies won't break on your side?

src/refmt_args.ml
```diff
@@ -56,6 +56,10 @@ let in_place =
   let doc = "reformat a file in-place" in
   Arg.(value & flag & info ["in-place"] ~doc)
 
+let add_printers =
+  let doc = "add auto-printers to user-defined types" in
+  Arg.(value & flag & info ["a"; "add-printers"] ~doc)
```
yunxing: "-a" option is probably not needed for daily interactive use,  maybe we can remove it.

tekknolagi: Good point.

_tags
hcarty: ppx_tools.metaquot does help a lot with keeping the code concise and compatible across OCaml AST changes.  That said, once the ocaml-migrate-parsetree work is merged that library can be used to smooth over AST side.

tekknolagi:
@yunxing vendored `Ast_convenience`

yunxing:
@tekknolagi Let's write a least a quick example in the test folder?

tekknolagi:
@yunxing done! Though I made a new directory since it wouldn't fit well in `typedTests` -- needs its own CLI flag.

_tags
```diff
@@ -4,8 +4,9 @@ true: warn(@5@8@10@11@12@14@23-24@26@29@40), bin_annot, safe_string, debug
 <node_modules/**>: -traverse
 "formatTest": -traverse
 "src": include
-<src/*>: package(menhirLib unix compiler-libs.common ocamlbuild findlib merlin_extend)
+<src/*>: package(menhirLib unix compiler-libs.common ocamlbuild findlib merlin_extend result dynlink findlib.dynload)
```
yunxing: Maybe add 'result' to package.json.in and opam.in file, and run a make pre_release.

tekknolagi: Good call.

tekknolagi:
Licenses seem compatible (`Ast_convenience` being MIT and included in a BSD 2.0 project)

yunxing:
Looks very concise and straightforward! While waiting for the CI, Let me patch this locally to see if that works as expected.

yunxing:
@tekknolagi Updated the PR with some fixes.

bobzhang:
Can you use `[%show: my_type]` instead of `show_my_type`, the latter exposed the name mangling to users which will be very hard to remove it in the future.

jordwalke:
Curious why it doesn't work with "and" type definitions.

tekknolagi:
@jordwalke it should!

tekknolagi:
@bobzhang Given that we're adding and changing syntax left and right, I think this is perhaps one of our lesser worries; it is currently an opt-in experimental feature. If we expect code written to target this compiler version to always work in the future, then we should be making many other decisions differently.

vendor/ppx_deriving/ppx_deriving_show.ml
tekknolagi: I think @let-def added some steps to generate this file in his PR.

let-def: There: https://github.com/let-def/reason/tree/404-rebased/vendor/ppx_deriving
I tried to follow tips from @tekknolagi  :)

Issue #1046, opened 2017-02-08
refmt to merlin.syntax-error


rickyvetter:
Currently refmting a file with something like:

```reason
{
  pub render () => {
    let state: {foo:Bar.t};
  }
};
```

turns the entire file into
```reason
[%%merlin.syntax-error "Record type is not allowed"];
```

This is a pretty aggressive way of telling me I'm wrong :(

chenglou:
cc @SanderSpies

SanderSpies:
Fixed by ^

chenglou:
Dat response time

PR #1049, opened 2017-02-09
Fix the --recoverable flag.

SanderSpies:
Should help with: https://github.com/facebook/reason/issues/1046


jordwalke:
Thanks @SanderSpies !

Issue #1051, opened 2017-02-10
`let f => x` implicitly inserts a `()`


chenglou:
It gets re-formatted to `let f () => x`. I feel like there's not much value in having this particular shortcut (that gets refmt-ed away anyway); newcomers are sometimes confused by `=` and `=>` already.

PR #1055, opened 2017-02-11
Remove "let f => x" support.

SanderSpies:
#1051


kennetpostigo:
@SanderSpies I've used this everywhere in my guide. What should I use instead?

jordwalke:
@kennetpostigo Are you sure you used this exact form? Notice that there is no argument (and I'm just now finding out that it is parsed as a single `()` argument). I had no idea this existed!

kennetpostigo:
Sorry, I misinterpreted the PR. This has to do with `unit` being returned implicitly right?

yawaramin:
@kennetpostigo it does have to do with function expression syntax--not `unit` being returned implicitly. Specifically,

```ocaml
let f => x;
```

is being interpreted as

```ocaml
let f () => x;
```

This is tripping people up.

kennetpostigo:
@yawaramin Interesting, thank you for explaining!

src/reason_parser.messages
```diff
@@ -1,4 +1,4 @@
-# ------------------------------------------------------------------------------
+©# ------------------------------------------------------------------------------
```
chenglou: @SanderSpies lol?

chenglou:
Now passes. Merging! Thanks.

src/reason_parser.messages
SanderSpies: Yikes.

PR #1056, opened 2017-02-12
Clarify Reason record literal syntax

yawaramin:
Also warn about code example tables in readme.

### DIFF
--- a/docs/javaScriptCompared.html
+++ b/docs/javaScriptCompared.html
@@ -322,9 +322,91 @@ functions.
 
 ### Record Fields
 
-In `Reason`, type definitions for records must be "in scope" if you want to create
-records or access record fields. This means you have to `open` the module in which
-that record's type is defined in.
+In `Reason`, you must ensure your record literal has an unambiguous type if you
+want to create records or access record fields. In the simplest case, the record
+type is already in scope and you can unambiguously create values just by using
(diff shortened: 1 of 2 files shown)


chenglou:
Hummmm I'm not sure this is the best intro example for record. It's a bit up-front about a case that newcomers won't encounter just yet. Maybe move it after the description and do "If the record you're trying to access is somewhere else, then ..."?

yawaramin:
@chenglou I see what you mean. Let me try out your suggestion.

hcarty:
The third example (`{ Module.field1; field2; ...}`) is probably worth highlighting first as it is the safest example.  The other two perform a local open of the parent module which affects the environment beyond exposing the record field names.

docs/javaScriptCompared.html
```diff
@@ -322,9 +322,91 @@
...
+  /&ast;
+  Alternatively, using field name punning,
+  let make id name => { id, name };
+  &ast;/
+};</pre>
+    </td>
+  </tr>
+</table>
+
+When the record type is declared in a different modue, you have to give Reason a
```
Schmavery: modue

yawaramin:
Bump :-)

chenglou:
Looks good now. Thanks!

PR #1059, opened 2017-02-14
Move ocaml into a hard dep

chenglou:
This makes it so that packages requiring reason doesn't need to add ocaml as their dep themselves, each time.


yunxing:
Looks good.

PR #1067, opened 2017-02-15
Add check to confirm release

tekknolagi:
Following a number of accidental releases that I have seen and executed in my short stay as an intern on the Reason team, I would like to propose a pre-release check mechanism. Idea: you are much less likely to publish to npm accidentally if you are required to type a pseudo-random string to confirm.

### DIFF
--- a/release-check.sh
+++ b/release-check.sh
@@ -1,12 +1,22 @@
 #!/usr/bin/env bash
 
-RANDSTR=`head -c2 </dev/urandom|xxd -p -u`
+die () { echo "${1}"; exit 1; }
+
+# Check that the current HEAD is on origin/master.
+HEAD=`git rev-parse --verify HEAD`
+MASTERURL="https://api.github.com/repos/facebook/reason/git/refs/heads/master"
+HEADMSG="Current HEAD is not on upstream master. This is a requirement before releasing.
+If you are sure it is, try switching branches to master and pulling changes."
+echo "Checking HEAD against upstream master..."
+curl --silent $MASTERURL | grep "sha" | grep $HEAD || die "$HEADMSG"
+
+# Confirm that the user actually means to release.
+RANDSTR=`head -c2 </dev/urandom | xxd -plain -u` # -u is uppercase
 echo "Do you want to release? This *WILL* publish to GitHub AND npm!!!"
 read -p "If so, please type '${RANDSTR}' (no quotes): " inp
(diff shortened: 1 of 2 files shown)


chenglou:
lol, I only use https://github.com/sindresorhus/np nowadays

tekknolagi:
Then why do we have `make release`?

chenglou:
I meant for my own repos. Try it!

tekknolagi:
I don't publish things to npm 🙃

tekknolagi:
Derp, just kidding. Apparently I have three packages.

release-check.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
yunxing: maybe use #!/usr/bin/env bash (http://stackoverflow.com/questions/16365130/the-difference-between-usr-bin-env-bash-and-usr-bin-bash)

tekknolagi:
@yunxing Added a check!

tekknolagi:
Unclear of the preferred order of these checks.

release-check.sh
```diff
@@ -0,0 +1,22 @@
...
+
+die () { echo "${1}"; exit 1; }
+
+# Check that the current HEAD is on origin/master.
+HEAD=`git rev-parse --verify HEAD`
+MASTERURL="https://api.github.com/repos/facebook/reason/git/refs/heads/master"
+HEADMSG="Current HEAD is not on upstream master. This is a requirement before releasing.
+If you are sure it is, try switching branches to master and pulling changes."
+echo "Checking HEAD against upstream master..."
+curl --silent $MASTERURL | grep "sha" | grep $HEAD || die "$HEADMSG"
```
yunxing: Does this only work if the HEAD is exactly same as the HEAD of master?

yunxing: Do you think we should use this api instead? :https://developer.github.com/v3/repos/commits/#get-a-single-commit 

Looks like this API tells you if one commit is currently on master or not.

tekknolagi: It looks like our current release workflow makes a version on top of master and pushes that. If we want to support arbitrary commits for a release, then we would have to change the Makefile as well. Perhaps a future project.

yunxing:
@tekknolagi Looks good then. Thanks! Let's merge this now since CI won't test this anyway.

PR #1075, opened 2017-02-18
Use correct dependency for bytecode reason

yunxing:
This ensures correct dependencies are linked for reason.cma


yunxing:
Fastest CI run I've ever had in a while.

PR #1076, opened 2017-02-18
Remove compilerlibs from bytecode version

yunxing:
This is needed to avoid double linking with compilerlibs


yunxing:
Let's merge since CI doesn't actually test this logic currently.

PR #1080, opened 2017-02-19
Fixes #1073

### LINKED ISSUE #1073: Better error for JSX without spaces
From discord:

> So the first line compiles, but the second does not:
> `<li> <a>(ReactRe.stringToElement item)</a></li>`
> `<li><a>(ReactRe.stringToElement item)</a></li>`

Can we warn this better?

chenglou:
This error code seems very low. Is this the correct thing to do?

### DIFF
--- a/src/reason_parser.messages
+++ b/src/reason_parser.messages
@@ -13375,6 +13375,22 @@ Expecting one of the following:
   - "(" + expression + ")"
   - "{" + expression + "}"
 
+implementation: LESSIDENT WITH
+##
+## Ends in an error in state: 19.
+##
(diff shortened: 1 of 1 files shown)


SanderSpies:
Can you check if this error also occurs when not correctly ending a JSX tag?

chenglou:
@SanderSpies you mean like `<li> <foo /></li>`? That one parses.

SanderSpies:
More like `let a = <Foo bar;`.

chenglou:
Yeah that triggers this error too. Hummm...

PR #1083, opened 2017-02-21
Enable --add-printers in reasonbuild

tekknolagi:
This will create printers for user-defined types by default.

### DIFF
--- a/.merlin
+++ b/.merlin
@@ -1,4 +1,4 @@
-PKG utop merlin_extend compiler-libs ocamlbuild easy-format MenhirLib cmdliner
+PKG utop merlin_extend compiler-libs easy-format MenhirLib cmdliner ocamlbuild
 B _build/src
 B _build
 # ^ for package.ml -- avoids errors in Merlin
--- a/.merlin
+++ b/.merlin
@@ -1,4 +1,4 @@
-PKG utop merlin_extend compiler-libs easy-format MenhirLib cmdliner
+PKG utop merlin_extend compiler-libs ocamlbuild easy-format MenhirLib cmdliner
 B _build/src
 B _build
 # ^ for package.ml -- avoids errors in Merlin
--- a/src/reasonbuild.ml
+++ b/src/reasonbuild.ml
@@ -5,7 +5,6 @@ let ext_obj = !Options.ext_obj;;
 let x_o = "%"-.-ext_obj;;
 
 let refmt = "refmt --print binary"
-let refmt_printers = refmt ^ " --add-printers"
 let add_printers_tag = "reason_add_printers"
 
 let ocamldep_command' tags =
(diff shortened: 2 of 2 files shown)


yunxing:
@tekknolagi Do you think we should make this as a default behavior, or do you think we should make this as an option and make it default once we think this is stable enough?

tekknolagi:
@yunxing Fair point. What's the best way to make this an option?

yunxing:
@tekknolagi We don't control ocamlbuild so I don't think we can add a command line option to it. Maybe here we can use a environment variable? 

This will be temporary for people to try out before we switch this into a default behavior so I'm not too worried about it.

jordwalke:
Would it be reasonable to have an extra entrypiont `rebuildwithprint` while we test it out?

yunxing:
@jordwalke Implementation-wise, reusing the same command would be much easier.

tekknolagi:
Environment variable is kind of gross but if it's *definitely* only temporary then I can do that.

hcarty:
Since this is based on ocamlbuild using a tag might make sense.  That would make it easy for users to apply the flag per-file.

Also, if there's a generic feature here which is generally useful in ocamlbuild Gabriel Scherer is usually very open to PRs for the project.

tekknolagi:
@hcarty What do you mean exactly?

hcarty:
ocamlbuild uses tags, either specified on the command line, in a `_tags` file or `myocamlbuild.ml`, to specify most compilation flags.  It's possible (and somewhat common) to specify custom tags.

So a user could have a tags file which says:
```
true: add_type_printers
```
to add `show_*` functions for all of their types in all of their files, for example.  Or:
```
src/lib.re: add_type_printers
```
to only add printers for the types in `src/lib.re`.

Some documentation is available: https://github.com/ocaml/ocamlbuild/blob/master/manual/manual.adoc#tags-and-the-code-_tags-code-file and https://github.com/ocaml/ocamlbuild/blob/master/manual/manual.adoc#flag-declarations - the rest of the ocamlbuild manual may be helpful too.

For examples of tag implementations you can see https://github.com/ygrek/mybuild and https://github.com/hcarty/ocamlbuild-plugins.  https://github.com/hcarty/ocamlbuild-plugins/tree/master/tests/static is a relatively simple example of defining and using a new tag.

tekknolagi:
Looks like right now the easiest non-PR way to do this would be to add the tag `pp(refmt --print binary --add-printers)` to files they want to add printers to.

tekknolagi:
@hcarty @yunxing @jordwalke if anybody has feedback on how we might approach this with parametrized flags (ie `reason(add_printers ...)`), please let me know.

tekknolagi:
@yunxing I think we can just ship as-is for now.

yunxing:
This is good to merge. I don't think the CI is currently testing rebuild.

Issue #1099, opened 2017-02-27
Overeager JSX punning


rickyvetter:
```reason
<Bar foo=Baz.foo />
```

refmt =>

```reason
<Bar foo />
```

Should be no change.

Seems very related to #962

IwanKaramazow:
I'm debugging currently.

IwanKaramazow:
PR has been made 🙌
Actual bugfix took me 3min, installing Reason on opam 40 min. 😂

chenglou:
Fixed.

PR #1102, opened 2017-02-28
Fix overeager JSX punning #1099

IwanKaramazow:
Fix for https://github.com/facebook/reason/issues/1099

A long identifier with a dot shouldn't be punned in jsx because it changes the semantic meaning of that identifier.

### DIFF
--- a/src/reason_pprint_ast.ml
+++ b/src/reason_pprint_ast.ml
@@ -3672,7 +3672,8 @@ class printer  ()= object(self:'self)
       | (lbl, expression) :: tail ->
          let nextAttr =
            match expression.pexp_desc with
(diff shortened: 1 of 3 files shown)


src/reason_pprint_ast.ml
```diff
@@ -3672,7 +3672,8 @@ class printer  ()= object(self:'self)
       | (lbl, expression) :: tail ->
          let nextAttr =
            match expression.pexp_desc with
-           | Pexp_ident (ident) when (Longident.last ident.txt) = lbl -> atom lbl
+           | Pexp_ident (ident) when ((isLongIdentWithDot ident.txt) == false
```
chenglou: `not`?

chenglou: Eh, who cares =P

PR #1103, opened 2017-02-28
Make vendored reason in BS work with windows

chenglou:
Credits to @bobzhang


yunxing:
Thanks!

What do you mean by "things" ? This doesn't have support for any of our .sh scripts, right?

chenglou:
Corrected the title. The vendoring of Reason in BS cuts out the scripts we usually need

yunxing:
@chenglou Thanks, diff looks good to me.

PR #1104, opened 2017-03-01
Document vendoring of Reason though bspack

chenglou:


chenglou:
"Great code chenglou, let's merge it"

Issue #1101, opened 2017-02-28
Support for private extensible variants?


rickyvetter:
Right now it doesn't look like private constructors are respected in reason.

Original OCaml:
```ocaml
type foo += private C of int
```
Actual:
```reason
type foo +=
  | C int;
```

Expected:
```reason
type foo +=
  | private C int; /* Error: Unbound value pri */
```

chenglou:
Hot damn what are you trying to do

rickyvetter:
@chenglou don't worry about it 🕵🏻

jordwalke:
Great first task and high pri because it's regarding completeness!

IwanKaramazow:
Should be familiar terrain for me, will look into it

SanderSpies:
@IwanKaramazow 💪

IwanKaramazow:
Ok, this was already implemented.

```reason
type water = ..;
type molecules = string;
type water += pri Ocean molecules;
```

Reason uses `pri` instead of `private`, this probably originates from @SanderSpies's new object syntax. Do we just document this? Or should I make it `private` in this case?

rickyvetter:
Ahh interesting - when I refmt that code I get:

```reason
type water = ..
type molecules = string
type water +=
  | Ocean of molecules
```

So maybe it's just an issue with refmt not handling it? It doesn't work in any direction ml <=> re or self formatting.

IwanKaramazow:
So it parses, but prints incorrectly. I'll fix it.

SanderSpies:
In general it would be great if we could try to convert more OCaml libraries to Reason and see if it works.

jordwalke:
Might want to do this on top of Fred's diff.

IwanKaramazow:
Opened a PR that should be compatible with Fred's diff, cherry-picking the commit should suffice?

PR #1105, opened 2017-03-01
Fix printing of private extensible variants

IwanKaramazow:
Fixes https://github.com/facebook/reason/issues/1101

Update: due to consistency with normal variants, all constructors will always break recursively, regardless of the amount of constructors & line length.
(https://github.com/facebook/reason/issues/1106)

Example:
```reason
type water = ..;

type water +=
  pri
  | Ocean;

type water +=
  pri
  | MineralWater
  | SpringWater
  | TapWater
  | TableWater;
```

### DIFF
--- a/formatTest/unit_tests/expected_output/variants.re
+++ b/formatTest/unit_tests/expected_output/variants.re
@@ -490,9 +490,11 @@ let listPatternMayEvenIncludeAliases x =>
 type attr = ..;
 
 /* `of` is optional */
-type attr += Str string;
+type attr +=
+  | Str string;
 
-type attr += Point int int;
+type attr +=
+  | Point int int;
 
 type attr +=
   | Float float
@@ -510,11 +512,14 @@ module Graph = {
   type node = ..;
 };
 
-type Graph.node += Str = Graph.Str;
+type Graph.node +=
+  | Str = Graph.Str;
 
 type water = ..;
 
-type water += pri Ocean;
+type water +=
+  pri
+  | Ocean;
 
 type water +=
   pri
@@ -523,7 +528,9 @@ type water +=
   | TapWater
   | TableWater;
 
-type Graph.node += pri Node = Expr.Node;
+type Graph.node +=
+  pri
+  | Node = Expr.Node;
 
 type Graph.node +=
   pri
--- a/formatTest/unit_tests/expected_output/variants.re
+++ b/formatTest/unit_tests/expected_output/variants.re
@@ -490,11 +490,9 @@ let listPatternMayEvenIncludeAliases x =>
 type attr = ..;
 
 /* `of` is optional */
-type attr +=
-  | Str string;
+type attr += Str string;
 
-type attr +=
-  | Point int int;
+type attr += Point int int;
 
 type attr +=
   | Float float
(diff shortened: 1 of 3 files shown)


formatTest/typeCheckedTests/expected_output/jsx.re
```diff
@@ -585,3 +585,16 @@ notReallyJSX foo::1 [] bar::2 [@JSX] [@bla];
 <Optional1 required=(Some "hi") /> [@bla];
 
 <Optional1 required=(Some "hi") /> [@bla];
+
+/* Overeager JSX punning #1099 */
```
jordwalke: Just curious, is this intended?

IwanKaramazow: My mistake, accidentally branched from my jsx-punning-fix-branch, that commit has been merged into master already.

src/reason_pprint_ast.ml
```diff
@@ -3672,7 +3677,8 @@ class printer  ()= object(self:'self)
       | (lbl, expression) :: tail ->
          let nextAttr =
            match expression.pexp_desc with
-           | Pexp_ident (ident) when (Longident.last ident.txt) = lbl -> atom lbl
+           | Pexp_ident (ident) when ((isLongIdentWithDot ident.txt) == false
```
yunxing: Curious: what's this needed for?

formatTest/typeCheckedTests/expected_output/jsx.re
IwanKaramazow: Rebased the commit-history.

src/reason_pprint_ast.ml
IwanKaramazow: Hmm, this is an artefact from https://github.com/facebook/reason/pull/1102. I did branch from that PR. Not sure why this is still in here, I thought I rebased 🤔

IwanKaramazow: Rebase came through 😄

formatTest/unit_tests/expected_output/variants.re
```diff
@@ -514,3 +514,25 @@ module Graph = {
...
   | Str = Graph.Str;
+
+type water = ..;
+
+type water +=
+  pri
+  | Ocean;
+
+type water +=
+  pri
```
yunxing: I like how we always break the lines here so that it's more obvious that the pri is used to decorate the entire list of constructors, as opposed to a single constructor.

jordwalke:
Want to rebase this on master now? Thanks for doing this @IwanKaramazow. We really appreciate the fixes!

IwanKaramazow:
Rebased & pushed. Let's pray CI comes through.

formatTest/unit_tests/expected_output/variants.re
jordwalke: I agree. And the leading bar makes it clear that it applies too all not just the first constructor.

jordwalke:
Thank you @IwanKaramazow.
It's the many small improvements like this that create a high quality experience.

jordwalke:
You can see that the 4.02.3 tests are failing on travis, but they are the expected failures. We should create a directory that's only ran for 4.03 or greater, don't you think?

PR #1107, opened 2017-03-02
Vendor result

chenglou:
cc @tekknolagi @bobzhang


tekknolagi:
Why do we include bspack stuff here?

Issue #821, opened 2016-10-20
ppx attributes on polymorphic variants get pretty printed away


Schmavery:
We (@bsansouci @sgrove) were playing with interfacing with webGL yesterday and ran into this.

```
external foo : something => ([`array_buffer [@bs.as 12353] | `array_something_buffer [@bs.as 1231232132]]) => unit = "foo" [@@bs.send];
```

pretty prints to

```
external foo : something => ([`array_buffer | `array_something_buffer ]) => unit = "foo" [@@bs.send];
```

bsansouci:
@bobzhang ^

sgrove:
Hitting this with the BuckleScript example:

```
external readFileSync :
  name:string ->
  ([ `utf8
   | `my_name [@bs.as "ascii"]
   ] [@bs.string]) ->
  string = ""
  [@@bs.module "fs"]
```
 
Goes to:

```
external readFileSync : name::string => [ | `utf8 | `my_name] [@bs.string] => string =
  "" [@@bs.module "fs"];
```

IwanKaramazow:
I'll fix it today.

IwanKaramazow:
@sgrove PR is awaiting review 🏊

wokalski:
This one can be closed (resolved by #1109).
cc @chenglou

chenglou:
Thanks!

PR #1109, opened 2017-03-02
Fix printing of attributes on polymorphic variant leafs

IwanKaramazow:
Fix for https://github.com/facebook/reason/issues/821

example:
```reason
external readFileSync :
  name::string =>
  [ | `utf8 | `my_name [@bs.as "ascii"]]
  [@bs.string] =>
  string =
  "" [@@bs.module "fs"];

external readFileSync2 :
  name::string =>
  [
    | `utf8 [@bs.as "ascii"]
    | `my_name [@bs.as "ascii"]
  ]
  [@bs.string] =>
  string =
  "" [@@bs.module "fs"];
```

### DIFF
--- a/reason-parser/src/reason_pprint_ast.ml
+++ b/reason-parser/src/reason_pprint_ast.ml
@@ -2566,12 +2566,12 @@ class printer  ()= object(self:'self)
                 first::(List.map (formatOneTypeDefStandard (atom "and")) (tlhd::tltl))
               )
 
-  method type_variant_leaf ?attrs_on_leaf:(attrs=[]) ?opt_ampersand:(a=false) ?polymorphic:(p=false) = self#type_variant_leaf1 attrs a p true
-  method type_variant_leaf_nobar ?opt_ampersand:(a=false) ?polymorphic:(p=false) = self#type_variant_leaf1 [] a p false
+  method type_variant_leaf ?opt_ampersand:(a=false) ?polymorphic:(p=false) = self#type_variant_leaf1 a p true
+  method type_variant_leaf_nobar ?opt_ampersand:(a=false) ?polymorphic:(p=false) = self#type_variant_leaf1 a p false
 
   (* TODOATTRIBUTES: Attributes on the entire variant leaf are likely
    * not parsed or printed correctly. *)
-  method type_variant_leaf1 attrs_on_leaf opt_ampersand polymorphic print_bar x =
+  method type_variant_leaf1 opt_ampersand polymorphic print_bar x =
     let {pcd_name; pcd_args; pcd_res; pcd_loc; pcd_attributes} = x in
     let (arityAttrs, docAtrs, stdAttrs, jsxAttrs) = partitionAttributes pcd_attributes in
     let prefix = if polymorphic then "`" else "" in
@@ -2618,10 +2618,9 @@ class printer  ()= object(self:'self)
       | (_::_, None) -> add_bar nameOf (normalize args)
       | (_::_, Some res) -> add_bar nameOf (normalize (args@[res]))
     in
-    let allAttrs = List.concat [stdAttrs; attrs_on_leaf] in
     let everythingWithAttrs =
-      if allAttrs <> [] then
-        formatAttributed everything (self#attributes allAttrs)
+      if stdAttrs <> [] then
+        formatAttributed everything (self#attributes stdAttrs)
       else
         everything
     in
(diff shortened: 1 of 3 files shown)


src/reason_pprint_ast.ml
```diff
@@ -2597,9 +2597,21 @@ class printer  ()= object(self:'self)
         | (_::_, None) -> add_bar nameOf (normalize args)
         | (_::_, Some res) -> add_bar nameOf (normalize (args@[res]))
       in
+      (*
+        Attach attributes sitting on the entire variant leaf.
+          Example:
+            [ | `utf8 | `my_name [@bs.as "ascii"]]
+            type variantType = | Foo int [@onInt]
+        We don't want [@bs.as "ascii"], nor [@onInt] dropped. Because of [@bs.as "ascii"]
+        coming in from the Ast nodes outside of the Parsetree.constructor_declaration, while
```
yunxing: Not sure if the comment is true: I think both of the attributes are inside of contructor_declaration.  The difference is that in the first example, the type is an abstract type and all Rtags are defined in the manifest section of constructor_declaration. In the second example, the attribute [@onInt]  is decorating a variant type, which is also defined in constructor_delcaration.

IwanKaramazow: You're right, I misinterpreted the whole thing. Thanks for pointing this out! Is the overall approach ok here?

IwanKaramazow:
Has been rebased on master.

IwanKaramazow:
@yunxing just rebased this one again on master.
Should I improve something in this PR? Someone ran into this problem this week & mentioned it on Discord.

jordwalke:
Thanks for the rebase onto master! Let's merge this by EOD? Ping me if it's not in by then?

IwanKaramazow:
Ping @jordwalke ☺️

reason-parser/src/reason_pprint_ast.ml
```diff
@@ -2618,9 +2618,10 @@ class printer  ()= object(self:'self)
       | (_::_, None) -> add_bar nameOf (normalize args)
       | (_::_, Some res) -> add_bar nameOf (normalize (args@[res]))
     in
+    let allAttrs = List.concat [stdAttrs; attrs_on_leaf] in
```
jordwalke: Why weren't the attrs_on_leaf part of the standard attributes to begin with?

reason-parser/src/reason_pprint_ast.ml
```diff
@@ -2566,12 +2566,12 @@ class printer  ()= object(self:'self)
...
 
-  method type_variant_leaf ?opt_ampersand:(a=false) ?polymorphic:(p=false) = self#type_variant_leaf1 a p true
-  method type_variant_leaf_nobar ?opt_ampersand:(a=false) ?polymorphic:(p=false) = self#type_variant_leaf1 a p false
+  method type_variant_leaf ?attrs_on_leaf:(attrs=[]) ?opt_ampersand:(a=false) ?polymorphic:(p=false) = self#type_variant_leaf1 attrs a p true
+  method type_variant_leaf_nobar ?opt_ampersand:(a=false) ?polymorphic:(p=false) = self#type_variant_leaf1 [] a p false
 
   (* TODOATTRIBUTES: Attributes on the entire variant leaf are likely
    * not parsed or printed correctly. *)
-  method type_variant_leaf1 opt_ampersand polymorphic print_bar x =
+  method type_variant_leaf1 attrs_on_leaf opt_ampersand polymorphic print_bar x =
```
jordwalke: I haven't looked at this code in a while, so my memory is rusty here, but how is it that we have attributes that are on the "entire leaf", joined with the "standard attributes", which are all printed as equals, and then later when we parse we somehow can recover which was which?

IwanKaramazow:
@jordwalke 
So the problem was:
```
       | Ptyp_variant (l, closed, low) ->
          let pcd_loc = x.ptyp_loc in
          let pcd_attributes = x.ptyp_attributes in
          let pcd_res = None in
          let variant_helper rf =
            match rf with
              | Rtag (label, attrs, opt_ampersand, ctl) ->
                let pcd_name = {
                  txt = label;
                  loc = pcd_loc;
                } in
                let pcd_args = Pcstr_tuple ctl in
                let all_attrs = List.concat [pcd_attributes; attrs] in
                self#type_variant_leaf ~opt_ampersand ~polymorphic:true {pcd_name; pcd_args; pcd_res; pcd_loc; pcd_attributes = all_attrs}
```
If you look add `| Rtag (label, attrs, opt_ampersand, ctl)`, it has attributes which fell outside of the `pcd_attributes` because they're in the `Rtag`. I just modified the PR to drop the weird logic I introduced and just join the `'Rtag attrs'` with the normal `pcd_attributes`. This way our printer handles them correctly.

PR #1111, opened 2017-03-03
Implement decorators

SanderSpies:
Or actually: move (item-)attributes to the front.

Not completely finished yet, there are still some formatting challenges and the expected outputs of the tests need to be adjusted.


jordwalke:
"decorators" 😆 
I think this is much more intuitive to people new to OCaml.

SanderSpies:
Opening a new one, because I'm not a Git rebase master.

PR #1112, opened 2017-03-03
Implement decorators

SanderSpies:
Or actually: move (item-)attributes to the front.

Current status: ready for review.


tekknolagi:
How do these work and where can I find documentation on them?

SanderSpies:
@tekknolagi The OCaml documentation on attributes: http://caml.inria.fr/pub/docs/manual-ocaml/extn.html#sec246. 

This PR is very simple: it moves the attributes (`[@foo]`) and item-attributes (`[@@foo]`) to the front. So `let foo = bar [@bar][@@foo]` becomes `[@@foo]let foo = [@bar] bar`. The idea is to make it more similar to what's normal in other languages like Java, C#, JavaScript, etc.

src/reason_parser.mly
```diff
@@ -3980,84 +4154,96 @@ constructor_arguments:
...
+    attributes mutable_flag as_loc(LIDENT)
       {
        let loc = mklocation $symbolstartpos $endpos in
-       let ct = mkct $2 in
-       (Type.field $2 ct ~mut:$1 ~loc, $3)
+       let ct = mkct $3 in
+       (Type.field $3 ct ~mut:$2 ~loc, $1)
       }
-  |  mutable_flag as_loc(LIDENT) attributes COLON poly_type attributes
+  |  attributes mutable_flag as_loc(LIDENT) COLON poly_type
```
SanderSpies: This is a potential area for concern, within the AST it's possible to set an attribute here - although it's not clear to me which syntax would trigger this.

src/reason_parser.mly
```diff
@@ -4236,12 +4431,22 @@ _core_type:
...
   | core_type2 AS QUOTE ident
       {
-        let core_type_loc = mklocation $startpos($1) $endpos($1) in
+        let core_type_loc = mklocation $symbolstartpos $endpos($1) in
         let ct = only_core_type $1 core_type_loc in
         Core_type (mktyp(Ptyp_alias(ct, $4)))
       }
 ;
 
+/* core_type_no_attr:
```
SanderSpies: Clean up.

src/reason_parser.mly
```diff
@@ -4268,15 +4482,22 @@ _core_type2:
       match $2, $4 with
       | Core_type ct, Core_type ct2 -> Core_type (mktyp(Ptyp_arrow(Labelled $1, ct, ct2)))
       | _ -> syntax_error()
-      }
-  | core_type2 EQUALGREATER core_type2
+    }
+;
+
+/* core_type2_no_attr: mark_position_typ2(_core_type2_no_attr) {$1}
```
SanderSpies: Clean up

SanderSpies:
Ready for review @jordwalke @yunxing

tekknolagi:
Can you rebase on top of current master to fix the 4.02.3 tests on Travis?

SanderSpies:
@tekknolagi That was done already

tekknolagi:
@SanderSpies huh. Shoot. I wonder why Travis fails.

SanderSpies:
@tekknolagi implicit_arity goes wrong for some reason.

src/reason_parser.mly
```diff
@@ -1216,65 +1217,120 @@ parse_pattern:
 
 mark_position_mod(X):
 	x = X
-	{ {x with pmod_loc = {x.pmod_loc with loc_start = $symbolstartpos; loc_end = $endpos}} }
+	{
+    if x.pmod_loc.loc_start.Lexing.pos_lnum <> 0 then
```
SanderSpies: The repositioning of the attributes has consequences for the start_pos of many first tokens. Instead of always overwriting it we now only do it when there is none.

yunxing:
@SanderSpies implicit_arity should have been fixed on master.

SanderSpies:
The build successfully completes now also on Travis.

SanderSpies:
*crickets*

SanderSpies:
FYI rebased again.

chenglou:
@jordwalke @yunxing are we good with this change?

jordwalke:
I'm absolutely excited about this change. I just haven't had time to review it thoroughly. It's going in for sure.

chenglou:
For the JS side, here's a stress test: https://pastebin.com/r2HQhWFB

chenglou:
Funny case:

```reason
[@@bs.module "MyBootIDs"] [@@bs.return null_undefined_to_opt]
external getSupportInboxItemIDFromThreadID : Js.null_undefined string => option string =
  "";
```

Note the line break at the end.

jordwalke:
For that example, it looks like it's not breaking at the right places. I would have expected:

```reason
[@@bs.module "MyBootIDs"]
[@@bs.return null_undefined_to_opt]
external getSupportInboxItemIDFromThreadID : Js.null_undefined string => option string =  "";
```

Or at least:


```reason
[@@bs.module "MyBootIDs"]
[@@bs.return null_undefined_to_opt]
external getSupportInboxItemIDFromThreadID :
  Js.null_undefined string => option string =  "";
```

hcarty:
Moving over from Discord - I think type definitions are the more interesting case in general.  It's one of the key areas where Reason brings a significant benefit to newcomers to the language, regardless of the platform they target.

Assuming a single `@` syntax, Is it better to have
```reason
@deriving yojson
type t = {
  @yojson.key "xkey" x : int,
  @yojson.key "ykey" y : @yojson.default 0 int
};
```
or
```reason
type t = {
  x : int @yojson.key "xkey",
  y : (int @yojson.default 0) @yojson.key "ykey"
} @deriving yojson;
```

Both, to my eyes, are harder to parse than keeping the `[]` delimiters because they make the start and end of the attribute definition explicit.  I don't have to understand what the attribute is doing to know where it begins and ends.

Prefix and suffix with `[]` delimiters:
```reason
[@deriving yojson]
type t = {
  [@yojson.key "xkey"] x : int,
  [@yojson.key "ykey"] y : [@yojson.default 0] int
};
```
or
```reason
type t = {
  x : int [@yojson.key "xkey"],
  y : (int [@yojson.default 0]) [@yojson.key "ykey"]
} [@deriving yojson];
```

reason-parser/src/reason_parser.mly
```diff
@@ -1200,7 +1201,7 @@ _use_file:
 parse_core_type:
     core_type EOF
       {
-        let core_loc = mklocation $startpos($1) $endpos($1) in
+        let core_loc = mklocation $symbolstartpos $endpos($1) in
```
SanderSpies: Yes, there is a difference. From the Menhir manual:
```
The position $startpos is sometimes “further towards the left” than one would like. For example, in the
following production:
declaration: modifier? variable { $startpos }
the keyword $startpos represents the start position of the optional modifier modifier?. If this modifier
turns out to be absent, then its start position is (by definition) the end position of the most recently parsed
symbol. This may not be what is desired: perhaps the user would prefer in this case to use the start position
of the symbol variable. This is achieved by using $symbolstartpos instead of $startpos
```

reason-parser/src/reason_parser.mly
```diff
@@ -1484,67 +1541,73 @@ _structure_item:
...
       mkstr
-        (Pstr_primitive (Val.mk $2 (only_core_type $4 core_loc) ~prim:$6 ~attrs:$7 ~loc))
+        (Pstr_primitive (Val.mk $3 (only_core_type $5 core_loc) ~prim:$7 ~attrs:$1 ~loc)) ~loc
     }
   | many_type_declarations {
       let (nonrec_flag, tyl) = $1 in
-      mkstr(Pstr_type (nonrec_flag, List.rev tyl))
+      let hd = List.hd (List.rev tyl) in
+      let last = List.hd tyl in
+      mkstr(Pstr_type (nonrec_flag, List.rev tyl)) ~loc:{loc_start = hd.ptype_loc.loc_start; loc_end = last.ptype_loc.loc_end; loc_ghost = false}
```
SanderSpies: Makes sense.

reason-parser/src/reason_parser.mly
```diff
@@ -1757,80 +1822,83 @@ signature:
...
-      mksig(Psig_value (Val.mk $2 (only_core_type $4 core_type_loc) ~prim:$6 ~attrs:$7 ~loc))
+      let core_type_loc = mklocation $startpos($5) $endpos($5) in
+      mksig(Psig_value (Val.mk $3 (only_core_type $5 core_type_loc) ~prim:$7 ~attrs:$1 ~loc)) ~loc
     }
   | many_type_declarations {
-        let (nonrec_flag, tyl) = $1 in
-        mksig(Psig_type (nonrec_flag, List.rev tyl))
+      let (nonrec_flag, tyl) = $1 in
+      let hd = List.hd (List.rev tyl) in
+      mksig(Psig_type (nonrec_flag, List.rev tyl)) ~loc:hd.ptype_loc
```
SanderSpies: I had quite some challenges with the handling of whitespace and this resulted in some not so pretty hacks. This is definitely one of them, however I'm not sure how to improve upon it without breaking the handling of whitespace.

reason-parser/src/reason_parser.mly
let-def: Ah yes but I think in this case it is not going to do anything because core_type is not inlined and not nullable, so `$symbolstartpos == $startpos(1)` (not nullable means it cannot be absent, not inlined means it is not going to be expanded in one or more producers that could be nullable themselves).
It is correct nonetheless, so that's fine :).

reason-parser/src/reason_parser.mly
```diff
@@ -161,6 +161,13 @@ let make_ghost_loc loc = {
 
 let ghloc ?(loc=dummy_loc ()) d = { txt = d; loc = (make_ghost_loc loc) }
 
+let rev_list_extract_loc prj lst =
+  let last = List.hd lst in
+  let lst = List.rev lst in
+  let head = List.hd lst in
+  let loc = mklocation (prj head).loc_start (prj last).loc_end in
+  (lst, loc)
+
```
let-def: @SanderSpies: this is an implementation of the idea I suggested. Do you think it could be a problem with handling of whitespaces? How so?

reason-parser/src/reason_parser.mly
jordwalke: It would be interesting to see if it ends up effecting the locations. Does it affects more than *just* optionals? Suppose you have empty sequences (like attributes)?

jordwalke:
Off topic, but for posterity - in the future, I would suggest only omitting the `[ ]`  for payload-less attributes. It also seems to be a good heuristic for when to force a break between the attribute and the thing it describes.

```rust

@async let x = something ();


[@longerThing withPayload foo]
let y = something ();
```

reason-parser/src/reason_parser.mly
let-def: The effect of changing `$symbolstartpos` to `$startpos($1)`:

```
-        let _v : (Ast_404.Parsetree.core_type) = 
-        let _symbolstartpos = _startpos__1_ in (
-        let core_loc = mklocation _symbolstartpos _endpos__1_ in
+        let _v : (Ast_404.Parsetree.core_type) =       (
+        let core_loc = mklocation _startpos__1_ _endpos__1_ in
```

jordwalke:
So, final thing remaining is - what's our plan to not break people. This is a breaking change, and people should have the ability to upgrade easily. Thankfully, there's something that should work for that called [reup](https://github.com/facebook/reason/blob/master/src/reup.sh).

- We should test `reup`.
- People should be modeling their dependencies, so that it would be impossible for their builds to break as long as they're depending on the right reason version. That means if I write my syntax in 1.13, then it's marked as a dependency.
- As long as that's the case, we can commit.

There will likely be more breaking changes in the future with syntactic overhauls so people must get into best practices anyways.

Eventually, I imagine something like ocaml-migrate-parse-tree, but for *publishing* pre-processed code, so that package version conflicts can't happen when one package depends on reason x, and another depends on reason y.

reason-parser/src/reason_parser.mly
let-def: `$symbolstartpos` is determined locally. Exactly the same with attributes: an empty sequence is an optional (nullable producer) that isn't there.

To determine symbolstartpos:
```
let rec symbolstartpos i =
  if i >= producers then $endpos
  else if null($i) then symbolstartpos (i+1)
  else $startpos($i)
in
symbolstartpos 0
```

The code generated is a bit more efficient, by unrolling the loop manually up to the first non-nullable producer.
But the main point is that this is purely local: it doesn't do anything that you couldn't by manually manipulating `$startpos($i)`

let-def:
Then we need a list of breaking changes, with names for each non-compatible version of the language.

reason-parser/src/reason_parser.mly
SanderSpies: This looks great, I don't see how this would cause any issues.

reason-parser/src/reason_parser.mly
SanderSpies: Some extra background: I changed all the `$startpos($1)` to `$symbolstartpos` as that seems to be the one we want to have always anyway.

reason-parser/src/reason_parser.mly
```diff
@@ -3980,84 +4154,96 @@ constructor_arguments:
...
-    mutable_flag as_loc(LIDENT) attributes
+    attributes mutable_flag as_loc(LIDENT)
       {
        let loc = mklocation $symbolstartpos $endpos in
-       let ct = mkct $2 in
-       (Type.field $2 ct ~mut:$1 ~loc, $3)
+       let ct = mkct $3 in
+       (Type.field $3 ct ~mut:$2 ~loc, $1)
       }
-  |  mutable_flag as_loc(LIDENT) attributes COLON poly_type attributes
```
let-def: Note: the last `attributes` on the line is the only way to fill the `Type.fields ~attrs` / `pld_attr` field.
There is no way to fill that in the new syntax. But it seems this information is dropped later anyway so it is safe to lose it ?!.

SanderSpies: Do you know what equivalent OCaml syntax would be for this?

bsansouci:
How's this PR going? Did we decide on some other way to simplify ppx?

hcarty:
This is included in #1299 now I think.

formatTest/typeCheckedTests/expected_output/attributes.rei
```diff
@@ -11,29 +11,32 @@ let test: int;
 /**
  * #990: don't strip attributes in interface files
  */
-let x: int [@@bs.val];
+[@@bs.val] let x: int;
```
jaredly: I think I'd want the decorator to be printed on the previous line

formatTest/unit_tests/expected_output/infix.re
```diff
@@ -891,30 +891,30 @@ let containingObject = {
...
     let result = 2 + (- add 4 0);
     /* That same example but with ppx attributes on the add application */
-    let result = 2 + (- add 4 0 [@ppx]);
+    let result = 2 + (- [@ppx] add 4 0);
     /* Same as */
-    let result = 2 + (- add) 4 0 [@ppx];
+    let result = [@ppx] 2 + (- add) 4 0;
     /* Same as */
-    let result = 2 + (- add 4 0 [@ppx]);
+    let result = [@ppx] 2 + (- add 4 0);
```
jaredly: huh this got semantically quite different. maybe it was wrong before?

formatTest/unit_tests/expected_output/infix.re
```diff
@@ -891,30 +891,30 @@ let containingObject = {
     /* Same as */
     let result = 2 + (- add 4 0);
     /* That same example but with ppx attributes on the add application */
-    let result = 2 + (- add 4 0 [@ppx]);
+    let result = 2 + (- [@ppx] add 4 0);
     /* Same as */
-    let result = 2 + (- add) 4 0 [@ppx];
+    let result = [@ppx] 2 + (- add) 4 0;
```
jaredly: I can't figure out how this works syntactically -- shouldn't `4 0` be inside the `add` parens?

formatTest/typeCheckedTests/expected_output/attributes.rei
jordwalke: Yeah, perhaps unless there is no payload so that we could do cool ppx transforms like:

```
@async let foo = (x, y) => ...;
```

formatTest/unit_tests/expected_output/infix.re
jordwalke: Maybe in this PR, the original *input* file might have been changed to different semantics as well, which explains why the output would be changed semantically.

chenglou:
Incorporated into #1299 

Come back @SanderSpies we miss you lol

PR #1117, opened 2017-03-06
[Test] Use OCaml version that is installed

yunxing:


formatTest/test.sh
```diff
@@ -9,6 +9,7 @@ INFO=''
 DEBUG=''
 RESET='\033[0m'
 VERBOSE=${VERBOSE:-}
+OCAML_VERSION=`echo $(ocaml -version) | egrep -o '[0-9]+.[0-9]+.[0-9]+' | head -1`
 OCAML_VERSION=${OCAML_VERSION:-"4.02.3"}
```
jordwalke: Did you mean to remove this following line?

yunxing: Let's leave it there in case we can't get OCAML_VERSION from line 12.

PR #1126, opened 2017-03-07
Remove predef option in AST

yunxing:
We needed to add a type constraint "*predef*.option" in 4.02.3. Now that we have migrate-parsetree to take care of this, we don't need to manually add it anymore.


let-def:
If necessary, it is possible to add a test for this. In ocaml syntax:
```
module X : sig
  val x : ?x:unit -> unit -> unit
end = struct
  let x ?(x=()) () = ()
end
```
If *predef* is in the ast, compiling the file will succeed with 4.02 and fail with 4.04.

jordwalke:
Good find @yunxing

PR #1128, opened 2017-03-09
Add git diff exit code in right place

tekknolagi:
I misunderstood the Travis config file when I first read it


yunxing:
Good catch.

PR #1134, opened 2017-03-11
I think push needs to have the "tag" prefix

tekknolagi:
Otherwise my tags don't get pushed unless I manually do `--follow-tags` post-release


yunxing:
Thanks!

PR #1137, opened 2017-03-13
I think this should fix unbound module Ast_404

tekknolagi:
I don't really know how OCaml works exactly, but my I think this fixes @rickyvetter's issue. @rickyvetter can you try on this branch and see if your issue is fixed?

### DIFF
--- a/pkg/META.in
+++ b/pkg/META.in
@@ -18,7 +18,9 @@ archive(byte, toploop, -pkg_utop) += "reason_toploop.cmo"
 archive(byte, toploop, pkg_utop) += "reason_utop.cmo"
 
 archive(byte) = "@result/result.cmo"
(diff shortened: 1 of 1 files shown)


yunxing:
@tekknolagi Do we still need this? I think Ricky has fixed this on his side.

rickyvetter:
Yeah I don't think this is needed. Closing since it's working on my end. Thanks for the help! :)

jordwalke:
Thanks everyone. Mind explaining the fix for posterity

tekknolagi:
@rickyvetter added ocaml-migrate-parsetree to his package.json, as far as I am aware.

jordwalke:
Ah, that's what I thought would solve the problem. @yunxing Note that our theory was correct.

PR #1147, opened 2017-03-19
[Docs] Update docs for emacs support

rubas:
The old integration example was still based on a global opam installation. Let's
use the reason-cli instead and add a section for `Spacemacs`.

> The best editor is neither Emacs nor Vim, it's Emacs _and_ Vim!

### DIFF
--- a/editorSupport/emacs/README.md
+++ b/editorSupport/emacs/README.md
@@ -60,7 +60,8 @@ This associates `reason-mode` with `.re` and `.rei` files. To enable it explicit
 
 ### Spacemacs
 
-There is currently no reason layer available, but you can install the `reason-mode` package automatically.
+There is currently no offical reason layer available, but you can install the `reason-mode` package automatically.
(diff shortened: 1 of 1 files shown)


editorSupport/emacs/README.md
```diff
@@ -8,36 +8,80 @@ https://github.com/rust-lang/rust-mode
...
+
+Afterwards add the [snippet](#manual-installation) to your `dotspacemacs/user-config`.
 ### Features
 
 #### Auto-format before saving
 
 If you have refmt installed, you can add this to your `.emacs` file to enable
 auto-format:
-```
+``
```
iensu: There's a missing backtick here :)

chenglou:
Note that Reason-cli is still work in progress. I'll hold this off for a week or two. Thanks!

chenglou:
Bah we'll come back and update that bit later. Thanks!

PR #1148, opened 2017-03-19
Updated reason special-types list in reason-mode

fredyr:
- removed old rust types
- added a few reason/ocaml types that were missing
- changed the special types to be using builtin-face instead of type-face


chenglou:
Thanks!

chenglou:
cc @yunxing

PR #1150, opened 2017-03-19
[docs]: reword section about Functors

anmonteiro:


chenglou:
Thanks! Should have run that through Word.

PR #1151, opened 2017-03-20
Add basic docs for `while`

anmonteiro:
Noticed that the section on `while` was empty in the docs, so went ahead and added some basic documentation for it.

Let me know if it needs any tweaking, or if it was empty for any other reason.


chenglou:
Thanks! I'll add a pattern to break out of the loop

PR #1154, opened 2017-03-20
[Docs] missing backslash in Reason translation of /*

anmonteiro:


tekknolagi:
Why does it need additional slashes?

anmonteiro:
to render the HTML properly. this is a simple typo, other examples also have them

tekknolagi:
When I tested your diff, it rendered incorrectly in my browser, but the original renders just fine.

anmonteiro:
There might be a misunderstanding here. What is incorrectly for you?

I expect it to render `/\*`, but the current live website is rendering `/*`.

chenglou:
Yeah this is right. Thanks!

Issue #1156, opened 2017-03-21
bspacks is broken by adding ocaml-migrate-parsetree


tekknolagi:
@let-def we're having trouble bspacking ocaml-migrate-parsetree. More specifically, having trouble including the appropriate files after preprocessing with `make`.

```
File "migrate_parsetree_versions.ml", line 696, characters 23-42:
Error: Unbound module OCaml_OCAML_VERSION
```

We're trying to figure how we can preprocess the source files then pack them all together as we do with result, reason, refmt, etc.

chenglou:
Precision: `./bspacks/reason_bspack.sh` uses bspack, to which we add `-I src/paths`, to pack every file into a single file (good for cutting out intermediate tooling, for windows, etc.)

chenglou:
Btw, somewhat related: upgrading bspack flow with #1157

let-def:
If the version of OCaml is fixed, you can directly run sed with s/OCAML_VERSION/40x/g.

The processed files have the .pp suffix, here you want:
`ocaml-migrate-parsetree/_build/default/src/migrate_parsetree_versions.pp.ml`

chenglou:
@let-def thanks. After patching things & renaming files & bringing `Result` into scope (I'm on 4.02.3), I get this:

```
File "ppx_deriving.ml", line 7, characters 5-24:
Error: Unbound module Ast_convenience_404
```

let-def:
Then ppx_tools_versioned is not in scope.

chenglou:
@let-def ok thanks. @tekknolagi and I managed to move until this point now:

```
File "reason_parser.ml", line 5, characters 2-42:
Error: Unbound value MenhirLib.StaticVersion.require_20160825
```

A bit lost here. Do you know what's happening? See the updated #1157.

jordwalke:
Perhaps we should just get some prebuilt binaries?

PR #1157, opened 2017-03-21
Including more of the recent deps in bspack workflow

chenglou:
@tekknolagi

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ node_modules
 /bspacks/**/*.exe
 /bspacks/migrate_parsetree_pp_src
 /bspacks/ppx_deriving_ppx_src
+/bspacks/ocaml-migrate-parsetree.1.0
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,4 @@ node_modules
 /bspacks/**/*.o
 /bspacks/**/*.exe
 /bspacks/migrate_parsetree_pp_src
+/bspacks/ppx_deriving_ppx_src
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ node_modules
 /bspacks/**/*.cm*
 /bspacks/**/*.o
 /bspacks/**/*.exe
+/bspacks/migrate_parsetree_pp_src
--- a/bspacks/reason_bspack.sh
+++ b/bspacks/reason_bspack.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/zsh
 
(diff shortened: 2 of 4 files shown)


chenglou:
ok @tekknolagi this works, after following https://github.com/facebook/reason/issues/1156#issuecomment-288009930
Guess what the new error is though!

```
File "ppx_deriving.ml", line 105, characters 8-12:
Uninterpreted extension 'expr'.
```

Time to macroexpand this? I'm fine doing this at build time rather than pre-expanding them

tekknolagi:
@chenglou I'd rather expand at build-time, yes. We can use the `-ppx` flag.

tekknolagi:
What if, instead of sed, we just mock the `Result` module?

chenglou:
migrate_parsetree_ast_io.ml and migrate_parsetree_ast_io.mli assume > 4.02. We can just qualify these with `Result.result` in that repo and that should do the trick.

tekknolagi:
Rebased on master.

chenglou:
@let-def Sorry, we're a bit blocked on this one and don't know how to proceed. Do you know what's the error here?

```
File "reason_parser.ml", line 5, characters 2-42:
Error: Unbound value MenhirLib.StaticVersion.require_20160825
```

let-def:
This happens if the parser was not generated by the same version of menhir as the one installed. Is that possible?
In that case, it might be possible to bundle menhirlib version (if you embed the parser, then you should also embed the corresponding menhirLib).

let-def:
(MenhirLib is licensed under LGPL2)

chenglou:
@yunxing pushed our latest changes, can you se if you can repro this:

```
File "refmt_main.ml", line 1:
Error: No implementations provided for the following modules:
         Parsetree referenced from refmt_main.cmx
         Asttypes referenced from refmt_main.cmx
```

Note the installation requirements at the top of the file. Thanks.

chenglou:
New instructions out at https://github.com/facebook/reason/commit/575feb86bf1c780085cf1f7a54b951dd06a2f7ed

> The compiler pipeline discards the output of ppx

bspack works on source files and doesn't support ppx. For now we'll trail one version behind and get the artifacts of reason from `npm install -g reason-cli` (which itself uses opam under the hood), pack them, and do whatever with them.

Hopefully one day we'll be able to pack the repo and put the workflow under CI or something. This way `npm install reason` would just compile a few big reason source files: refmt, reactjs_jsx_ppx, refmttype, and ocamlmerlin-reason (currently not possible to pack?). No need to emulate opam with the npm install flow.

Issue #1146, opened 2017-03-19
[HIGH PRI] Bug in JSX parsing of empty list.


jordwalke:
```reason
let div ::children () => 0;

/* Works */
let x = Some <div> </div>;

/* Doesn't parse correctly */
let x = Some <> </>;
```

In the second example, it thinks that we're not applying the empty `<> </>` as the argument to `Some`.

To fix this, determine what is different about the parsing precedence between `<>`, `</>` and `<div>`, `</div>` and eliminate that difference.

In general, we should think about all tags as being "simple" expressions like `(stuff)` or `[ ]` or `{ ... }` because the endpoints perfectly balance. This is true of everything, but for some reason, the empty tags are being parsed at a different precedence than an opening `<div`.

@SanderSpies

chenglou:
Already fixed

PR #1158, opened 2017-03-21
Fix JSX empty list precedence.

SanderSpies:
Fix for #1146


chenglou:
This seems already fixed, weird

PR #1162, opened 2017-03-23
[Docs] Use structural equality for while-true example

chenglou:
This was previously committed but something something tag master
publish something. Uncommitted and making a PR instead.

@tekknolagi

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -874,7 +874,7 @@ Example to break out of a while-true loop:
 Random.self_init ();
 let break = {contents: false};
 while (not break.contents) {
-  if (Random.int 10 === 3) {
+  if (Random.int 10 == 3) {
     break.contents = true
   } else {
     print_endline "hello"


yunxing:
These two are exactly the same, in terms of performance and semantics, right?

hcarty:
@yunxing For this specific example that's true since, at least for ocamlc/ocamlopt, integers are tagged rather than boxed.  But in the case of boxed values they are different semantically.

jordwalke:
For integers, yes - and I believe it will even be specialized to integer comparison if the types are clear.

jordwalke:
There was a thread in Discord about the differences and pitfalls of each. Can't wait for modular implicits! In the mean time, I choose to only use structural comparison for certain types that I know don't have pitfalls: Pretty much only Strings, and Floats, but *even* floats have pitfalls in almost every popular encoding - you typically want to use a float epsilon comparison to test if a float is within some range. This is not a problem with structural comparison so much as it is a problem with floating point representation in general.

hcarty:
Physical equality (OCaml's `==` and Reason's `===`) is almost never the right choice for comparison.  If it's used in intro example code then there will be lots of questions about this seemingly nonsensical result happens:
```reason
Reason # 2.0 === 2.0;
- : bool = false
```
As @jordwalke said there are lots of issues with the default structural equality.  But it's far more likely to represent what user's want from basic comparison.

bsansouci:
My opinion is that this matches JS's behavior more closely. The only exception as you've pointed out is for strings (floats can't be compared like that in JS either).

hcarty:
Ultimately, on the `==`/`===` front, the "right" answer is _don't use them_.  But that's not always an option without using external libraries and/or lots of local boilerplate.  And for quickly thrown together code `==` can be quite useful.

Ideally `===` would get some other not-so-easy-to-typo-into name like `physeq` since it's not something a beginner is likely to need.

Schmavery:
I feel like the original reason we created a `===` was to be more familiar/intuitive to js coders who are used to `===` meaning some kind of "more precise" equality.  It seems however (to me at least) that the semantics in reason are sufficiently different that `===` existing has the opposite effect (ie behaves confusingly).
Maybe something like `=*` (instead of `===`) would have some mnemonic value re physical/pointer equality.

hcarty:
I like `=*` a lot from a C-ish perspective. I'm not sure how well the meaning carries across languages. Maybe `==*` to leave space for the opposite in `!=*`? That starts to be a little crazy looking but maybe that's​ best for something with limited applicability.

jordwalke:
@Schmavery I made the case in the Discord chat that Reason's `===` is actually not that different from ES6's `===`. The primary difference we observe come from a difference in how values are represented, not the conceptual meaning of `===`.

Here's an overview of the main arguments *against* matching JS's operators, and why I think those arguments aren't as strong as you might think.

> " In JS,`"a" === "a"` is `true`, but with `Reason`, `"a" === "a"` is false".

  - I think this argument doesn't hold up strongly because you can think of the JS string literals as being "interned" or lifted into the data section at compile time, so of course even if they are allocated on the heap, you could think of them as having the same pointer address so reference equality `===` would return true.
    - *However* - you can construct a scenario for strings where reference equality *doesn't* obey that "allocation lifting": `new String("a") === new String("a")` evaluates to false.
    - In `ocamlopt` Strings are still mutable, so they cannot be "lifted" to the data section and so basically every string you create is like doing `new String("a")` in JS - so if you consider `Reason`'s `===` to be an analog for JS's `===`, the similarity becomes more clear.
    - When Strings in OCaml are fully immutable, you could imagine them being lifted into the data section, and then even Reason with `===` in `ocamlopt` will behave like JS's `===` for Strings. (I'm not sure if/when this will happen).
    - In BS, string literals use JS string literals, and are known to be immutable and therefore *conceptually* lifted, and therefore in BS, `"a" === "a"` behaves as it does in JS, and behaves how OCaml might one day behave if strings were guaranteed to be immutable.
- In JS, for *mutable* allocations, people very much understand that `x === x` being false implies nothing, and that is the same as it is for Reason's `===`.
- `===` has many pitfalls but they're mostly familiar to JS developers, and the ones that aren't familiar could possibly go away completely, or have more to do with the differences in how values are represented than the differences with `===` itself.

> "In JS `1.0 === 1.0` but in Reason `1.0 === 1.0` is false"

This is the one that I think would cause the most surprises. But again, this is more about the representation of values in JS vs. in OCaml's runtime. JS engines often perform nanboxing and therefore the two separate floats with the same exact physical encoding, are packed into one "pointer", and therefore comparing their physical address amounts to comparing their float encoding itself. (I don't know how VMs without nanboxing do it). OCaml runtime boxes float allocations and so comparing physical address isn't comparing their float encoding. *However* - I'm actually not concerned about this because comparing floats using *either* `==` or `===` in JS, *or* Reason is fundamentally flawed *anyways* and we should be comparing to some epsilon difference. For example, in JS: Both `0.1 + 0.2 === 0.3` *and* `0.1 + 0.2 == 0.3` evaluate to *`false`*! The same is true of `Reason`'s `==` and `===`.  So clearly, using equality to compare floating point numbers is flawed in the first place in either language. The difference is that for Reason, there is a difference between `==` and `===` for *some* encoded pairs of floats. `1.0 == 1.0` is true and `1.0 === 1.0` is false. But again, equality on floats is just broken anyways and you shouldn't be using it. So I propose that differences in float equality between JS's `===` and Reason's `===` matter less than you might think.

- Still, I believe that Reason's `==` has many more pitfalls than Reason's `===`, and nearly all of those pitfalls of `==` are things that JS developers have never encountered (crashing your app when used on cyclical data, or if anything reachable in that memory graph has a function). If I had to pick one to ban - it would be `==` (actually - we can replace it with modular implicits' equality).

For now, `Reason`'s `===` and `!==` make quite a bit of sense to me coming from JS, and the corresponding `==` and `!=` make less sense, which is why I nominate them for being replaced by modular implicits. However, as it is, once you learn `===` and `!==` as "referential equality", the syntax of `==` and `!=` as "more loose equality" is a mnemonic device that helps you at least remember which is which. With OCaml's current syntax, not only is equality broken for many things (like it is in JS and Reason) but it's hard to even remember *which* broken one to use effectively. At least with a mnemonic devices we can help people how to use the right broken tool effectively.

cristianoc:
@jordwalke it would be great to have per-type deprecation, as equality on floats has historically created a lot of damage. Something for modular implicits, I guess.

jordwalke:
Modular implicits would be awesome and perhaps we can do something in the short term.
Is there a way to blacklist *specific* types? A sort of anti-type?

We could shadow the `==` / `!=` and make people access them through `polymorphic_compare`.

bsansouci:
@jordwalke Thanks for the writeup! This was a very detailed description of what's going on.

My opinion here is similar to @Schmavery in that `===` not working like JS for strings and floats _is_ a stumbling block for newcomers and so maybe we should focus on finding a way to fix that. I remember hearing about a way to write a ppx on the tree of types, is that usable today?

By the way, I just looked at the docs (and tried in rtop) and it seems like `==` actually infinite loops on cyclic data structures and throws on function comparison.

hcarty:
@jordwalke 
> We could shadow the == / != and make people access them through polymorphic_compare.

I think that's worth doing for all of `==`/`!=`/`===`/`!==`.  None of them are really what one wants in the general case.

@bsansouci ppx_deriving has derivers for comparison and equality.

chenglou:
Well this sure bike shedding into oblivion =P

PR #1166, opened 2017-03-26
[Docs] Clarify include vs open

chenglou:


docs/modules.html
```diff
@@ -120,6 +120,8 @@
 };
 ```
 
+**Note**: `include` and `open` are very different! The former literally copies over definitions into the new module for it to appropriate it. The latter simply refer to stuff in the `open`ed module's scope.
```
hcarty: > **Note**: `include` and `open` are very different! The former literally copies over definitions into the new module for it to appropriate it. The latter simply refer to stuff in the `open`ed module's scope.

Some suggestions for wording tweaks:
- `for it to appropriate it` -> `for it to appropriate`
- `the later simpler refer to stuff in` -> `the later simply allows unqualified access to values, types and constructors in`

chenglou: "stuff" was intentional lol. I'll bring a clarification though.

chenglou:
Updated wording, thanks.

PR #1170, opened 2017-03-27
Require user to manually add ppx_deriving runtime

tekknolagi:
This fixes some slowdowns when using compiled JS code.

/cc @chenglou


tekknolagi:
Should probably bump version after this.

PR #1173, opened 2017-03-28
Remove rebuild hack

yunxing:
We had a lot of hacks into rebuild so that we could statically link our
custom ocamlbuild rules into rebuild. It was a pain to maintain and debug.

Max and I found a much less hacky approach to link the custom rules defined
in reasonbuild.ml:

1. Move reasonbuild.ml into a tmp directory, call it myocamlbuild.ml
2. Run ocamlbuild with '-just-plugin' option in that directory
3. Step 2 will generate a binary called "myocamlbuild", rename it and ship it
as "rebuild"

After step 3, 'rebuild' can now be used as an augmented 'ocamlbuild' with custom rules.


jordwalke:
Will this approach play with projects that already have their own myocamlbuild? Will it union the two rules?

tekknolagi:
@jordwalke it looks like yes, this is basically the same as old workflow.

yunxing:
@jordwalke Yes. This prebuilds the augmented version of ocamlbuild in a temp directory when installing reason itself. It doesn't conflict with myocamlbuild.ml

yunxing:
Less hack and better engineering!

PR #1179, opened 2017-03-29
Minor fixes to Makefile and release script

yunxing:
This change calls make reason-parser in reason-parser first to
generate correct package.json and opam files for reason-parser.

@tekknolagi


tekknolagi:
Looks good. Can't wait to test this!

Issue #1167, opened 2017-03-26
refmt: allow --in-place to accept many files


chenglou:
This way we can enable `refmt --in-place src/*.re`

tekknolagi:
This would more generally require `refmt` to support multiple arguments. Certainly a possibility.

PR #1180, opened 2017-03-29
Allow refmt to refmt many files at once

tekknolagi:
Fixes https://github.com/facebook/reason/issues/1167

/cc @chenglou @yunxing


tekknolagi:
Just one thing: this doesn't make sense except for with `--in-place`, since it will otherwise concatenate all of the reformatted output.

PR #1183, opened 2017-03-31
Small fix for comments and refmt shortcut.

bsansouci:
This is a step towards better Sublime support!

Ideally we'd dig into the project's node_modules to find `refmt` after trying to see if there's a global one.


chenglou:
@bsansouci we should pull this out and make it legit.

PR #1185, opened 2017-04-02
Export 

SanderSpies:
WORK IN PROGRESS

This PR is not finished and only contains the initial work for the parser. It is intended to start the discussion on the `export` keyword. 

It builds upon the decorators PR.


SanderSpies:
`export` will be implemented with help of the `include` statement and probably some attributes.

For example:
```
export type t = {myRecordField: int};
let x = 10;
```
will be parsed as 
```
  include ({
      type t = {myRecordField: int};
      let x = 10;
    } : {
      type t = {myRecordField: int};
    });
```

PR #1189, opened 2017-04-03
Use built-in emacs functions to resolve true filename rather than realpath

sgrove:
Thanks to @rubas for the heads up that this needs to be updated.

This is one of the last blockers (alongside `python2` not being symlinked in the latest OSX) I've seen in setting up ~10 OSX computers.

Thanks to @kevzettler for being a test subject here!


jordwalke:
cc @yunxing just to spot any things that wouldn't work.

yunxing:
LGTM

PR #1196, opened 2017-04-06
Move sublime integration to a dedicated place

chenglou:
The new place fixes refmt & snippets and can be installed from Package
Control.


chenglou:
CircleCI passes

PR #1203, opened 2017-04-08
Change type 'integer' to 'int' in doc

irrigator:
I think here 'integer' should be 'int' or is there something that I am missing?


chenglou:
No you're right, thanks!

PR #1215, opened 2017-04-14
Change reason-bin installation instructions

twoquestions:


twoquestions:
Signed agreement as requested.

chenglou:
Oh crap, I've accidentally pushed that, sorry. I've reverted it back to reason-cli for now.

The plan is to test reason-bin more before recommending it. I'll leave this PR open until we deem it solid. Thanks for catching this!

yunxing:
Let's have more tests before making this the official instruction: https://github.com/yunxing/reason-bin

Issue #1207, opened 2017-04-10
Auto-break records with two or more fields


IwanKaramazow:
Refmt gave me the following:
```reason
type t 'a = {size: int, buffer: array 'a, pre: int, post: int};
```
This doesn't linebreak because the line-length isn't exceeded.
For readability purposes, shouldn't this break over multiple lines?
```reason
type t 'a = {
  size: int,
  buffer: array 'a,
  pre: int,
  post: int
};
```
I would propose that all records with two or more fields break over multiple lines for readability & ease of extension. Any opinions on this one?

yunxing:
To me it looks better than putting them into one line.

Since we don't see many single field records, maybe we can always break.

andreypopp:
I propose breaking record into multiple lines if in input there was a break somewhere before the closing `}`. The idea is that a line break in input should signify an intent to format a record with line breaks.

#### Should break
```
type t = {
  a: int, b: int};
```
```
type t = {a: int,
  b: int};
```
---
```
type t = {
  a: int,
  b: int
};
```
#### Should NOT break
```
type t = {a: int, b: int};
```
```
type t = {a: int, b: int
};
```
---
```
type t = {a: int, b: int};
```

andreypopp:
Btw with the current syntax #1216 was invalidated and `refmt` doesn't break records with two or more fields in it.

Also the thing I proposed mimics `prettier` behaviour which I'm personally enjoying.

IwanKaramazow:
Shouldn't be too hard, I'll see what I can do

anmonteiro:
works in the latest reason

PR #1216, opened 2017-04-14
The joy of records

IwanKaramazow:
https://github.com/facebook/reason/issues/1207 Always break records with two or more record rows.
The goal of this PR is to make working with records more readable and enjoyable.
How rare single-row records might be, I didn't break them because
inline records (variant) are pretty uggly when they break.

### DIFF
--- a/formatTest/unit_tests/expected_output/features403.re
+++ b/formatTest/unit_tests/expected_output/features403.re
@@ -12,20 +12,18 @@ type nonrec u 'a =
 
 type expr 'a =
   | Val {value: 'a} :expr 'a
-  | Add
-      {
-        left: expr int,
-        right: expr int
-      }
-      :expr int
+  | Add {
+      left: expr int,
+      right: expr int
+    }
+    :expr int
   | Is0 {test: expr int} :expr bool
-  | If
-      {
-        pred: expr bool,
-        true_branch: expr 'a,
-        false_branch: expr 'a
-      }
-      :expr 'a;
+  | If {
+      pred: expr bool,
+      true_branch: expr 'a,
+      false_branch: expr 'a
+    }
+    :expr 'a;
 
 let rec eval: type a. expr a => a =
   fun e =>
--- a/formatTest/unit_tests/expected_output/features403.re
+++ b/formatTest/unit_tests/expected_output/features403.re
@@ -13,7 +13,11 @@ type nonrec u 'a =
 type expr 'a =
   | Val {value: 'a} :expr 'a
   | Add
-      {left: expr int, right: expr int} :expr int
+      {
+        left: expr int,
+        right: expr int
+      }
+      :expr int
   | Is0 {test: expr int} :expr bool
   | If
       {
--- a/formatTest/unit_tests/expected_output/wrappingTest.re
+++ b/formatTest/unit_tests/expected_output/wrappingTest.re
@@ -397,7 +397,10 @@ type blahType = string;
 
 let x: blahType = "asdf";
 
-type nameAge = {age: int, name: string};
+type nameAge = {
+  age: int,
+  name: string
+};
 
 type hasABunch = {
   /*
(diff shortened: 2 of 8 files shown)


chenglou:
Looks good.

Maybe we should have a predictable convention here. Variants, `|>` and record break into multiple lines inconsistently.

IwanKaramazow:
We could generalise this into a `two-or-more`-rule. Variants currently always break, even if there's one. Which might make sense in the  case of a variant, but doesn't in the case of an inline record.
`|>` & others like `>>=` fall in the `two-or-more`. But what do you do with `^` ?
```
let concattedString =
     "a"
   ^ "b"
   ^ "c"
```

jordwalke:
For some reason this seems like the right thing to do for types, even though it would be too aggressive for values.

formatTest/unit_tests/expected_output/wrappingTest.re
```diff
@@ -2060,11 +2063,18 @@ type myRecordWithReallyLongName = {
 
 type doubleEqualsRecord =
   myRecordWithReallyLongName =
-    {xx: int, yy: int};
+    {
+      xx: int,
+      yy: int
+    };
```
yunxing:
```
type doubleEqualsRecord =
    myRecordWithReallyLongName = {
      xx: int,
      yy: int
    };
```

I think we should somehow lift the left "{" to the line above it so it could be more consistent. Not sure if our current formatting API has the power to do that though.

IwanKaramazow: I'll try, with a bit of luck it's just inserting a label.

IwanKaramazow:
@yunxing 
the left "{" has been lifted.

formatTest/unit_tests/expected_output/features403.re
```diff
@@ -13,7 +13,11 @@ type nonrec u 'a =
 type expr 'a =
   | Val {value: 'a} :expr 'a
   | Add
-      {left: expr int, right: expr int} :expr int
+      {
```
hcarty: Can the same `{` lifting to the previous line happen here?  It would be more consistent with the rest of the styling.

formatTest/unit_tests/expected_output/features403.re
```diff
@@ -13,7 +13,11 @@ type nonrec u 'a =
...
+        left: expr int,
+        right: expr int
+      }
+      :expr int
```
hcarty: This may be worth lifting to the previous line too.  I'm less certain here though.

IwanKaramazow:
```
  | Add {
      left: expr int,
      right: expr int
    } :expr int
```
The last `:expr int` (on the record), looks a bit strange, doesn't it?

formatTest/unit_tests/expected_output/features403.re
IwanKaramazow: Shouldn't be much of a problem, will improve the formatting

formatTest/unit_tests/expected_output/features403.re
hcarty: A space after the `:` for the constructor's type may help in either case and is more consistent with the formatting used in records:

```
| Add {
    left: expr int,
    right: expr int
  } : expr int
```
```
| Add {
    left: expr int,
    right: expr int
  }
  : expr int
```
Looking at those two example, I like that the first is more compact.  In the second the `: expr int` looks a little strange on its own.  I think either would work.

IwanKaramazow: the space was missing ☺️
Will try & implement the first one, looks smoother.

IwanKaramazow:
@hcarty bad news, placing the closing `}` & gadt res like `:expr int` on the same line, isn't possible with the current formatting api. It would take a major rewrite. 
I did manage to place improve the indentation of `{`& `}` with inline records.
```reason
  | Add {
      left: expr int,
      right: expr int
    }
```

This PR is ready.

hcarty:
@IwanKaramazow Thank you for looking into it!  This new formatting is already a big improvement.

chenglou:
Thanks!

PR #1223, opened 2017-04-18
Fix one semicolon in mlCompared.html

arichiardi:
My first PR on this project on what I think is a typo (not sure though, as I am still studying Reason's syntax).


jordwalke:
It is not actually not really correct to imply that lambdas end with semivolons. Imagine passing them to Array.map.

arichiardi:
Great this is exactly that kind of feedback I was hoping for when opening the issue, could you write me an example of "phrase" that would imply the above?

jordwalke:
Cool, thanks for checking out Reason. Never hesitate to send a pull request!

PR #1234, opened 2017-04-25
reason-parser: ppx_tools_versioned should at least be beta

let-def:
The alpha has incorrect dependency specifications that can cause build failures.


jordwalke:
Thanks!

PR #1235, opened 2017-04-25
Made ; optionals when non ambiguous

let-def:
This PR doesn't change printing, but allow to omit `;` at the end of sentences when it is not ambiguous:

For instance:
```reason
type t = int
let x = 5
let module X = {
  let x = 5
  let y = 7
}
let xy = {
  let x = 5
  let y = 7;
  x * y
}
```

However, `item_extension_sugar` is disabled for now. I am looking at reimplementing it, but I am not sure whether this feature is used or not.


jordwalke:
This would save a ton of pain when editing. I usually hit `refmt` in the editor but often forgot a semicolon. With this PR, it would just insert them where needed!

let-def:
Do you know about `item_extension_sugar` uses? It is very hard to prevent ambiguities with this sugar.

let-def:
No it is not on top of @SanderSpies change.

let-def:
I will look at that, I suspect that putting attributes on top introduce more opportunities for ambiguities.

let-def:
About `item_extension_sugar`, the comment below (found in the parser), makes me want to skip it (I did a few attempt at implementing it unambiguously, but it ends up making the grammar very ugly):
```
   * Note, this form isn't really super useful, but wouldn't cause any parser
   * conflicts. Not supporting it though just to avoid having to write the
   * pretty printing logic.
   *
   *   [@attrsOnExtension] %extSugarId [@attrOnLet] LET ..
   *
   * We won't document it though, and probably won't format it as such.
   *  | PERCENT attr_id item_attribute post_item_attributes {
   *     ($3::$4, $2)
   *    }
```

jordwalke:
Yeah that's fine. I think that sugar belongs as part of sander's effort.

jordwalke:
It's probably best to do this on top of your major refactor, which includes sander's changes. Is the diff stack similar enough to keep this PR open?

let-def:
I don't think so. Sander's changes already make it harder to remove semicolons and with my patch on top I guess a different approach is needed. I close this for now.

PR #1236, opened 2017-04-25
ES6 style function

let-def:
This patch adds a new syntactic sugar for functions:
`(a) => ...` is `fun a => ...`
`(a,b) => ...` is `fun a b => ...`
`() => ...` is `fun () => ...`

The updated testsuite shows how that looks like.
I will add a few improvements to the printer, for instance `(a : int, b : int) => a + b` is printed as `((a : int), (b : int)) => a + b`.

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/jsx.re
+++ b/formatTest/typeCheckedTests/expected_output/jsx.re
@@ -207,7 +207,7 @@ let selfClosing3 =
     b="cause the entire thing to wrap"
   />;
 
-let a = <Foo> <Bar c=(fun a => a + 2) /> </Foo>;
+let a = <Foo> <Bar c=((a) => a + 2) /> </Foo>;
 
 let a3 = <So> <Much> <Nesting /> </Much> </So>;
 
@@ -463,7 +463,7 @@ let asd2 = video test::false 10 [@JSX] [@foo];
 
 let div ::children => 1;
 
-((fun () => div) ()) children::[] [@JSX];
+(((()) => div) ()) children::[] [@JSX];
 
 let myFun () =>
   <>
--- a/formatTest/unit_tests/expected_output/wrappingTest.re
+++ b/formatTest/unit_tests/expected_output/wrappingTest.re
@@ -263,7 +263,7 @@ let explictlyPassed =
     /* None; */
     b::?None;
 
-let complex_default ::callback=(fun k d => 4) x => 3;
+let complex_default ::callback=((k, d) => 4) x => 3;
 
 let myList = /*CommentAfterEqualBeforeList */ [
   1,
@@ -691,24 +691,26 @@ let echoTheEchoer
 
 /* Nothing annotated fun, passed to func */
 echoTheEchoer (
-  fun (
-        a,
-        b,
-        c,
-        d,
-        e,
-        f,
-        g,
-        h,
-        i,
-        j,
-        k,
-        l,
-        m,
-        n,
-        o,
-        p
-      ) => (
+  (
+    (
+      a,
+      b,
+      c,
+      d,
+      e,
+      f,
+      g,
+      h,
+      i,
+      j,
+      k,
+      l,
+      m,
+      n,
+      o,
+      p
+    )
+  ) => (
     a,
     b,
     c,
@@ -1885,9 +1887,9 @@ let (the, type_, and_, value, should, both, wrap): (
   "wrap"
 );
 
-let myPolyFunc: 'a .'a => 'a = fun o => o;
+let myPolyFunc: 'a .'a => 'a = (o) => o;
 
-let myNonPolyFunc: 'a => 'a = fun o => o;
+let myNonPolyFunc: 'a => 'a = (o) => o;
 
 let locallyAbstractFunc (type a) (input: a) => input;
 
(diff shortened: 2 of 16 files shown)


jordwalke:
Would we want to make function invocation and variant / type parameter syntax match in the same diff?

ForbesLindesay:
For the printer, it would better match how people write JavaScript to do:

```js
(a: int, b: int) => a + b
```

rather than

```js
(a : int, b : int) => a + b
```

In general though, this is a really exciting improvement in the syntax :)

let-def:
The same syntax for application introduces a few problems:
- `simple_expr_list` used by jsx in the grammar will need to be delimited (probably with comma)
- `while`, `for`, `if` construct will likely need parentheses (to disambiguate `if f(x)`: `if f x` or `if (f(x)) ...`).

jordwalke:
"simple_expr_list used by jsx in the grammar will need to be delimited (probably with comma)"

Why would that be the case?

let-def:
`x (y) z` is parsed as `(x y) (z)`

jlongster:
Just chiming in to say that I'd love to require parens when invoking function calls. I think that's going to be the biggest blocker for Reason adoption. Though I am aware it's quite an invasive change.

For example, this surprised me the other day: `(Config.get "env" ^ "/item/public_token/exchange")`. The string concat operator is actually lower precedence so that's equal to `((Config.get "env") ^ "/item/public_token/exchange")`, which is really surprising. There are lots of little cases like that where forcing parens makes it a lot clearer: `Config.get("env") ^ "/item/public_token/exchange"`.

Maybe that should be taken up in another PR though?

let-def:
@jlongster The `^` makes sense to me, but that's arbitrary, it's just an habit. But I agree with the goal: I would like to require parens for function calls asap.

jordwalke:
Requiring parens for function calls should probably go in the same diff/release as this PR. That would be best to avoid thrash.

formatTest/typeCheckedTests/expected_output/jsx.re
```diff
@@ -463,7 +463,7 @@ let asd2 = video test::false 10 [@JSX] [@foo];
 
 let div ::children => 1;
 
-((fun () => div) ()) children::[] [@JSX];
+(((()) => div) ()) children::[] [@JSX];
```
jaredly: would it be hard to have this just be `() => div`?

formatTest/unit_tests/expected_output/wrappingTest.re
```diff
@@ -1940,7 +1942,7 @@ let df_locallyAbstractFuncNotSugared
  */
 let df_locallyAbstractFuncAnnotated:
   type a. a => a => inputEchoRecord a =
-  fun (input: a) (input: a) => {inputIs: input};
+  ((input: a), (input: a)) => {inputIs: input};
```
jaredly: can we now drop the parenthesis around type annotations?

jaredly:
Sorry for the noob question, but I checked out this branch, ran `make build`, and tried out `./refmt_impl.native` on a file containing `let x = (a) => a + 1;`, and it failed w/ a syntax error. Also `make test` has a bunch of errors where the pretty printer isn't doing the es6 style.
Am I missing a step?

jordwalke:
@jaredly there's also a parser directory inside the repo. Was that correctly built?

formatTest/unit_tests/expected_output/wrappingTest.re
let-def: Yes, the printer doesn't optimize yet

formatTest/typeCheckedTests/expected_output/jsx.re
let-def: Yes, printer didn't optimize at the time.

chenglou:
Incorporated into #1299

freebroccolo:
I suppose it's a bit late to comment on this now but one issue with `(a) => ...` vs `fun a => ...` is that the former isn't possible to highlight accurately with textmate grammars in general.

The problem is that you can't know whether `(a)` is a value or a pattern without looking ahead until you find a `=>` or some terminating token like `)`, `]`, `}`, `;`, etc.

As long as `=>` is on the same line as the opening `(` you can *probably* highlight it correctly with some effort but there's still the issue that the highlighting cannot be determined until the `=>` is typed into the buffer. And if the `=>` is not on the same line, the highlighting will just be wrong (it will show as a value).

In general, any syntax that requires looking ahead arbitrarily in order to determine the current parsing context is a problem for textmate grammars because there is no way to look beyond the end of the current line before making a choice.

This is an unfortunate limitation but I've come around to the view that it's an informative constraint as far as the language design is concerned because it also has an impact on how responsive and informative the diagnostic feedback (e.g., type inference, etc.) can be as you are editing.

jaredly:
how does javascript's highlighter work?

freebroccolo:
There are several different javascript highlighters implemented as textmate grammars and they all behave slightly differently. Last I checked, I believe the ones from the textmate repos, atom, sublime, and vscode are all maintained separately and give different results. And then there are several 3rd party ones available on the vscode extension marketplace and probably elsewhere too.

But none of the javascript highlighters I've ever seen have tried to be accurate down to the point where they are parsing the actual syntax like the reason one does. Mostly they just use approximate highlighting and tend to fail on certain corner cases, usually involving multiline syntax fragments. Of course the reason highlighter still fails in some cases too but that's due to bugs in the rules rather than through a conscious choice to parse approximately to begin with.

Javascript also doesn't have as many clear cut syntactic categories as reason does, so in their case this ambiguity that arises between patterns and values isn't really an issue to begin with. That's why there's this sort of loss of fidelity in removing the `fun`.

I suppose it may be an acceptable tradeoff to just try and do more approximate parsing in the reason highligher and give up on accuracy. But to do that, some of the stuff that is currently distinguished now will probably need to be simplified in order to appear consistent. That may have other consequences though. For instance, types can occur in patterns too…

One thing I've found is that once you start to make the highlighting parser more approximate, it generally becomes more difficult to maintain because it's harder to keep track of which cases need to be fudged in order to get good results. There's no way I know of to do this without a lot of hand tuning.

PR #1249, opened 2017-05-09
Revise clone command for installs to work

ev1stensberg:
I had to use the direct clone from github in order to make the clone work. This will make the instructions work, as SSH is limited for some use cases?

skjermbilde 2017-05-09 kl 11 13 57

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ opam init
 opam update
 opam switch 4.03.0
 eval $(opam config env)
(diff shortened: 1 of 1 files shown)


jordwalke:
Hmm, what are the cases where that command fails? I can use the `git@` version for any repo on github even if I don't have rights to the repo. I suppose we can change this though.

PR #1250, opened 2017-05-09
[Docs] New emacs support

chenglou:


chenglou:
Thanks!

docs/tools.html
```diff
@@ -219,8 +219,8 @@
```
jordwalke: Noticed the vim plugin doesn't point to your new repo. Want to transfer to reasonml and we can update the docs for Vim too?

PR #1254, opened 2017-05-11
Remove our own emacs plugin

chenglou:
cc @arichiardi is this fine? Just removing our own files.

cc @sgrove too


arichiardi:
Good for me!

PR #1256, opened 2017-05-11
Improve README readability

hughrawlinson:
It wasn't immediately clear to me that these two links were different - I put them in a list like below to make it more apparent.


radiodario:
lol

Issue #1066, opened 2017-02-15
refmt long chain of simple |> doesn't break at all


kyldvs:
`formatTest/customMLFiles/kad.re`:
```
let x = foo |> somelongfunctionname "foo" |> anotherlongfunctionname "bar" 1 |> somelongfunction |> bazasdasdad;
```
`_build/src/refmt_impl.native formatTest/customMLFiles/kad.re --print-width 80` results in on master:
```
let x =
  foo |> somelongfunctionname "foo" |> anotherlongfunctionname "bar" 1 |> somelongfunction |> bazasdasdad;
```

bobzhang:
I would like it to see it formatted as:

```re
let x = 
    foo
    |> f 
    |> g 
    |> x;
```

jordwalke:
I like that breaking too. It might be easier to just do this for *all* infix, if it looks okay.

hcarty:
@jordwalke What would that mean for `1 + 1`?

jordwalke:
@hcarty There are two issues right now with `|>` that I can see. The first is that it's not breaking at the correct width, and then (as Bob suggested) when it does break, we want it to break differently (aligned to a consistent vertical column). So for addition, the formatting could be just like `|>`, or at least configurable to be like `|>`.

hcarty:
@jordwalke I see, I read the suggestion as 'break at every `|>` which makes a lot of sense for `|>` and maybe the monadic infix operators but wouldn't play well with most others.

IwanKaramazow:
I'm taking a look atm

kyldvs:
@IwanKaramazow depending on how deep you get into it #1045 also seems related.

jordwalke:
@IwanKaramazow What did you find with this formatting challenge? Bob's example of formatting would be great for all infix identifiers:

```
let x = 
    foo
    |> f 
    |> g 
    |> x;
```

Did you see a specific challenge that I can help with?

kyldvs:
@IwanKaramazow thanks for taking a look! Do you have any updates, this is my single biggest issue with refmt at the moment.

This was my attempt at a fix in #1045 but couldn't quite get it right: https://github.com/facebook/reason/compare/master...kyldvs:pipe-chaining?expand=1

IwanKaramazow:
The challenge is mainly in generalizing to all infix operators. `>>=`& family is parsed differently than `|>`. The first is a 'right-fold', while the latter is a 'left-fold' in terms of how the ast looks.
[...]

Issue #1045, opened 2017-02-07
refmt issues with chained |>


kyldvs:
When chaining `|>` refmt makes some weird decisions that I believe to be undesirable formatting. I start with:

```
let code = JSCodegen.Code.(
  create
  |> lines Requires.(
    create
    |> import_type local::"Set" source::"Set"
    |> import_type local::"Map" source::"Map"
    |> import_type local::"Immutable" source::"immutable"
    |> require local::"invariant" source::"invariant"
    |> require local::"Image" source::"Image.react"
    |> side_effect source::"monkey_patches"
    |> render_lines
  )
  |> new_line
  |> new_line
  |> new_line
  |> new_line
  |> render
);
```

And after formatting end with:

```
let code =
  JSCodegen.Code.(
    create |>
    lines
      Requires.(
        create |> import_type local::"Set" source::"Set" |>
        import_type local::"Map" source::"Map" |>
        import_type local::"Immutable" source::"immutable" |>
        require local::"invariant" source::"invariant" |>
        require local::"Image" source::"Image.react" |>
        side_effect source::"monkey_patches" |> render_lines
      ) |> new_line |> new_line |> new_line |> new_line |> render
  );

```

I believe there are 3 distinct issues with this new format:

1. There shouldn't need to be a break between `let code =` and `JSCodegen.Code.(`. This just makes all the code below it more likely to break lines and wastes horizontal space. I believe this also causes the break between `|> lines` and `Requires.(`, but that could be a separate 4th issue as well. (This appears to have been an unintended regression in https://github.com/facebook/reason/pull/812)
2. Once a single `|>` breaks in a chain, I believe they should all be on a separate line. For example there should be a break after `create |>` or more noticeably after each `new _line |>`
3. I think `|>` should start the line rather than end it. Admittedly I could see this being more of a preference thing, but being at the start feels more consistent with this kind of formatting:

```
type foo =
  | Bar string
  | Baz string
  | Buz string;
```

kyldvs:
More support for (3) in the Reason source code itself :) (although .ml not a .re file...)

https://github.com/facebook/reason/blob/master/src/reason_format_type.ml#L24-L36

kyldvs:
Here's my attempt at making some progress on this, probably a good place to start from (or maybe not, I have no idea :P)

https://github.com/facebook/reason/compare/master...kyldvs:pipe-chaining?expand=1

This:

```
let code = JSCodegen.Code.(
  create
  |> lines Requires.(
    create
[...]

PR #1259, opened 2017-05-12
Pretty printing of expression applications with infix operators

IwanKaramazow:
Improve printing of chained |> (#1066, #1045)

### Result
```reason
let x = foo |> z;

let x =
  foo
  |> f
  |> g;

let x =
  foo
  |> somelongfunctionname "foo"
  |> anotherlongfunctionname "bar" 1
  |> somelongfunction
  |> bazasdasdad;

let code =
  JSCodegen.Code.(
    create
    |> lines
         Requires.(
           create
           |> import_type local::"Set" source::"Set"
           |> import_type local::"Map" source::"Map"
           |> import_type local::"Immutable" source::"immutable"
           |> require local::"invariant" source::"invariant"
           |> require local::"Image" source::"Image.react"
           |> side_effect source::"monkey_patches"
           |> render_lines
         )
    |> new_line
    |> new_line
    |> new_line
    |> new_line
    |> render
  );

let code = JSCodegen.Code.(create |> render); 
```

### Overview 
```reason
/* (1) */
let x = foo |> f; 

/* (2) */
let x = 
  foo
  |> f
  |> g;
```

We have two cases here: the first shouldn't break, the second should always break.
Note that `|>` results in some sort of recursive structure 'going deeper' on the 'left side' of the printed result.
The printing is some sort of left fold.
```
let x =  foo |> f |> g;

/* is going to be parsed as: */
((foo |> f) |> g)
````
This is important because at a certain moment in the unparsing recursion,
we're going to see `foo |> f`. This is exactly the same as the first case!
If we just recursively print everything where the last one is treated as the first case,
we would get:
```
let x =
  foo |> f
  |> g;
```
How do we make a difference between the two cases? 

```ocaml
(* list containing all infix operators with the 'left-fold' printing behaviour *)
let leftRecInfixOperators = ["|>"]
```
While unparsing an expression, check if a printedIdent is a member of the above list.
If so, we check if the left child of the application arguments contains another
application with the same printedIdent (`isLeftInfixChain`) 
```
Is (foo |> f) another Pexp_apply with a |> in ((foo |> f) |> g) ?
Yes, it is.
```
If so mark the unparsing with the parameter `~infixChain`. 
`~infixChain` indicates we're chaining an application with an identifier ("|>" here).
Based on the value of `~infixChain`, we know we always have to break.
The innermost (foo |> f) of ((foo |> f) |> g), will break and result in the following:
```reason
let x =
  foo
  |> f
  |> g;

/* we won't get 
let x =
  foo |> f
  |> g;
*/
```

### Future work
This PR works as advertised for `|>`, but it would be nice if we could standardize
printing for all infix operators. `>>=` & co has the same problem, but prints as a right fold instead
of a left fold. (`>>=` has some too-much-parens problems too).
Can we drop the `leftRecInfixOperators` in the future? Is there a way to detect those?
Note that we can't just take all infix operators with left associativity here, because of e.g. "1 + 2 + 3"
Printing all left associative infix application as `|>` would result in:
```
let x =
  1
  + 2
  + 3
```
With `|>`, it always breaks as soon as there's more than one `|>` chained.
In the case of `+`, we want break `IfNeed` style.

### DIFF
--- a/formatTest/unit_tests/expected_output/infix.re
+++ b/formatTest/unit_tests/expected_output/infix.re
@@ -155,10 +155,16 @@ first || second || third;
 
 /* No parens should be added/removed from the following when formatting */
 let seeWhichCharacterHasHigherPrecedence =
-  (first |> second |> third) ^> fourth;
+  (
+    first
+    |> second
+    |> third
+  ) ^> fourth;
 
 let seeWhichCharacterHasHigherPrecedence =
-  first |> second |> third;
+  first
+  |> second
+  |> third;
 
 let seeWhichCharacterHasHigherPrecedence =
   first + second + third;
@@ -245,7 +251,10 @@ let res = (|>) first;
 /* Infix has lower precedence than function application */
 first |> second ::third;
 
-let leftAssocGrouping = first |> second |> third;
+let leftAssocGrouping =
+  first
+  |> second
+  |> third;
 
 let rightAssocGrouping =
   first ^> second ^> third;
@@ -948,3 +957,49 @@ let containingObject = {
       "newvalue" [@shouldBeRenderedOnString]
   }
 };
+
+let x = foo |> z;
+
+let x =
+  foo
+  |> f
+  |> g;
+
+let x =
+  foo
+  |> somelongfunctionname "foo"
+  |> anotherlongfunctionname "bar" 1
+  |> somelongfunction
+  |> bazasdasdad;
+
+let code =
+  JSCodegen.Code.(
+    create
+    |> lines
+         Requires.(
+           create
+           |> import_type
+                local::"Set" source::"Set"
+           |> import_type
+                local::"Map" source::"Map"
+           |> import_type
+                local::"Immutable"
+                source::"immutable"
+           |> require
+                local::"invariant"
+                source::"invariant"
+           |> require
+                local::"Image"
+                source::"Image.react"
+           |> side_effect
+                source::"monkey_patches"
+           |> render_lines
+         )
+    |> new_line
+    |> new_line
+    |> new_line
+    |> new_line
+    |> render
+  );
+
+let code = JSCodegen.Code.(create |> render);
--- a/formatTest/unit_tests/expected_output/infix.re
+++ b/formatTest/unit_tests/expected_output/infix.re
@@ -313,10 +313,10 @@ let includesACommentCloseInIdentifier = ( *\*\/ );
 let shouldSimplifyAnythingExceptApplicationAndConstruction =
   call "hi"
   ^ (
-      switch x {
-      | _ => "hi"
-      }
-    )
+    switch x {
+    | _ => "hi"
+    }
+  )
   ^ "yo";
 
 /* Add tests with IF/then mixed with infix/constructor application on left and right sides */
@@ -1009,15 +1009,14 @@ let server = {
     body
     |> Cohttp_lwt_body.to_string
     >|= (
-          fun body =>
-            Printf.sprintf
-              "okokok" uri meth headers body
-        )
+      fun body =>
+        Printf.sprintf
+          "okokok" uri meth headers body
+    )
     >>= (
-          fun body =>
-            Server.respond_string
-              ::status ::body ()
-        )
+      fun body =>
+        Server.respond_string ::status ::body ()
+    )
   };
   Server.create
     ::mode (Server.make ::callback ())
--- a/formatTest/unit_tests/expected_output/infix.re
+++ b/formatTest/unit_tests/expected_output/infix.re
@@ -155,16 +155,10 @@ first || second || third;
 
 /* No parens should be added/removed from the following when formatting */
 let seeWhichCharacterHasHigherPrecedence =
-  (
-    first
-    |> second
-    |> third
-  ) ^> fourth;
+  (first |> second |> third) ^> fourth;
 
 let seeWhichCharacterHasHigherPrecedence =
-  first
-  |> second
-  |> third;
+  first |> second |> third;
 
 let seeWhichCharacterHasHigherPrecedence =
   first + second + third;
@@ -173,12 +167,16 @@ let comparison = (==);
 
 /* Why would the following two cases have different grouping? */
 let res =
-  blah ||
-  DataConstructor 10 || DataConstructor 10 && 10;
+  blah
+  || DataConstructor 10
+  || DataConstructor 10
+  && 10;
 
 let res =
-  blah &&
-  DataConstructor 10 && DataConstructor 10 + 10;
+  blah
+  && DataConstructor 10
+  && DataConstructor 10
+  + 10;
 
 /* This demonstrates how broken infix pretty printing is:
  */
@@ -251,10 +249,7 @@ let res = (|>) first;
 /* Infix has lower precedence than function application */
 first |> second ::third;
 
-let leftAssocGrouping =
-  first
-  |> second
-  |> third;
+let leftAssocGrouping = first |> second |> third;
 
 let rightAssocGrouping =
   first ^> second ^> third;
@@ -270,13 +265,17 @@ let seeWhichCharacterHasHigherPrecedence =
   first ^> (second |> third) |> fourth;
 
 let res =
-  blah &&
-  DataConstructor 10 && DataConstructor 10 + 10;
+  blah
+  && DataConstructor 10
+  && DataConstructor 10
+  + 10;
 
 /* Should be parsed as */
 let res =
-  blah &&
-  DataConstructor 10 && DataConstructor 10 + 10;
+  blah
+  && DataConstructor 10
+  && DataConstructor 10
+  + 10;
 
 let (++) ::label ::label2 => label + label2;
 
@@ -312,12 +311,13 @@ let includesACommentCloseInIdentifier = ( *\*\/ );
 let includesACommentCloseInIdentifier = ( *\*\/ );
 
 let shouldSimplifyAnythingExceptApplicationAndConstruction =
-  call "hi" ^
-  (
-    switch x {
-    | _ => "hi"
-    }
-  ) ^ "yo";
+  call "hi"
+  ^ (
+      switch x {
+      | _ => "hi"
+      }
+    )
+  ^ "yo";
 
 /* Add tests with IF/then mixed with infix/constructor application on left and right sides */
 
@@ -512,9 +512,7 @@ let containingObject = {
     x + something.contents := y;
     x + something := y;
     /* Should be parsed as: */
-    x + (
-      something.contents = y
-    ); /* Because of the #NotActuallyAConflict above */
+    x + (something.contents = y); /* Because of the #NotActuallyAConflict above */
     x + (something = y); /* Same */
     x + something.contents := y;
     x + something := y;
@@ -539,14 +537,12 @@ let containingObject = {
     bigArr.{0} || something + 1 ?
       hello : goodbye;
     str.[0] || something + 1 ? hello : goodbye;
-    x.contents || (
-      something + 1 ? hello : goodbye
-    );
+    x.contents
+    || (something + 1 ? hello : goodbye);
     y || (something + 1 ? hello : goodbye);
     arr.(0) || (something + 1 ? hello : goodbye);
-    bigArr.{0} || (
-      something + 1 ? hello : goodbye
-    );
+    bigArr.{0}
+    || (something + 1 ? hello : goodbye);
     str.[0] || (something + 1 ? hello : goodbye);
 
     /**
@@ -566,14 +562,12 @@ let containingObject = {
     bigArr.{0} && something + 1 ?
       hello : goodbye;
     str.[0] && something + 1 ? hello : goodbye;
-    x.contents && (
-      something + 1 ? hello : goodbye
-    );
+    x.contents
+    && (something + 1 ? hello : goodbye);
     y && (something + 1 ? hello : goodbye);
     arr.(0) && (something + 1 ? hello : goodbye);
-    bigArr.{0} && (
-      something + 1 ? hello : goodbye
-    );
+    bigArr.{0}
+    && (something + 1 ? hello : goodbye);
     str.[0] && (something + 1 ? hello : goodbye);
 
     /**
@@ -665,8 +659,7 @@ let containingObject = {
         str.[0] = somethingElse : goodbye;
 
     /** And this */
-    y :=
-      something ? y := somethingElse : goodbye;
+    y := something ? y := somethingElse : goodbye;
     arr.(0) :=
       something ?
         arr.(0) := somethingElse : goodbye;
@@ -677,8 +670,7 @@ let containingObject = {
       something ?
         str.[0] := somethingElse : goodbye;
     /* Should be parsed as */
-    y :=
-      something ? y := somethingElse : goodbye;
+    y := something ? y := somethingElse : goodbye;
     arr.(0) :=
       something ?
         arr.(0) := somethingElse : goodbye;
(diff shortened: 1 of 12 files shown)


formatTest/unit_tests/expected_output/infix.re
```diff
@@ -948,3 +957,49 @@ let containingObject = {
...
+  |> bazasdasdad;
+
+let code =
+  JSCodegen.Code.(
+    create
+    |> lines
+         Requires.(
+           create
+           |> import_type
+                local::"Set" source::"Set"
```
hcarty: Why do these labeled arguments break onto a new line?

IwanKaramazow: Because of the printing width of the tests.
See Example section in my PR description for the default printing width.

hcarty: I see - thank you for the explanation.  The discrepancy between the PR description and the expected output is what threw me.

jordwalke:
>  `/* is going to be printed as: */`

I think you mean "parsed" as?

jordwalke:
What did you mean by this?

> "Note that we can't just take all infix operators with left associativity here, because of e.g. "1 + 2 + 3"

By the way, this is a very good explanation of the problem/challenges.

reason-parser/src/reason_pprint_ast.ml
```diff
@@ -3375,8 +3399,8 @@ class printer  ()= object(self:'self)
     | PotentiallyLowPrecedence itm -> formatPrecedence ~loc:reducesAfterRight.pexp_loc itm
     | Simple itm -> itm
 
-  method ensureExpression expr ~reducesOnToken =
-    match self#unparseExprRecurse expr with
+  method ensureExpression ?(infixChain=false) ~reducesOnToken expr =
+    match self#unparseExprRecurse ~infixChain expr with
     | SpecificInfixPrecedence ({reducePrecedence; shiftPrecedence}, leftRecurse) ->
```
jordwalke: This would likely be easier to implement this change, as well as all the other customizations you mentioned, if the result of unparsing returned an intermediate representation of the *concrete* AST. That is, it would return a tree where all the precedence has already been resolved. Imagine if you had something like this:

```
InfixApplicationConcrete (
  "|>",
  InfixApplicationConcrete ( "|>", SimpleConcrete itm),
  SimpleConcrete itm
)
```

Where printing is trivial, and all parenthesis/precedence has already been handled.
Then you can easily do things like
```
computeConsecutiveLeftInfixChains (concreteAst)
```

which would return a convenient list for you to format in a specific way.

This division separates all the technical details of precedence and *correctness* from the opinionated *styling* of formatting.

Maybe the current `FunctionApplication`/`Simple`/`SpecificInfixPrecedence` could *become* that Concrete AST I mentioned? It's pretty close to it.

IwanKaramazow:
> "Note that we can't just take all infix operators with left associativity here, because of e.g. "1 + 2 + 3"

I thought of printing all left associative infix application as `|>`, but clearly this won't with `+` for example.
```
let x =
  1
  + 2
  + 3
```
With `|>`, it always breaks as soon as there's more than one `|>` chained.
In the case of `+`, we want break `IfNeed` style.

reason-parser/src/reason_pprint_ast.ml
IwanKaramazow: This is a very good idea. An intermediate representation should solve the nuances of printing different infix applications. `~infixChain` felt very shaky from the start, I just couldn't think of something better at the moment 😄
I'll go ahead and implement it.

jordwalke: Great! I think we could/should start extending that pattern through the rest of the printer. Imagine having a completely separate stage to manage all the stylistic/opinionated sections - separate from a stage that determines *correct* grouping according to precedence. It will be much easier for people to contribute fixes to stylistic formatting.

jordwalke:
For typical use cases of `|>` would it be bad to format on an `IfNeed` basis? I genuinely don't know.

I don't mind `+` wrapping like this:

```
let x =
  1
  + 2
  + 3
```

as long as it's on an `IfNeed` basis.

Also, I believe that `>>=` is actually left associative right? If so, then it is one case where we *don't* want the infix to dock to the left, despite being left associative.

IwanKaramazow:
Should I change it to an `IfNeed` basis?
```reason
let uri = req |> Request.uri |> Uri.to_string;
let meth = req |> Request.meth |> Code.string_of_method;
let headers = req |> Request.headers |> Header.to_string;
```
In the above it absolutely makes sense, but when I'm writing
```reason
let nextState =
    state
    |> updateFlappy
    |> updateTime;
```
I want it to break. Hmm, food for thought.

jordwalke:
I don't feel strongly. @bordoley, our resident `|>` connoisseur may have opinions.

bordoley:
@IwanKaramazow why do you want line breaks in your nextState case? 

Looking through my own code, I think the IfNeed basis would accurately layout most code the same way I would by hand.

IwanKaramazow:
In the `nextState` case it's just for readability, and there's a big chance I might extend it. It would eventually break. The `IfNeed` basis might be harder since the `Auto` breaking of labels here originally didn't work & spawned the issue 😄
Back to the drawing board...

jordwalke:
The original issue here which was caused by `IfNeed` was due to the fact that they were multiple nested labels. I had originally had something kind of like the "concrete parse tree" representation that we just discussed, but I removed it in favor of this implementation (labels) to first focus on correctness, with the intent of eventually evolving it back into a kind of "concrete parse tree".

chenglou:
@IwanKaramazow ready? =D

IwanKaramazow:
It works, I just need to rename some functions so it makes more sense. I can do that this weekend.
@jordwalke should I wait & do this on top of Fred's new diff?

chenglou:
I think this should get in before the breaking change, since current users can benefit

kyldvs:
Would love to see this get merged, this is my biggest issue with the formatter at the moment :)

jordwalke:
I want to make sure @let-def is aware of this one.

reason-parser/src/reason_pprint_ast.ml
```diff
@@ -125,6 +125,7 @@ and ruleCategory =
      that it's easier just to always wrap them in parens.  *)
   | PotentiallyLowPrecedence of layoutNode
```
jordwalke: @let-def Not urgent, but in case you stumble upon this: this `ruleCategory` type is the closest thing to an intermediate representation that we already have to work with. You could decide to evolve it to be more sophisticated, or start fresh eventually.

reason-parser/src/reason_pprint_ast.ml
```diff
@@ -125,6 +125,7 @@ and ruleCategory =
...
   | Simple of layoutNode
+  | InfixApplicationConcrete of ruleInfoData * string * ruleCategory * ruleCategory
```
jordwalke: We discussed in chat, that this would likely take the place of `SpecificInfixPrecedence`. Do you still believe this is the case? Is still seems there's a large overlap in their purposes.

reason-parser/src/reason_pprint_ast.ml
```diff
@@ -579,9 +584,21 @@ let special_infix_strings =
 let updateToken = "="
 let requireIndentFor = [updateToken; ":="]
 
+(*
+ * list containing all infix operators which exhibit a 'left-fold' printing behaviour
```
jordwalke: This comment may be out of date, is it?

reason-parser/src/reason_pprint_ast.ml
```diff
@@ -617,6 +634,17 @@ let isSimplePrefixToken s = match printedStringAndFixity s with
   | AlmostSimplePrefix _ -> true
   | _ -> false
 
+(*
+ * function to determine if the left child of an expression's application arguments
+ * is another Pexp_apply with the given ident
+ *)
+let isLeftInfixChain ident = function
```
jordwalke: This may no longer be used, is it?

reason-parser/src/reason_pprint_ast.ml
```diff
@@ -3364,39 +3437,60 @@ class printer  ()= object(self:'self)
      token "+", in `x + a * b` should not reduce until after the a * b gets
      a chance to reduce. This function would determine the minimum parens to
      ensure that. *)
-  method ensureContainingRule ~withPrecedence ~reducesAfterRight =
+  method ensureContainingRule ~withPrecedence ~reducesAfterRight () =
     match self#unparseExprRecurse reducesAfterRight with
     | (SpecificInfixPrecedence ({reducePrecedence; shiftPrecedence}, rightRecurse)) ->
-      if higherPrecedenceThan shiftPrecedence withPrecedence then rightRecurse
+      if higherPrecedenceThan shiftPrecedence withPrecedence then begin 
+        Simple rightRecurse
```
jordwalke: I think this is a misnomer to wrap in `Simple`. This function `ensureContainingRule` ensures that `reducesAfterRight` will reduce *before* the containing rule with `withPrecedence` reduces. I realize that's pretty confusing and I have to read that sentence several times to make sense of it.

I think a way I could have wrir way to write the function definition to make it more clear would have been:

```
method ensureContainingRule ~withPrecedence:precedence ~reducesAfterRight:rightExpr () =
```

But the point of the function is to ensure that `rightExpr` will reduce at the proper time when it is reparsed, possibly wrapping it in parenthesis if needed. But that doesn't mean it is necessarily *simple*. For example, in `4 + 5*3`, `5*3` will reduce before the `+` operation, yet `5*3` is not "simple". Simple means it is clearly one token (such as `(anything)` or `[anything]` or `identifier`).

reason-parser/src/reason_pprint_ast.ml
```diff
@@ -3364,39 +3437,60 @@ class printer  ()= object(self:'self)
      token "+", in `x + a * b` should not reduce until after the a * b gets
      a chance to reduce. This function would determine the minimum parens to
      ensure that. *)
-  method ensureContainingRule ~withPrecedence ~reducesAfterRight =
+  method ensureContainingRule ~withPrecedence ~reducesAfterRight () =
```
IwanKaramazow: Yes, it has become something I always write. It makes it so much easier when revisiting/extending the code at a later point. No point in making life harder.

reason-parser/src/reason_pprint_ast.ml
IwanKaramazow: I'll revisit this tomorrow.

chenglou:
This got out of sync since we've re-merged back reason-parser. Would be great to have it working again! Number one request for formatting right now =)

IwanKaramazow:
Rebased on master, now continuing

chenglou:
This should be something we get in before the potential function syntax refactor. cc @let-def just in case

TheSpyder:
What ended up happening with infix operators like `>>=`? I can't quite follow what happens to them in the change list.

While I agree it can't be "all left associative infix operators", any infix operator that's more than one character is _probably_ worth applying this to.

IwanKaramazow:
@TheSpyder 
`>>=` should be printed a lot better. Check the result of some classic cohttp code:
```
let server = {
  let callback _conn req body => {
    let uri =
      req
      |> Request.uri
      |> Uri.to_string
      |> Code.string_of_uri
      |> Server.respond
      |> Request.uri;
    let meth =
      req
      |> Request.meth
      |> Code.string_of_method;
    let headers =
      req |> Request.headers |> Header.to_string;
    body
    |> Cohttp_lwt_body.to_string
    >|= (
          fun body =>
            Printf.sprintf
              "okokok" uri meth headers body
        )
    >>= (
          fun body =>
            Server.respond_string
              ::status ::body ()
        )
  };
  Server.create
    ::mode (Server.make ::callback ())
};
```

hcarty:
The parens are unfortunate but it sounds like they're pretty difficult to remove.

@IwanKaramazow This is so much better that the current output - thank you!

TheSpyder:
This is getting a bit off topic for the pipe operators, so I'll log a new task if you'd prefer, but in the case of infix operators could we avoid the indent inside parens? Or perhaps just when the brackets contain a single function?

e.g.
```
>|= (fun body =>
      Printf.sprintf
        "okokok" uri meth headers body
    )
```

IwanKaramazow:
@TheSpyder I addressed the weird indentation. 😁

formatTest/unit_tests/expected_output/infix.re
```diff
@@ -948,3 +947,77 @@ let containingObject = {
...
+      |> Request.uri;
+    let meth =
+      req
+      |> Request.meth
+      |> Code.string_of_method;
+    let headers =
+      req |> Request.headers |> Header.to_string;
+    body
+    |> Cohttp_lwt_body.to_string
+    >|= (
```
bsansouci: Does your latest commit fix the `fun body` being on a new line?

bsansouci: Just asking, I think this PR is amazing and we should merge it before tackling on other fun problems <3

IwanKaramazow: The tests have a ridiculously short print-width of 50.
If you have something like 80 you'll get:
```
body
    |> Cohttp_lwt_body.to_string
    >|= (fun body => Printf.sprintf "okokok" uri meth headers body)
    >>= (fun body => Server.respond_string ::status ::body ())
```
What has been fixed is the closing `)` indenting on the same height as the beginning of `>|=` and `>>=`. :)

src/refmt_args.ml
```diff
@@ -42,7 +42,7 @@ let print =
 let print_width =
   let docv = "COLS" in
   let doc = "wrapping width for printing the AST" in
-  Arg.(value & opt (int) (100) & info ["w"; "print-width"] ~docv ~doc)
+  Arg.(value & opt (int) (90) & info ["w"; "print-width"] ~docv ~doc)
```
chenglou: Why is this 90 now?

chenglou:
Soooo, which one should we merge first, @let-def's function diff, or this one? =P

hcarty:
@chenglou This one! Fixes some significant paper cuts before opening up a world of fun from overhauling the entire function and constructor syntax. It will also (hopefully) keep the conversation around the big @let-def function change focused on just that change.

src/refmt_args.ml
IwanKaramazow: `reason_pprint_ast.ml` contains `defaultSettings` 90.
Not sure why I automatically changed it here to 90 lol.
I can change it back, your call.

IwanKaramazow:
@chenglou just checked, this PR & the js-syntax-branch shouldn't cause any crazy merging conflicts. Had to touch some parts of the code let-def didn't touch ⛷

chenglou:
Alright I've thought about this. We're gonna do right by the community and make sure you don't feel conned into using the function application PR just for better pipes lol.

So I'll merge this PR, make a release, before we get crazy with the function diff. cc @let-def, I'll re-rebase your diff on top of this one. This way everyone's happy.

As always, thanks @IwanKaramazow!

let-def:
Thanks @chenglou!

formatTest/unit_tests/input/infix.re
```diff
@@ -755,3 +755,42 @@ let containingObject = {
     something.contents = "newvalue" [@shouldBeRenderedOnString];
   };
 };
+
+let x = foo |> z;
+
+let x = foo |> f |> g;
+
+let x = foo |> somelongfunctionname "foo" |> anotherlongfunctionname "bar" 1 |> somelongfunction |> bazasdasdad;
```
jordwalke: Looks like this should be breaking at some point. It's 113 chars long, yet the print width for test is about 50 chars iirc.

PR #1260, opened 2017-05-12
Remove useless files

chenglou:


jordwalke:
Thanks merge as soon as CI passes. How did you test that merlin still works? Did you uninstall reason-cli and then use opam pin?

chenglou:
yeah

PR #1261, opened 2017-05-13
Port to jbuilder

rgrinberg:
Quicky and dirty POC. Gets as far as building refmt, reasonbuild, and the merlin
plugin

TODO:

- [x] rtop
- [ ] Make the ppx rewriters jbuilder friendly
- [x] Write correct metadata in the various opam files
- [ ] install emacs/vim plugins and documentation
- [x] introduce topkg-jbuilder
- [ ] Port menhir related makefile tasks to jbuilder


jordwalke:
I think there's some files you can delete as part of this change? Topkg stuff?
Does this make it easier to release versions of reason, reason-parser (and the new package you are creating for rebuild)? It's kind of a hassle to release versions to opam as it is.

Also, do you have advice about when jBuilder will be stable / tested enough to actually merge a PR like this? What's the ETA on jBuilder coming out of beta?

rgrinberg:
> I think there's some files you can delete as part of this change? Topkg stuff?
Does this make it easier to release versions of reason, reason-parser (and the new package you are creating for rebuild)? It's kind of a hassle to release versions to opam as it is.

Yes. Although, all the necessary pieces for this might have not landed to opam yet. In my opinion manual releases aren't too tough however.

> Also, do you have advice about when jBuilder will be stable / tested enough to actually merge a PR like this? What's the ETA on jBuilder coming out of beta?

Only @diml can answer this. My impression of this is that the core of jbuilder isn't changing very much and all of the work is going into implementing a long tail of pet features that everyone requests.

rgrinberg:
@jordwalke Have a look at the opam files introduced.

diml:
Basically i'm just waiting for the big projects i know of that are switching to jbuilder to make sure there is enough to handle the common needs of packages. There are also a few features that i think should go in v1 amd are not finished yet such as doc generation.

The jbuild spec might be extended a bit but this won't break existing projects. Jane Street packages and a few other already rely on jbuilder

rgrinberg:
I've added topkg-jbuilder integration for making correct OPAM releases.

Try doing `$ topkg lint` for example.

hcarty:
What's the current state of this work?  Reason can't currently be used with several libraries on opam due its  topkg version constraint,

jordwalke:
We are looking for ways to make it easier to contribute to reason. This diff would certainly help since it speeds up build time of Reason, but I believe it now needs to be rebased on top of master since we got rid of the reason-parser package, and we squashed them into one opam package. If that was done, I'd definitely consider merging this now that jbuilder has matured.

Issue #1263, opened 2017-05-16
JSX parsing error:


jordwalke:
This parses correctly:

```
Container.make [| <Button onClick=handleStaleClick />, <Button onClick=handleStaleClick /> |]
```

This doesn't parse correctly (only difference is no spaces)

```
Container.make [|<Button onClick=handleStaleClick />, <Button onClick=handleStaleClick />|]
```

The `[|<` are likely tripping it up, and if that's the case that's not so bad as long as the printer retains the spaces - it doesn't currently.

IwanKaramazow:
```reason
let x = [|<Button onClick=handleStaleClick />, <Button onClick=handleStaleClick />|];

let y = [<Button onClick=handleStaleClick />, <Button onClick=handleStaleClick />];

let z = (<Button onClick=handleStaleClick />, <Button onClick=handleStaleClick />);
```
Forgot to report this one, encountered this one last week. Might have a solution, will send PR.

IwanKaramazow:
https://github.com/facebook/reason/pull/1264

This is really a suboptimal solution, with EasyFormats current api.
Changed the `~wrap:("( ", " )"` to include a space if there's jsx inside the structure..
I really want something like: `~listConfigIfBreaks`.
If we're breaking now, we have an extra redundant space around the wrapping brackets...

jordwalke:
I left a suggestion about how to get better spacing without wrapping in `("[ ", " ]")` but Fred might be able to suggest a lexer hack that achieves what we want.

PR #1264, opened 2017-05-18
Add space between list, array, tuple and opening/closing jsx bracket

IwanKaramazow:
https://github.com/facebook/reason/issues/1263

Fixes printing of jsx inside a list/array/tuple.

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/jsx.re
+++ b/formatTest/typeCheckedTests/expected_output/jsx.re
@@ -246,29 +246,29 @@ let spaceBefore2 = <So> <Much /> </So>;
 let siblingNotSpaced =
   <So> <Much /> <Much /> </So>;
 
-let jsxInList = [<Foo />];
+let jsxInList = [ <Foo /> ];
 
-let jsxInList2 = [<Foo />];
+let jsxInList2 = [ <Foo /> ];
 
-let jsxInListA = [<Foo />];
+let jsxInListA = [ <Foo /> ];
 
-let jsxInListB = [<Foo />];
+let jsxInListB = [ <Foo /> ];
 
-let jsxInListC = [<Foo />];
+let jsxInListC = [ <Foo /> ];
 
-let jsxInListD = [<Foo />];
+let jsxInListD = [ <Foo /> ];
 
-let jsxInList3 = [<Foo />, <Foo />, <Foo />];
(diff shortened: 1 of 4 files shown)


formatTest/typeCheckedTests/expected_output/jsx.re
```diff
@@ -246,29 +246,29 @@ let spaceBefore2 = <So> <Much /> </So>;
 let siblingNotSpaced =
   <So> <Much /> <Much /> </So>;
 
-let jsxInList = [<Foo />];
+let jsxInList = [ <Foo /> ];
```
jordwalke: It's kind of unfortunate to have to add spaces around everything like `[]` and `()` just because it's needed around `[| |]` in the *one* case where you use a self-closing tag.

Even `[|<X> </X>|]` parses correctly. It's just `[|<X />|]` that doesn't.

formatTest/typeCheckedTests/expected_output/jsx.re
```diff
@@ -375,58 +375,62 @@ let thisIsAlsoOkay =
...
   <> <Bar /> </>
-];
+ ];
 
 let listOfListOfJsx = [
   <> <Foo /> </>,
   <> <Bar /> </>,
   ...listOfListOfJsx
-];
+ ];
```
jordwalke: This diff currently uses the `~wrap:("[ ", " ]") config to achieve the spacing, and as a result it causes indentation issues like this.

There's a formatting argument called `~pad:(true, true)` which will ensure there is spaces before the closing and after the opening, which should avoid these issues.

jordwalke:
Let's see what @let-def thinks about this:

There's this issue where *self-closing* JSX inside of an Array fails to parse:

```
[|<X />|]
```

The issue is that `/>|` is in fact a valid infix identifier.

However - `/>|` would never come before a `]`, so perhaps there's a much simpler solution at the lexer level that doesn't require us to change our formatting conventions in ways we wouldn't otherwise.

IwanKaramazow:
Gonna close this; should be fixed in the lexer.

formatTest/typeCheckedTests/expected_output/jsx.re
rickyvetter: What is the parsing ambiguity here?

PR #1271, opened 2017-05-23
Minor example fix

voluntadpear:
Parameter label for radius was missing in the drawCircle example

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -262,7 +262,7 @@ Labeled function arguments can be made optional during declaration. You can then
 
 ```reason
 /* radius can be omitted */
-let drawCircle ::color ::radius=? () => {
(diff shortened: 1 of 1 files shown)


docs/index.html
```diff
@@ -262,7 +262,7 @@
 
 ```reason
 /* radius can be omitted */
-let drawCircle ::color ::radius=? () => {
+let drawCircle ::color radius::r=? () => {
```
SanderSpies: I think that in this case it's nicer to leave the `radius` argument as-is and change the `r` below into `radius` instead.

voluntadpear: Sure, I'll change that

SanderSpies:
LGTM

SanderSpies:
#sanderMerge

PR #1272, opened 2017-05-23
Add more information about the file => module transformation.

bsansouci:


chenglou:
not sure this should be in the modules section. Should this be in FAQ instead? There's a mention of modules there

chenglou:
This is way stale ¯\_(ツ)_/¯

We'll have proper namespacing soon too

PR #1275, opened 2017-05-26
Spelling correction in No Arguments block in /docs

cwhitten:
argumetn -> argument


chenglou:
Oops sorry, I just saw #1280 and merged that one instead... Thanks for spotting it though!

PR #1279, opened 2017-05-28
Add a dependency example without `include` keyword

ajhyndman:
As a JavaScript developer, this section was initially confusing for me.  I assumed that `include` was a replacement for `import`, and got confused.

I think adding another example that doesn't have any such import keyword might be sufficient to demonstrate that `include` is performing an unrelated function.


ajhyndman:
Looks like something died in the ocamlbuild step?

```
ocamlfind: Error from package `threads': Missing -thread or -vmthread switch
```

Is there anything I can do to resolve this?

chenglou:
The CI error is unrelated. Thanks for the PR!

PR #1280, opened 2017-05-30
Correct typo on no argument.

LongYC:
Correct a simple typo in doc where "no argument" was spelled "no argumetn".


chenglou:
Thanks!

PR #1281, opened 2017-05-30
Use ocamlrun when calling utop

jordwalke:
This will help with the issue of relocatability.


jordwalke:
Doesn't really seem to be needed anymore with the relocatable releases we have now.

Issue #1278, opened 2017-05-27
Error running 'opam install reason'


FrontMage:
```
▶ opam install reason                                          

=-=- Synchronising pinned packages =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
[reason] /home/xinbg/Downloads/reason/ already up-to-date
The following actions will be performed:
  - install reason 1.13.5*

=-=- Gathering sources =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
[reason.1.13.5] /home/xinbg/Downloads/reason/ already up-to-date

=-=- Processing actions -=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
[ERROR] The compilation of reason failed at "./build.native build --native true --native-dynlink true --utop true".

#=== ERROR while installing reason.1.13.5 =====================================#
# opam-version 1.2.2
# os           linux
# command      ./build.native build --native true --native-dynlink true --utop true
# path         /home/xinbg/.opam/4.02.3/build/reason.1.13.5
# compiler     4.02.3
# exit-code    1
# env-file     /home/xinbg/.opam/4.02.3/build/reason.1.13.5/reason-9240-05296d.env
# stdout-file  /home/xinbg/.opam/4.02.3/build/reason.1.13.5/reason-9240-05296d.out
# stderr-file  /home/xinbg/.opam/4.02.3/build/reason.1.13.5/reason-9240-05296d.err
### stdout ###
# [...]
# ocamlfind ocamldep -package 'menhirLib unix compiler-libs.common ocamlbuild findlib merlin_extend result dynlink findlib.dynload ocaml-migrate-parsetree reason-parser' -modules src/reason_util.ml > src/reason_util.ml.depends
# ocamlfind ocamlc -c -I +ocamldoc -g -bin-annot -safe-string -package 'menhirLib unix compiler-libs.common ocamlbuild findlib merlin_extend result dynlink findlib.dynload ocaml-migrate-parsetree reason-parser' -w @5@8@10@11@12@14@23-24@26@29@40 -I src -I vendor/cmdliner -o src/reason_util.cmo src/reason_util.ml
[...]

PR #1282, opened 2017-05-30
Fix utop compilation

cristianoc:
https://github.com/facebook/reason/issues/1278

### DIFF
--- a/_tags
+++ b/_tags
@@ -1,5 +1,6 @@
 true: warn(@5@8@10@11@12@14@23-24@26@29@40), bin_annot, safe_string, debug
 
+<src/reason_utop.ml>: thread
 <editorSupport/**>: -traverse
 <node_modules/**>: -traverse
 <bspacks/**>: -traverse
--- a/src/rtop_init.ml
+++ b/src/rtop_init.ml
@@ -1,5 +1,6 @@
 #!/usr/bin/env ocaml
 #require "ocaml-migrate-parsetree";;
+#require "ocaml-migrate-parsetree";;
 #require "menhirLib";;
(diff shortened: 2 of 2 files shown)


chenglou:
Feel free to self-merge

cristianoc:
Actually tests are failing. Specifically the utop test fails when doing:
```
utop -init src/rtop_init.ml
...
Error: Reference to undefined global `Misc'
```

This happens when either of the following are run in utop:
```
#require "ocaml-migrate-parsetree";;
or
#require "reason";;
```

Somehow this happens on utop versions greater than 1.19.3.

cristianoc:
@let-def any thoughts?

let-def:
This unit is defined in `compiler-libs/ocamlcommon.cma`, maybe try with:
`#require "compiler-libs.common";;`.

cristianoc:
@letdef that did not work, but this seems to wort for me:
```
#load "compiler-libs/ocamlcommon.cma"
```
wondering how reliable that is.

let-def:
compiler-libs.common is a direct dependency of ocaml-migrate-parsetree, it would be weird if it had worked by requiring it manually...
But this dep should trigger the loading of "ocamlcommon.cma":
```
$ ocamlfind ocamlc -only-show -linkpkg -package compiler-libs.common
findlib: [WARNING] Interface topdirs.cmi occurs ...
ocamlc.opt -I /.../lib/ocaml/compiler-libs /.../lib/ocaml/compiler-libs/ocamlcommon.cma
```
That's the only purpose of compiler-libs.common package.
There is something fishy :/.

PR #1288, opened 2017-06-04
New jsx ppx tweaks

chenglou:
Accepts a command line flag. If it's `-version 2`, then use the new JSX
transform. Otherwise use the old transforms.

At the file-level, user can do `[@@@bs.config {jsx: 2}]` and it'll use the new version, regardless of the command line flag passed (aka what's configured in `bsconfig.json`). Likewise for `{jsx: 1}`.

It'll remove `[@@@bs.config {...}]` if `jsx` is the only keep in the record.

This makes sure nothing break when we upgrade everyone to the new
transform (used by rr2), *and* that when they start a new project, they
can put a `{"reason": {"react-jsx": 2}}` in `bsconfig.json` and have the
new transform without needing to put any config at the top of the files
(bsb will take care of invoking the ppx with the right flags depending
on bsconfig here).


chenglou:
Don't merge yet. Waiting for @bobzhang

chenglou:
Done. Updated comment

chenglou:
@jordwalke this has been merged into BS

PR #1291, opened 2017-06-05
Use 2 ppxes for jsx again

chenglou:
Turned out using a single ppx + using a CLI flag isn't ideal because of the way flags are handled + paths that contain spaces.


chenglou:
cc @bobzhang

src/reactjs_jsx_ppx_2.ml
```diff
@@ -0,0 +1,223 @@
...
+           pexp_attributes
+         } when Syntax_util.attribute_exists "JSX" pexp_attributes ->
+         let attributesNoJsx =
+           List.filter (fun (attribute, _) -> attribute.txt <> "JSX") pexp_attributes in
+         handleJsxCall mapper callExpression callArguments attributesNoJsx
+       (* Delegate to the default mapper, a deep identity traversal *)
+       | e ->
+         default_mapper.expr mapper e) in
+
+  Reason_toolchain.To_current.copy_mapper { default_mapper with structure; expr }
```
bobzhang: seems you can create a dedicated module for `To_current` to avoid pulling in such dependencies

PR #1299, opened 2017-06-14
Uniform parenthesized syntax for abstraction and application

let-def:
This PR pushes https://github.com/facebook/reason/pull/1236 a deeper.
It is not yet ready for merging but some feedback could help further development

### Expressions

```reason
/* Abstraction */
() => ...                      /* function taking unit */
(a) => ...                     /* function of one argument */
(a : int, b) => ...            /* function of two arguments, first being of type `int` */
((a,b)) => ...                 /* function of one argument that is a tuple */
fun (a,b) : t => ...           /* function of two arguments returning a `t` */
(:x, :y) => ...                /* function with two labelled arguments `x` and `y` */
(:x : int, :y : string) => ... /* function with two labelled arguments `x` and `y` of types `int` and `string` */
(:x = 5, :y = ?, :z = ? : string, ()) => ... /* function with three optional arguments and a unit, first defaulting to the integer `5`, second has no default, third has no default but should be a string */

/* Application */
f()                    /* applying f to `()` */
f(x)                   /* applying f to `x` */
f(x,y)                 /* applying f to `x` and `y` */
f((x,y))               /* applying f to the tuple made of `x` and `y` */
f(:x)                  /* applying labelled argument `x` */
f(:x?, ())             /* applying optional argument `x` and `()` */
f(:x 5, :z? f(x), ())  /* applying with explicit values */
```

### Arrow type

```reason
(a) => b      /* functions from a to b */
(a, b) => c   /* functions from a to (b to c), equivalent to (a) => (b) => c */
((a, b)) => c /* functions from a tuple (a, b) to c */
(:x int, :y string) => c /* functions with two labelled arguments to c */
(:x a?) => b  /* functions with an optional argument x of type a to b */
```

### Type constructors

```reason
/* Abstraction */
type t = ...           /* Ground type */
type t('a) = ...       /* One parameter, a */
type t('a, 'b) = ...   /* Two parameters */
class type t = ...     /* Same with class types */
class type t('a) = ... 
class type t('a, 'b) = ... 

/* Application */
t        /* constant type */
t(a)     /* t with actual parameter a */
t(a,b)   /* ... */
```

### Value constructors

```reason
/* Definition */
type t =
  | A                 /* no arguments */
  | B(int)            /* one argument of type int */
  | C(int, int)       /* two arguments that are of type int */
  | D((int, int))     /* one argument that is a tuple of int */
  | E{x: int, y: int} /* two named arguments (inline record) */

/* Construction */
A
B(1)
C(1, 2)
D((1, 2))
E{x: 1, y: 2}

/* Elimination */
let f = fun
  | A => ...
  | B(x) => ...
  | C(x, y) => ...
  | D((x, y)) => ...
  | D(tuple) => ...
  | E {x, y} => ...

/* Wildcard */
let f = fun
  | A(_) | B(_) | C(_) => ...
  | E(_) => ...
  | E {x, _} => ... 
```

### Modules

```reason
/* Application */
F1(Int)
F2(Int,String)
Effectful()
Effectful2(Int,())
Effectful2(Int)()

/* Abstraction */
module F1 (A : ASig) = { ... };
module F2 (A : ASig, B : BSig) () = { ... };
module Effectful () = { ... }
module Effectful2 (A : ASig, ()) = { ... }

/* Module types */
() => ...          /* One "unit" argument */
(A : ASig) => ...  /* One dependent argument */
(ASig) => ...      /* One non dependent argument */
(A : ASig, B : BSig) => ... /* Two arguments */
```

### Bindings

Previously, `=` was used for binding most values but the sugar for binding functions used `=>`.
Now `=` is always allowed and `{ ... }` can be used for function bodies.

```reason
let add(a,b) = a + b;

let add(a,b) {
  a + b
};

let immediate_object(a,b) = {
  pub a = a;
  pub b { b };
};
```

### Classes

```reason
class a(x) {
  /* one value parameter x */
  pub a = 0;
  pub a() = 0;
  pub a(x,y) = x + y;
  pub a(x,y) {
    let result = x + y;
    print_endline(" x + y = " ++ string_of_int(x) ++ " + " ++ string_of_int(y) ++ " = " ++ string_of_int(result));
    result
  };
} 

class container('a)(x) {
  /* one type parameter 'a, one value parameter x */
} 
```


IwanKaramazow:
@let-def How do you feel about value/type constructors with those extra parens?

yawaramin:
Can you show an example of how to write a GADT? E.g. (OCaml):

    type 'a t = A : int -> int t | B : bool -> unit t

TheSpyder:
Is the plan to keep both old and new styles moving forward, or at some point cut over and break source compatibility requiring the new style?

The overlap between parenthesised style multi-argument and a non-parenthesised style tuple argument means I'll probably avoid the new style for multi-arg functions when there's a chance of mixed styles in a codebase. I generally prefer spaces for argument separators in general though; it serves as a reminder that the arguments are curried.

I really like the removal of mandatory arrows for let function definitions, that always strikes me as weird in reason having done a fair bit of ocaml code.

yawaramin:
@TheSpyder from what I've read in the past (can't find a reference right now), the standard way to handle syntax changes is for updated versions of `refmt` to be able to migrate your codebase over from the old syntax to the new. `Refmt` understands multiple syntaxes, so you update the Reason toolchain, run `refmt` with the appropriate options, and you automatically get the new syntax.

TheSpyder:
Oh, have there been syntax migrations in the past? I've been around the community for a while but only actually using reason for a short time.

jordwalke:
@yawaramin @TheSpyder the way we can upgrade syntaxes, even between non-forwards-compatible changes, is that we simply parse with one version of `refmt`, then print with another!

With fast-to-install local sandboxed binary releases, it should be pretty easy:
```
mkdir old && pushd old && yes | npm init && npm install https://github.com/reasonml/reason-cli/archive/beta-v-OLDVERSION-bin-darwin.tar.gz && popd
mkdir new && pushd new && yes | npm init && npm install https://github.com/reasonml/reason-cli/archive/beta-v-NEWVERSION-bin-darwin.tar.gz && popd

./old/node_modules/.bin/refmt --print binary_reason file.re > file.intermediate.re
./new/node_modules/.bin/refmt --parse binary_reason --interface=false --print re file.intermediate.re > file.re 
```

let-def:
@IwanKaramazow `type t('a) = A(int) : t(int) | B(bool) : t(unit);`
@TheSpyder This is an experiment. The intention is clearly to move away from curried application, which leads to implicit partial applications. These doesn't match how hardware work and forces the compiler to do compile time analysis to detect under/over-applications, which departs from the overall predictable compilation model that reason/ocaml native backend offer.
Of course this work is only at the syntax level, so we cannot remove nor deprecate partial applications in the semantics, but this is the first step to appreciate the impact that would have on the language.

What about extra parentheses? I don't have much hind
In expressions, I feel it is definitely a win. For value constructors, it matches function application so it feels regular.
The two places where it feels a bit annoying are:
- arrow types, especially with unary functions, having to write `(unit) => unit` instead of `unit => unit` is confusing.
- unary type constructors, e.g `list(int)` or `option(bool)`.

But I have years of practice with a different syntax, with OCaml before Reason existed, I am biased. I prefer to have a bit more practice before drawing conclusions. On the other side, there is a huge win in uniformity which makes the code more readable.

Note that the current printer is not minimal in the number of parentheses.

let-def:
Things to do:

- [ ] optimize printer: less parenthesis (preserving correctness), adjust layout (aesthetic)

- [ ] tests for distinguishing
```reason
module type T = T with type a = a => a
module type T = (T with type a = a) => a
module type T = T with type a = (a => a)
```

- [ ] update outcometree printer

- [ ] fix parsing of inline record value construction, `A{x=5}`

- [ ] make some semicolons optional when non ambiguous?

TheSpyder:
Bringing in my comments from the discord chat about this, I don't mind parenthesis for definition _too_ much but for application it rubs me completely the wrong way. I can see why you want this syntax if you're avoiding partial application, but imo one of the big benefits of the FP style is partial application. I use curried functions in JS all the time despite the unweildy syntax.

With parenthesis making use of partial application will be even more confusing than with spaces, and arguably outright misleading.

neonsquare:
Hm... I have to agree to TheSpyder here (at least in some regards) - this parens and commas syntax thing started as something to make it look more familiar to JS people. I can follow that reasoning even if I think this isn't that much of a showstopper to newcomers. Now I see that the plan here seems to get extended not only to definition and call syntax, but there is also the plan to eliminate partial application as a language feature completely. This deems VERY wrong to me. When I last raised objections about the parenthesized syntax it was said that partial application would just work as before. While I prefer the non-paren syntax I at least know and sometimes use partial application in JS through libs like ramda or lodash. Reason with Bucklescript would make such use compile more efficient while looking similar to what JS people are used.

Please don't throw the baby out with the water!

let-def:
I will try to clear some misconceptions. First, this PR doesn't change anything fundamental: only the syntax of application, abstraction and arrow spaces. Partial application doesn't disappear.
"At worst", in the future there might be a warning for an implicit partial application.

Instead, the intention is to move toward an explicit syntax for partial application. The reason is that currying might not be a wise choice in a non-pure language:

1) There is no isomorphism between `(A×B)⇒C` and  `A⇒B⇒C`. Functions can have side effects at arbirary points.
A `gensym : unit => unit => int` function looks cool at first when you understand how it works, but such a behavior would benefit from a more explicit syntax. It has other cool applications, for instance in http://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.216.864 partial applications are used to share expensive computations.
But in 99% of cases, you don't need (or shouldn't rely) on these tricks. It leads to defensive programming techniques, see for instance https://blogs.janestreet.com/making-staging-explicit/. 

So a specific syntax for when you rely need to "under-apply" brings more benefits than a single syntax for partial application and "normal" application.

2) It is difficult to compile, and this is especially relevant in OCaml language(s): OCaml is designed with predictable performance in mind. But function calls don't fully enjoy this property.

The compiler must analyse direct and indirect calls. As a result abstract function calls are a bit more expensive than necessary. This is even worse in presence of labelled arguments, and the native compiler chose to approximate the semantics (but don't worry, this shouldn't be observable by any "sane" code). Of course, we cannot change that behavior at Reason level.

3) In the next years, algebraic effects will come to the language, and with them different flavor of arrows. The parenthesized syntax will be a good fit for pure arrows, while the "curried" syntax will be natural with effectful arrows.

let-def:
An example to illustrate these different uses of partial applications:

```reason
let add x y => x + y;
let sum x => {
  let k = ref x;
  fun y => k := !k + y; !k 
};
```

The two functions have the same type:
```reason
add : int => int => int;
sum : int => int => int; 
```

If restricted to "full application", these two functions behave the same:
```reason
let test1 x y = assert (add x y == sum x y);
```

However partial applications are easy to distinguish. Consider for instance:
```reason
let add_3 = add 3;
let sum_3 = sum 3;
```

`assert (add_3 x == sum_3 x)` will not hold long.

The PR doesn't have a solution to offer (once again, it is only syntactic), but we could imagine that in the future `add(3,_)`, `add(_,3)` will be allowed (explicit partial applications at arbitrary point) while a warning could be displayed suggesting `sum(3,4)` (suggesting to use `sum(3)(4)` to delimit effects).

jordwalke:
One thing that might be distracting, is that this diff as-is is still useful/helpful to newcomers even *if* nothing ever happens with Effects, and even if no compiler warnings are added for implicit/accidental partial application. I almost think those issues (which are a little more speculative) would be best to discuss in a completely separate issue. This PR can be discussed on its own, as it applies today. If that is the scope of discussion: this diff removes/adds no languages features - it is purely an alternative syntactic rendering for what we already do - and one that an *overwhelming* percentage of JavaScript developers have requested after initially trying Reason.

TheSpyder:
I don't mind if it's an alternative syntax that I can ignore and both are maintained. I may even use it occasionally for single-argument inline function definition. But I was getting the impression that it would be a forced change and require source migration!

jordwalke:
@TheSpyder This is the kind of thing that we would want to have uniformity around in all the docs, and flagship examples, but at a technical level we can make it so that you are not forced to upgrade - you would upgrade at your convenience, but still be able to interop with other libraries that use ocaml or any other future breaking version of Reason syntax.

formatTest/typeCheckedTests/expected_output/attributes.re
```diff
@@ -200,10 +200,9 @@ type variantType =
 
 [@@onVariantType]
 type gadtType('x) =
-  | Foo(int) : [@onFirstRow] gadtType(int)
-  | Bar
-      ([@onInt] int)
-      : [@onSecondRow] gadtType(unit)
+  | Foo(int): [@onFirstRow] gadtType(int)
```
jordwalke: I actually could have gone either way on this one. It did sometimes bother me how before the GADT colon was stuck to the constructor.

formatTest/typeCheckedTests/expected_output/attributes.re
```diff
@@ -393,7 +392,7 @@ external render : (reactElement, element) => unit =
 external createCompositeElementInternalHack :
   (
     reactClass,
-    Js.t({.. reasonProps : 'props}),
+    Js.t({.. reasonProps: 'props}),
```
jordwalke: But for record/object keys, it does make sense to have the colon be tightly bound. I'm leaning towards always printing the colon with spaces around it when it is a type annotation, and like `key: type` when it is a record/object label/type.

formatTest/typeCheckedTests/expected_output/oo.re
```diff
@@ -114,55 +114,55 @@ let (<..>) (a, b) = a + b;
 
 let five = 2 <..> 3;
 
-type nestedObj = {. bar : {. a : int}};
+type nestedObj = {. bar: {. a: int}};
```
jordwalke: 👍

formatTest/unit_tests/expected_output/features403.re
```diff
@@ -11,18 +11,16 @@ type nonrec u('a) =
...
 type expr('a) =
-  | Val{value: 'a} : expr('a)
-  | Add
-      {left: expr(int), right: expr(int)}
-      : expr(int)
-  | Is0{test: expr(int)} : expr(bool)
-  | If
-      {
+  | Val{value: 'a}: expr('a)
+  | Add{left: expr(int), right: expr(int)}: expr(
```
jordwalke: Is it difficult to clean up how this breaks?

formatTest/unit_tests/expected_output/variants.re
```diff
@@ -269,9 +269,9 @@ let thisWontCompileButLetsSeeHowItFormats =
  * GADTs.
  */
 type term(_) =
-  | Int(int) : term(int)
+  | Int(int): term(int)
   | Add: term(((int, int) => int))
-  | App(term((('b) => 'a)), term('b)) : term('a);
+  | App(term((('b) => 'a)), term('b)): term('a);
```
jordwalke: I notice that forcing the parenthesis around simple identifiers in lambdas (and their types) ends up making these nestings pretty paren-heavy. Compare:
```
 | App(term((('b) => 'a)), term('b)) : term('a);

 | App(term(('b => 'a)), term('b)) : term('a);
```

But this is probably partly because it looks like this has an additional parenthesis set in here. It could perhaps be as clean as:

```
 | App(term('b => 'a), term('b)) : term('a);
```

formatTest/unit_tests/expected_output/wrappingTest.re
```diff
@@ -2527,15 +2530,14 @@ let letsPutAWhereClauseOnTheLast (x) =
...
+                                   )
   | Add: wrappingGadt(((int, int) => int))
-  | App
-      (
-        wrappingGadt((('b) => 'a)),
-        wrappingGadt('b)
-      )
-      : wrappingGadt('a);
+  | App(
+         wrappingGadt((('b) => 'a)),
```
jordwalke: Same comment for additional parenthesis here. It could be one of the following:
```
       wrappingGadt('b => 'a)
```
or
```
       wrappingGadt(('b) => 'a))
```

jordwalke:
Another thought about what this syntax can solve. One common complaint for people using the Reason syntax, is that these kind of functions become unwieldy:

```
let myFunc a b =>
  switch () {
  | Really =>
  | Long =>
  | Body =>
  };
```

it happens with `if` expressions as well as `switch`.

If the function body had a `{ }` around it, it would be much easier to add a temporary "variable" before the switch/if etc. But we don't want functions to *always* have `{ }` because then they become less "lightweight".

This syntax offers a solution: We can parse
```
let add(a, b) = a + b;
```
But then if ever the expression on the right hand side of the `=` breaks, you could print it as:
```
let add(a, b) {
   a + b;
};
```

Semi-related: another request people have asked repeatedly for is to force the final semicolon in the sequence blocks. The reason is twofold - it helps copying and pasting lines around, and it also distinguishes the single field record visually. I think people would be really happy about that last one.

formatTest/unit_tests/expected_output/variants.re
let-def: Commit https://github.com/facebook/reason/pull/1299/commits/0640e0dc1cffbf803c3c87c077cfef787bb36c0d prepare the ground for that. Your proposal can be parsed now, but is not printed yet.

let-def:
Still heavy work in progress.

What's new:

1) Unary arrow types can now be written without parenthesis, `int => int` instead of `(int) => int`.
The arrow must have no label, and the type at the left of the arrow must be a type constructor (named type).

2) The conditional/scrutinee in `if`, `while`, `switch` and the body in a `try` statement must now be wrapped in any kind of parentheses (`(`, `{`, `[`, ...).
This makes the grammar much less ambiguous, while not requiring parentheses didn't made code much more lightweight in a statement that's already quite verbose.

This allows a lightweight call syntax for record/object arguments: `f {x: 5}`, `Some {contents: "hello"}` which would otherwise be ambiguous.

```reason
let not(b) = if (b) false else true;
try (f(a,b)) { | Not_found => (); };
try { 
  complex body
} { 
  | Not_found => ();
};
switch (a, f(b)) {
 | (None, Some(n)) => ...;
};
```
For uniformity, `for` loop also takes parentheses: `for (i in 0 to 10) print_int(i);`

3) Minimal escaping for operators:
- parsing accepts `\` at arbitrary position for operators,
- printing put the minimum number of `\` to avoid ambiguities,
- an identifier beginning with `\` is handled differently (used to pass names between OCaml and Reason that would conflict with a keyword otherwise)

4) `^` is now `++`, `!` is now `not`, and dereferencing is done with postfix `^`.

- string concatenation: `"hello " ++ "world"`
- boolean negation: `if (!validate(a,b)) prerr_endline("Wrong!");`
- dereferencing:
```reason
let counter() {
  let x = ref(0);
  () => { incr(x); x^ };
};
```
Postfix composes well with dot notation: `root^.child^.text`.

let-def:
Next work will focus on printer, which is becoming quite hard to work with.

jordwalke:
You should probably add a section to the main examples in this PR showing @SanderSpies's decorators.

jordwalke:
What's this docstring feature you added? :D

bobzhang:
I like the idea in general, you mentioned you want to give a warning on partial application, how do you achieve that?
@let-def are you aware of how bucklscript add uncurried calling convention support? the downside is that it does not allow polymorphic toplevel uncurried function support

jaredly:
(I think the warning on partial application stuff is a potential thing that would come later)

let-def:
@jordwalke I removed the docstring preview, I will reimplement later on top of the new attributes

@bobzhang No I don't know about that feature. Do you have a link that explains it?

let-def:
In https://github.com/facebook/reason/pull/1112, @SanderSpies moved the attribute at the beginning of the decorated items, bringing them closer to python style decorators.

```reason
[@deriving compare]
type t = int;
```

This work has been rebased on this PR. On top of that, I removed the distinction between `[@`, `[@@` and `[@@@`: now all attributes are introduced with `[@`.

Floating attributes ends with a semicolon in a structure, item attributes and normal attributes are disambiguised by syntax:

```
[@floating];

[@on type]
type t;

[@on value x]
let x = [@on expression foo] foo;
```

The only problematic case is with top-level expressions (`Pstr_eval`).
```
[@attribute] 42;
```
It is unclear whether `[@attribute]` applies at the expression level (`([@attribute] 42)`) or the structure level (`[@attribute] (42)`).

The current PR choice is that attribute applies to the structure and that one should put parentheses to decorate the expression. This forces to put parentheses to decorate top-level expressions, which is not that pretty yet it is a quite a fringe feature (except maybe in the toplevel).

The other way is possible: without parentheses, attributes attach to the expression, with parentheses to the structure.

By the way, structure level attributes are currently dropped by reason printer, see https://github.com/facebook/reason/pull/1329 .

jaredly:
I'd guess that it'll be more common to want an expression-level decorator on a top-level expression than to want a structure-level one. although probably neither will be that common

jordwalke:
Whoa, that's going to be way easier to teach people. I can hardly keep try of all the various `@/@@/@@@` forms myself.


@let-def I think we'll run into a similar situation with extension point "expressions" vs. module structures. We can apply the same convention that we decide on (what do parens imply?) to both attributes and extension points.

```
/* Is it an expression/eval extension point or a module structure item? */
[%this];
```

jaredly:
For extension points, I would vote that `[%thing]` at the top level is a structure item, and you have to wrap in parens `([%thing])` for it to be an expression.

cullophid:
Im not sure I think this is a step in the right direction.
As far as iI can tell from your example `f(x,y)` would mean applying `x` and `y` to `f` while `f (x, y)` would mean applying the Touple `(x,y)` to the function `f`.
That seems very confusing... 
Reason has a very nice and minimal syntax right now and I would be sad to see that change :(
I can understand removing unpopular languages traits for performance benefits, but currying is an essential part of reason and all ml languages.
I am also not quite sure what you mean by implicit partial application ?

cullophid:
I would also argue that in the example by @let-def with `add` and `sum` the problem isn't actually about partial application but about side effects. this bit of code has the exact same problem;
```
let k = ref 0;
let add a b => a + b;
let sum a b => {
  k := !k + 1;
  a + b + !k
};
```
The real issue is that side effects does not show up in the signature. Im not sure how to fix that (or if it should be fixed) but a keyword like `rec` but for side effects could be an option, as well as a signature that indicates an effect.

jordwalke:
@cullophid :
> Im not sure I think this is a step in the right direction.
As far as iI can tell from your example f(x,y) would mean applying x and y to f while f (x, y) would mean applying the Touple (x,y) to the function f.
That seems very confusing...

That is not the case. There is no sensitivity to spaces between the function name and argument list.

hcarty:
@cullophid There is work underway to add algebraic effects to OCaml - see https://github.com/kayceesrk/effects-examples for some examples of what may come with that.

cullophid:
@jordwalke ah so this new syntax would replace the old function invocation syntax? I like that even less :(
@let-def raised a concern earlier about curried functions with internal state (the add/sum) example.
I was just pointing out that the issue was actually about internal state, and not currying.
An issue could be that there is no way to tell from a type signature if a function has a side effect.
I could imagine that this could cause some errors if you think you are using a pure function.
I haven't used the language enough to know if this is actually an issue.

jordwalke:
@ullophid do you believe you are the target audience for this syntax change? If not, do you work with many people who would be the target audience? Also, do you like OCaml's syntax?

TheSpyder:
Looks like @cullophid has the same reaction I do to this change ;)

For the record, I'm in an environment where most of the code produced is JS but ML style syntax is not a problem (quite a few of my coworkers are haskell fans). I do like ML style syntax for a lot of things.

However I can understand that Reason is trying to appeal to JS devs - one of my coworkers did claim that for JS devs space separated function arguments would be far too weird. That, and some of the other improvements in this PR which I like the look of, mean I'll definitely be migrating to this updated syntax when it lands.

cullophid:
@jordwalke I don't know who the target audience for this syntax change is?
if its Javascript developers then I am absolutely the target audience (though I don't speak for all JS devs :))

If the point is to appeal to JS devs, I don't think this is the right move. Yes the new syntax would be more familiar, but the current is more elegant and simple and I found it very easy to pick up.

Partial application is a different story. Partial application definitely took some time. I learned about it in javascript about 3 years ago, and I didn't find it easy to learn. These days I don't think I could live without it (He said over dramatically).

I don't agree with the premise that making reason look more like JS will make it more appealing to JS devs. It will definitely make it feel familiar at first look, the syntax is never the hard bit about learning a new language. 

If you want people to switch language you have to give them a reason(PUN!). "Its just like javascript" is not a good reason to leave javascript. 

You have to people something they don't have. Take a lesson from Elm and make the compiler be a companion that guides you as a programmer. Make sure the tooling is best in class. Lets build an ecosystem that feels familiar.

I apologise for the long rant, but I think this a very important decision and I really think that this change could damage the language :(.

At least I think this should be paused while more important issues are addressed, like async, and friendlier compiler messages. Both of those will do more to help new devs pick up reason.
One thing I learned from npm is that if you want people to use you lib, documentation is way more important than the actual code.

Finally I want to add that I have the utmost admiration for the work you have done... I have only used reason for 2 weeks, but I absolutely love it. Great job!

jaredly:
@cullophid thanks for throwing your hat in! And welcome.

You spend some time talking about things that we should do *other* than change the syntax, like having good error messages, good tooling, etc. I would argue that we're also working on these things.

You also say a couple of times that "syntax isn't the hard part of learning a new language". With that in mind, why would changing the syntax to e.g. align more with c-like languages "damage the language"?
Put another way, it sounds like you both a) don't think syntax differences are important, and b) are very passionate about not moving to c-style function application.

I, for one, think that a) syntax differences can make a difference, and b) c-style is better.
In support of A, we've had a large number of people (primarily coming from javascript) advocating for moving to c-style function application. With the current syntax, there's a barrier to even reading the code (which doesn't take months to get over, but it exists). If we align with javascript, python, java, c, swift, go... all the most popular languages, then it's that much easier for people who are interested to get started. Why add an extra hurdle?

As for why I think it's better, I think that c-style function application (where the arguments are surrounded by parens & separated by commas) results in less cognitive load while reading, and less hassle while writing, than the ml/haskell method. My reason is *consistency*.

With c-style, it's *always* going to be `fname(arg1, arg2)`, regardless of what arg1 and arg2 are.
With ml-style, it might be any one of
- `fname arg1 arg2`
- orr if one of them is complex, you'll probably need to pop in some parentheses `fname arg1 (arg2)`
- and because adding parentheses after the fact is a pain, you can also do `fname arg1 @@ arg2`
- and of course if the result of this fn application is being used anywhere complex, you'll want to wrap the whole thing in parentheses `(fname arg1 arg2)`

`fname arg1 arg2` might feel more elegant and simple in the most basic of cases, but I think it's a false promise -- you end up having to do *more* work to both write & read the code in the general case.

When writing code, I would **so** much rather have One One Right Way to do it, instead of half a dozen rules to remember.

bsansouci:
@cullophid Thanks for describing your concern. We've had people express in detail how they felt about both sides, and after very careful consideration we decided to go for it. If I had to choose two reasons for it I would say:
- this isn't "js syntax" but basically all-other-languages syntax. It will be more familiar to everyone. 
- this isn't a final change! The way Reason is setup is to not be stuck with old syntax, allowing us to iterate on things. 

You seem super motivated to make Reason better, make sure to come on Discord. You'll find tons of things that need work that you can help with.

cullophid:
@jaredly My point wasn't that syntax isn't important. It is!, my point wast that it was not that important when learning a language. That being said I understand your arguments, and I agree with most if not all of them. 

I didn't mean to insinuate that the documentation and tooling around reason was not great. I think you guys have done amazing work. 

@bsansouci Im looking forward to trying out the new syntax. Im very happy that you guys are open and engaging in theses discussions, even with some guy who think he knows best after a having picked up the language a week and a half ago :)

wincent:
You can tell a lot about what a language values by looking at the relative importance that it dedicates to different syntactic constructs. Think of if kind of like Huffman coding for syntax, where the most valuable (most frequent) constructs are encoded in the most efficient way possible. Conversely, constructs that are allowed but not necessarily encouraged can be made somewhat distasteful with a dose of "syntactic vinegar" that makes their application a little less appealing.

When you look at largely functional languages like OCaml and Haskell, you can see just how important function application is to them, because they have dedicated the most valuable syntax of all (whitespace! which is pervasive, like the air we breathe) to function application.

One possibly subtle implication of this change is that function application will be not quite so effortless and uncluttered as before. This is not a negative value judgment; it's just different. Instead, this change makes a statement that consistency, and accessibility — especially for people coming from other languages — is to be valued over minimalism.

I think this is ok. I like the paren-less syntax and feel comfortable with it. But I'll feel comfortable with the new syntax too, and I think the teachability of it is going to be a useful advantage even if it comes at a small cost of punctuation "noise". This cosmetic change won't damage the bit I really treasure: the practical static type checking with great inference and (already) pretty decent error messages. Long live Hindley–Milner!

jordwalke:
Well said @wincent.

One thing to consider about the following statement:

> One possibly subtle implication of this change is that function application will be not quite so effortless and uncluttered as before. This is not a negative value judgment; it's just different. Instead, this change makes a statement that consistency, and accessibility — especially for people coming from other languages — is to be valued over minimalism.

One thing that other functional languages don't have is named arguments with all of the features that OCaml's underlying type system allows (defaults, optional-ness, type annotation inline). Add to that the ability (and desire) to pun named arguments, and it ends up consuming a ton of syntactic real estate in order to make function application unambiguous. I think the new syntax manages to cram all of those features in there more elegantly so that named args is more effortless. I hope this enables labeled arguments to be used in more cases where they would be appropriate.

neonsquare:
I'm personally (!) still inclined to the "this will make it much more cluttered and clumsy"-camp. The so callled "non-paren-Syntax" is actually a syntax much closer to Lisp syntax than the proposed Algol-Style. I prefer that style and I wholeheartedly disagree with anyone who argues that Algol-Style syntax is fundamentally "more readable", "more consistent", "easier to learn/read/use", "more familiar". If you think this way and you think this is something that has to get "fixed" you definitely should think about this being more about you than anyone else. Its subjective and being subjective doesn't make it a "doesn't matter" as some arguing suggests. Quite different: any arguing on this things being objective is definitely the wrong direction.

Been there done that (really): This kind of... erm... "fixing" is something that repeats in history. There are dozens if not even hundreds of efforts to "fix" Lisp syntax. Did anything succeed? Nope. Lisp syntax is still "unfixed". Mostly this projects are driven by newcomers or by catering for newcomers; which is just stating facts and not making such efforts bad as themself. Such projects/libs up to now didnt really get used that much because the natural thing often was that people learned to value the merits of Lisp syntax quicker than common sense seems to imply.

One prominent case of "fixing" the syntax for some people was Dylan. Which first was a language with a Lisp-style syntax that got "fixed" in an effort to win over developers from the Java camp. It didn't really work out though.

Sometimes there is arguing that "syntax doesn't matter" or the opposite "syntax does matter". This is conflated with arguing "which syntax you choose - doesn't really matter" which is a different thing. I think it really would matter if the new default syntax would closely resemble "Brainfuck". It might (nobody can know that) matter to choose either non-paren or paren syntax for Reason - or not. 

It might not matter because the project could fail anyways. It might not matter because the project would succeed anyways. I still have doubts that anyone can objectively argue that a change to Algol-Syntax will be necessary for the project to succeed or that it may lead into a fail.

There sometimes was some vision about Reason being actually concrete syntax agnostic. The particular "syntax style" might just be user configurable and the _real_ syntax is actually abstract. I think the strategical question for Reason should not be about having mandatory funcall parens and commas (or not) - it should be if all of this could really be abstract! If done this way I might even prefer a full S-Expr-based syntax over the current JavaScript/ML hybrid and others could have their beloved Algol-Style. It might even be possible to represent programs graphically in a non-textual style.

I like this vision. Following this discussion I don't see much of this being really planned. If things go wrong this will be just another point in history about "fixing Lisp".

jaredly:
@neonsquare 
> I prefer that style and I wholeheartedly disagree with anyone who argues that Algol-Style syntax is fundamentally "more readable", "more consistent", "easier to learn/read/use", "more familiar". If you think this way and you think this is something that has to get "fixed" you definitely should think about this being more about you than anyone else.

to sum up your argument: "A is better than B and if you think B is better than you're definitely wrong and it's on you". You're clearly passionate about this, but I don't think your argument came across very clearly. You present it as axiomatic that "lisp is the best and there has never been anything better", but again, you don't provide much support.

I for one am a huge fan of lisp and would love to see an s-exp-based syntax happen for reason, but I don't think the current PR takes us any farther from that than we were before. (ocaml is currently so far from lisp that I don't think comparison makes any sense -- there is a **ton** of syntax that's not s-expressions, and even with function application, see jordan's comments about named arguments, etc.)

As far as "what do we need before this vision is realized?" I think it will happen *after* we've gotten Reason ready for mainstream adoption. I don't want to see us do the Very Hard™ thing of figuring out how to have multiple syntaxes live harmoniously, only to find that no-one will use the language because of fundamental design flaws.

- First: build a solid stdlib, developer tooling, community, etc.
- And then: solve the problems associated w/ having multiple syntaxes.

What we have going in our favor is that ocaml is very well-suited to the "multiple syntaxes" use case, w/ first-class support for preprocessors in the compiler & build tools.

About a year ago, I started to develop an alternate syntax for reason ([here's the code](https://github.com/jaredly/myntax), though it's not well documented), and I stopped **because the tools/community/ecosystem weren't in a good place**, so I couldn't actually build real-world applications. Now we're much closer to that, but we still have a ways to go :D.

If you're interested in building a new syntax, please go for it!! Build the future you want to see, and we'll lend our support.

neonsquare:
@jaredly 
> You're clearly passionate about this, but I don't think your argument came across very clearly. You present it as axiomatic that "lisp is the best and there has never been anything better", but again, you don't provide much support.

no you did understand this completely wrong turned. It was in part a response to your post where you try to argue that c-style syntax is - as you say - axiomatically "the best thing ever and from anything". This is why I said one should start to think about oneself when holding such beliefs. In short: This is subjective and you present it as objective truth.

Dont get me wrong again: I do _not_ argue that it is wrong to go for a c style syntax. It is in the same way not wrong to just leave the paren behaviour like it is or to switch to an even more lisp style syntax. This all could be useable and working styles. THIS is my point. Your arguing in case of a singular style being somewhat superior to others is the wrong direction to argue. I know that alot of people will follow this arguments because they just FEEL that their hard learned Algolism deems natural and familiar to them. This is more about feel and beliefs than about real truth.

So: yes - I can follow an arguing that goes like "we just want to cater for those who really have cognitive dissonance problems about a syntax that looks even trivially different to what they know" - but than call it as that: a catering for those class of people and not a "fix" to a syntax that isnt "more wrong" than the alternatives.

while I can follow the arguing, I'm more sceptical about its value: I think it promises to much. Newbies may find superficial syntactical similarity and be careless about different semantics. I think it will lead to new didactical problems to go this route. Granted: it may solve others. The bigger problem is the message though: The message you propose is "we fixed syntax for you" which closes the mind of people to be open to alternatives. And you already did mention that syntax can matter! Matrices or complex numbers are in a way just syntax in the mathematical domain which made reasoning in certain fields much easier. And at the same way, alternate syntaxes of programming languages (in the sense of not algol style) can make some things easier and better or more natural to work with.

does this make it more clear what I mean?

bsansouci:
@neonsquare @jaredly @cullophid @jordwalke @wincent @TheSpyder Thanks a lot for having this very vivid debate, but I think this isn't the best place to do so. There are a lot of technical things that should be discussed here.

I created [this post](https://discuss.ocaml.org/t/reason-general-function-syntax-discussion/537) to move the discussion over there. I hope you're all ok with that.

let-def:
I added support for docstrings, although printing is not yet satisfactory.

The idea is that a docstring, while lexically looking like a comment with two stars is a sugar for`ocaml.doc` attribute.

`/** a */` is actually a short-hand for `[@ocaml.doc " a "]`.

This means that docstrings should follow the placement rules of attributes. A docstring occurring in a place where attributes are not accepted fall back to being a normal comment.

hcarty:
@let-def Would it be possible to add a warning when a docstring fails to attach properly?

let-def:
@hcarty Good point.

I am not sure how this should be implemented (warning in a preprocessor)?

hcarty:
Could an `ocaml.warning` attribute be attached to an floating/ghost AST node?  It's been a while since I dug into the AST structure so I don't recall if there is anything appropriate to use.  If so then this could piggyback on warning 50.

let-def:
Thanks to @chenglou who rebased this branch on current master.
I am still making progress on the parsing/printing of doc attributes.

let-def:
Apply the same syntax to polymorphic variants:
```reason
`x(5) : [ `x(int) ]
`x(5,5) : [ `x(int,int) ]
```

Conjunctive types:
```reason
[ `x&(int)&(string) ]
[ `x(int,int) &(string,int) ]
```

let-def:
Also, there is a problem with idempotent printing of comments when using ASTs from the OCaml parser.

Doc strings that were not attached to the AST are treated like normal comments. Their only peculiarity is that they start with a `*`.
The first time they get printed as comments.
But when parsed, they are treated as doc strings, and attached as `ocaml.doc` attributes.
Then they get printed with the attachment rules applicable to attributes (or when they are floating inside structures, they get suffixed with `;`).

chenglou:
@let-def so are you saying this is permeant or can be solved? a semicolon after a doc comment is rather bad

hcarty:
The semicolon could be a nice way to indicate an intentionally floating doc comment.

let-def:
It is not a matter of resolving or not... If we treat doc comments as attributes (and we do), then a semicolon after a doc string mean it is floating (unless we special case this one).

The solution for importing ocaml doc comments is to rewrite them by adding a `*`: a invalid doc comments `(** ... *)` in OCaml is printed as `/*** ... */`, which will not be treated as a doc comment in Reason.

hcarty:
Regarding partial application - we already have warnings/errors for partial application in sequences (`;` in OCaml). Could that same machinery be reused to create a "warn on all partial applications" feature, taking advantage of the new syntax in this PR? Then some sugar could be added to disable the warning when partial application is explicitly used and requested.

chenglou:
@let-def semicolon is still the biggest hurdle I've seen from working with newcomers. Btw I'm very, very interested in seeing #1235 in the future. I'd say for now, special-casing.

chenglou:
@let-def I'm getting the following:

```
File "src/reason_toolchain.ml", line 349, characters 51-60:
Error: This expression has type
         string Migrate_parsetree.Ast_404.Location.loc *
         Migrate_parsetree.Ast_404.Parsetree.payload
       but an expression was expected of type
         Parsetree.attribute = string Asttypes.loc * Parsetree.payload
       Type
         Migrate_parsetree.Ast_404.Parsetree.payload =
           Ast_404.Parsetree.payload
       is not compatible with type Parsetree.payload 
```

Freshly updated OPAM, OCaml 4.03.0

chenglou:
Putting this here in case we forget:

- We should special-case `switch ((a, b))` to `switch (a, b)`
- Does this mean array access with `foo[0]` can be unambiguous now?
- `let f(a) = a` is gonna re-trip up lots of folks. Apparently `let f(a) => a` is hard; maybe we should reconsider the sugar and set for the uniform `let f = (a) => a`

let-def:
> Does this mean array access with foo[0] can be unambiguous now?

Ahah yes, though for now I chose to allow `f[0]` and `f{x,y}` as shortcuts for `f([0])` and `f({x,y})`. However `f[0]` for array-access is probably a better use of syntax estate.

let-def:
Ok, I implemented `a[x]` for array access.
This has two side-effects:
- `f[x]` is no longer a sugar for `f([x])` (obviously)
- in JSX lists, we have to choose whether `<tag>a[x] </tag>` means "ident `a` then list `[x]`" or "index `x` of array `a`".

Right now the second interpretation is chosen, which means that a list in a JSX list has to be wrapped between parentheses. I am not familiar enough with JSX to know what is more appropriate.

chenglou:
Latter is fine for now; I don't think I've seen this pattern very often

formatTest/unit_tests/expected_output/modules.re
```diff
@@ -399,7 +399,7 @@ module rec A: {
     | Leaf(string)
     | Node(ASet.t);
   let compare (t1, t2) =
-    switch ((t1, t2)) {
+    switch (t1, t2) {
```
cristianoc: @let-def this is great.
So it was already valid syntax for the parser, but the pretty-printer was not using it.

let-def: Yes exactly.

hcarty:
Could `a{i, j, k}` be used for bigarray access?

chenglou:
@let-def this is still weird:

```
make(:foo bar)[||]
```

The array should be on the inside.

rickyvetter:
I'm concerned about the overloading of semicolon for types and named arguments. I think it will make things confusing coming from typescript/flow. Since names arguments are foreign to JS, would it make sense to keep the OCaml syntax and use ~ to denote a named arg?

chenglou:
Ok folks, time to merge this.

We've cut a release before this PR that includes the pipe formatting PR (a sore refmt bug before, thanks again @IwanKaramazow). It's now live on OPAM. Don't forget, 1.13.7, not 2.0.0. We've decided on a final release _before_ landing this PR because we didn't want it to look like we're conning you into using this syntax just for the pipe formatting =P.

If you were/are against this function syntax change, hopefully you can see that we try our best to make sure folks are happy with most of our changes! And thanks for the discussion so far. We've let it drag on a year because we understand this is a big topic. If we've ignored any opinion, it's accidental.

Additionally, as @let-def pointed out, the change does have some essence beyond pure bikeshedding. If the syntax ends up unpopular, then we can always revisit this decision. But give us some time to iterate on it.

Speaking of iteration: we will be giving you a smooth upgrade path when this is ready. The PR's the first step; the outcome printer, some tests and the error messages need to be fixed. After that, we'll consider changing some others small things, like @rickyvetter's suggestion. I'll then go test on our internal codebase and let my teammates suffer a bit =). Finally, we'll release the upgrade util if things go well.

Thanks again for all the input! And remember: most of you here are Reason "veteran" now. Whether you like this change or not, this PR isn't for you; it's for the future folks you'll be helping.

PR #1304, opened 2017-06-15
Fix printing of module type (precedence of Pmty_with inside Pmty_functor)

let-def:
1) Fix a printing bug when mixing functor types and module type "with" operator.
Add tests for this case.
2) Add printing of module type extensions.


jordwalke:
Great, thanks @let-def. Any other fixes like this that you find are definitely appreciated.

PR #1322, opened 2017-06-29
[Docs] Minor change in installing reason-cli

asolove:
There's a new version of reason-cli (0.13.6) and platform-specific downloads. Rather than have to keep this doc up to date with versions and platforms, let's just link to those.

(I would also be happy to inline the platform-specific download URLs into this doc just to avoid sending people elsewhere, but it would create a need to update it regularly as the CLI is updated.)

### DIFF
--- a/docs/gettingStarted.html
+++ b/docs/gettingStarted.html
@@ -71,10 +71,9 @@ Our [editor integration](./tools.html#editor-integration) need a few binaries to
 
 The tooling section explains what these binaries do.
 
-They can be installed globally through:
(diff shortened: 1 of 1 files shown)


jordwalke:
On the fence. I think we should just inline it and do the work ourselves to update it. By being diligent, we preserve a tiny bit of patience of the person learning the reason toolchain, but patience is a scarce resource.

jordwalke:
Merging this because it's better/more stable than what was currently mentioned in the README, but if anyone likes to you can help us inline and keep it up to date if anyone notices we release a new version. Thanks!

PR #1324, opened 2017-07-02
reason-cli beta-v-1.13.5 -> beta-v-1.13.6

anmonteiro:
Just installed 1.13.6 which is available


anmonteiro:
ooops I guess this is conflicting with #1322

PR #1326, opened 2017-07-03
[Docs] Inline react-cli install steps

asolove:
This resolves the discussion from #1322 and hopefully makes the install process one cognitive step simpler.


jordwalke:
Thank You!

Issue #1328, opened 2017-07-06
Merge reason-parser back into reason


chenglou:
cc @let-def. High pri?

I've discussed with @jordwalke and we've decided to merge reason-parser back into reason (aka kill reason-parser the independent library) to ease contribution and avoid opam pin/unpin and multiple `make`s.

If it's related: we'd like rtop out of the repo into reasonml/.

cc @SanderSpies @IwanKaramazow who've requested this as well.

IwanKaramazow:
Yes!!!!!!!!!!! Finally!!!! 🙌  
This has been bugging me and slowed me down a lot, kept me away from contributing...
The faster this happens, the better.

let-def:
High pri yes, that will make the workflow much easier.

chenglou:
@let-def thanks again. Btw any other related things that'd make your workflow easier?

PR #1330, opened 2017-07-06
merge reason-parser into main source code

let-def:
As requested in issue #1328.

I kept https://github.com/facebook/reason/compare/master...let-def:collapse-reason-parser?expand=1#diff-8d847b939af98498df86c09cf017a952L24 check but I don't think there is any other subpackage, maybe the check can be removed too.


bspacks/README.md
```diff
@@ -93,8 +92,6 @@ You can then run:
     -I "$reasonTargetDir" \
     -I "$reasonTargetDir/_build/src" \
     -I "$reasonTargetDir/vendor/cmdliner" \
-    -I "$reasonParserTargetDir/_build/src" \
-    -I "$reasonParserTargetDir/vendor/easy_format" \
```
chenglou: This needs to be added back as `"$reasonTargetDir/vendor/easy_format"`, otherwise it's not included.

scripts/release-check.sh
```diff
@@ -21,8 +21,7 @@ if [ "${MASTER}" != "${HEAD}" ]; then
 the release. **"
 fi
 
-read -p "STOP! Have you made sure to release the sub packages (including \
-reason-parser)? (y/n) " yn
+read -p "STOP! Have you made sure to release the sub packages? (y/n) " yn
```
chenglou: Yeah I think it's fine to remove it

chenglou:
I can't repro that menhirlib error in the CI. After fixing that `reason_bspack.sh` line the bspacking workflow works (assuming there's an omp folder with the artifacts of migrate-parsetree, I got it from reason-cli's cache). Also I've merged your #1329 first so there's the rebase conflict (sorry).

let-def:
I tried to fix stuff, but it is still failing and I just can't see the message (seems the CI is temporarily broken too :P).

chenglou:
Ok that's weird. Can you fix the merge conflict and push once and see if the CI retriggers and succeed?

chenglou:
Well, at least it's a new error now

let-def:
I suspect the bug is just the regression with latest utop (which hides some modules, see https://github.com/diml/utop/issues/213 ).
I will update rtop to use utop-full.

let-def:
Two bugs: the regression + wrong linking rules.
I hope it will work with these :) (I wonder why it didn't fail before in my local builds...)

chenglou:
Btw we're thinking of pulling out rtop into its own thing. Would that help here?

chenglou:
@let-def any blockers remaining? Eager to get this in!

let-def:
No I would like to get this in, I am trying to understand what is failing.

let-def:
Seems I just forgot to add a file to the repo, hopefully rebuild works this time!

chenglou:
Alright, I've patched it with some last fixes for bspack (which we should really automate and put on CI soon) and verified locally that the bspacking workflow works. Guess we're ready for merging? =D

let-def:
Fine for me :).

chenglou:
Thanks so much =)

cc @IwanKaramazow happy now? =P

PR #1357, opened 2017-08-09
Fix outcome printer object printing

chenglou:
Object used to be printed as `{foo: bar}`, not `{. foo: bar}` or `{..
foo: bar}`. After the object syntax change a while ago I think we didn't
update the outcome printer's printing.

Whitespace-less diff: https://github.com/facebook/reason/pull/1357/files?w=1


chenglou:
¯\_(ツ)_/¯ merging

PR #1365, opened 2017-08-19
Update rtop.sh (mandatory function-call brackets)

dc-mak:


chenglou:
Thanks! And sorry for the delay.

PR #1367, opened 2017-08-21
Add helpful error message for es6 function application paren wrapping.

IwanKaramazow:
Especially helpful for people coming from the pre-es6-fun-era.
Refmt gives you now a helpful error message on `let two = add 1 1;`.


chenglou:
@let-def how would this affect the error message work?

chenglou:
Good enough =D

PR #1371, opened 2017-08-24
Fix arityConversion tests

IwanKaramazow:
Update arityConversion tests
& update mlSyntax.re (4.02.3 test) with new syntax


jordwalke:
Thanks, @IwanKaramazow

PR #1372, opened 2017-08-25
Improved support for js_of_ocaml + OCaml 4.03+

SanderSpies:
I did this so Reason can build be with JSOO on OCaml 4.04.2.

- Remove dependency on utop as this should be optional
- Rename vendored `easy_format.ml` to `easy_format_vendored.ml` to avoid `warning 31`, which became the default from OCaml 4.03+ but can't be disabled with JSOO.
- Fix reason_without_utop.mllib which included duplicate stuff from reasonparser.mllib


chenglou:
Cool. So what happens if folks install Reason and expect rtop? And what about reason-cli

The bspack thing is likely broken. I can fix it later.

jordwalke:
Great! As far as optional dependency on utop, I believe that opam is deprecating that feature (optional dependencies). The test is failing because it can't find `utop`.

We should make rtop its own package though. `reason-cli` can still aggregate all the tools we want but there's no reason why the core Reason tool for syntax should include a command line).

bsansouci:
I want this PR because the profilers I'd like to use require 4.03. (https://github.com/LexiFi/landmarks)

anmonteiro:
no longer relevant

PR #1388, opened 2017-09-06
Fix arityConversion tests

chenglou:
Post syntax change, apparently `let _ = foo` doesn't format to `foo` anymore, just `let _ = foo`. `foo` still idempotently formats to `foo`. This is probably intended? cc @let-def. I've added the extra test cases.


chenglou:
Retrospective reviews welcome

PR #1393, opened 2017-09-08
Better jsx ppx workflow

chenglou:
See https://github.com/BuckleScript/bucklescript/pull/1988

This removes the need to bspack reactjs_jsx_ppx.


chenglou:
Merging on this side. cc @bobzhang

PR #1397, opened 2017-09-08
Deprecate printers

IwanKaramazow:
* Removes derived-printers
* Displays a deprecation message when trying to invoke auto-derived printers with refmt.


chenglou:
I'm super tempted to merge this _right now_. But we need a discussion on a better alternative rather than "you don't even get a half-decent one anymore". cc @bobzhang who has a few ideas.

jordwalke:
I'm supportive of removing them since we don't have anyone volunteering to fix the issues (we could bring them back if we needed them and had a volunteer to improve/mantain). A couple of questions:

Does removing this solve someone's problem right now?
We would want to bump the major version - it's technically breaking.

chenglou:
Yeah we can add them back later.
We'd bump a major version in the next version because of the new syntax anyways, they'd be batched together. Also it's a dev-time thing so it shouldn't be too bad.
I don't think the removal solves anyone's problem right now (and bspack is unaffected since we stubbed this part anyway). But having half-maintained options exposed isn't too good. So I'm merging and counting on @bobzhang to come up with a superb deriving_show =P (no pressure)

Issue #1392, opened 2017-09-07
[Feature request] JSX punning for explicitly passed optional


chenglou:
Pun

```reason
<Foo bar=?bar />
```

into

```reason
<Foo ?bar />
```

chenglou:
Fixed

PR #1401, opened 2017-09-11
Add support for punning of an explicitly passed optional

IwanKaramazow:
https://github.com/facebook/reason/issues/1392

Puns `<Foo bar=?bar />` into `<Foo ?bar />`.


chenglou:
Yeeeeeaaah

Issue #1402, opened 2017-09-11
when clause in pattern matching causes syntax error 


IwanKaramazow:
Works:
```reason
let make (:hoveredCompany, :company, :hoverStart, :hoverEnd, _children) = {
  ...component,
  render: (_self) => {
    open Types.Company;
    let {domain, name, logo} = company;
    <div key=domain onMouseEnter=((_) => hoverStart(name)) onMouseLeave=hoverEnd style>
      <a href=domain target="new"> <img src=logo /> </a>
      (
        switch (hoveredCompany) {
        | Some(c) when c === company.domain => <span> (ReasonReact.stringToElement(name)) </span>
        | _ => ReasonReact.nullElement
        }
      )
    </div>
  }
};
```

Cryptic syntax error:
```
/* same code with */
| Some(c) when c === domain => <span> (ReasonReact.stringToElement(name)) </span>
```

chenglou:
cc @let-def

let-def:
The problem is that this is parsed as a function `domain => <span> (ReasonReact.stringToElement(name)) </span>`, which fails.
The `=>` is ambiguous in this position...

PR #1409, opened 2017-09-15
Allow non-determinism in menhir interpretation loop

let-def:
This should fix https://github.com/facebook/reason/issues/1402 and also enables the `(args) : result_type => result` syntax.

# Why

```
expr ::=
| (expr, ...) // tuple former
| (pattern, ...) => expr // lambda function
```

Lambda function syntax is inherently ambiguous: the parser needs to look after the closing parenthesis to know whether an expression or a pattern should be parsed after the opening parenthesis.

This is beyond what LR parsers can handle, it is in the realm of [GLR parsers](https://en.wikipedia.org/wiki/GLR_parser).
That's not a good thing, but it is in a scope that is limited enough that it should be possible to handle it correctly. Furthermore, either we accept this GLR subset or we give up on ES6-like function syntax.

The current implementation solves the problem at lexing level: it looks for the token after the closing parenthesis, if it is a `=>` then it injects a special token (`ES6_FUN`) before the opening parenthesis.
This way the grammar is LR and Menhir is happy.

However, as discovered by @IwanKaramazow, finding a `=>` is not enough to decide between expressions and patterns. In `when` clauses, we have an expression preceding a `=>`: `when bla => bli` is parsed as `when (fun  bla => bli)`.

Worse, if we add `(args): type => expr` syntax then looking immediately after the closing parenthesis is not enough for deciding: we now have to parse arbitrarily many tokens of a type to look for the `=>`.

# How

The proposed interpreter loop forks the parser. If the lexer determined that an `ES6_FUN` should be injected, the parser will try both cases: with and without the `ES6_FUN` token.
That is it will try to parse the following tokens simultaneously as an expression and as a pattern. Hopefully, one branch will die quickly.

## Performance concerns

This patch can have an impact on performance, although it should be limited: since we are naively forking, an exponential worst case is possible. However the grammar doesn't make it easy to nest forking points, so the actual amount of parallelism is expected to be small (but maybe not bounded).

This is something to keep in mind in future extensions of the grammar!

### DIFF
--- a/src/reason_lexer.mll
+++ b/src/reason_lexer.mll
@@ -863,6 +863,10 @@ and skip_sharp_bang = parse
       tok :: fake_triple ES6_FUN tok :: acc
     | _ -> assert false
 
+  let is_triggering_token = function
+    | EQUALGREATER | COLON -> true
+    | _ -> false
+
   let rec lex_balanced_step closing lexbuf acc tok =
     let acc = save_triple lexbuf tok :: acc in
     match tok, closing with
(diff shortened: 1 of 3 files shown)


chenglou:
Merged! =D Thanks

PR #1412, opened 2017-09-16
Print last arg of function always inside the parens

IwanKaramazow:
Removes the swift-like sugar to print the last argument outside the parens in a function application.

```
/* before */
let x  = myFunc(a, b)();

/* becomes */
let x = MyFunc(a, b, ());
```

(Purely printing, parsing already doesn't support this)

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/attributes.re
+++ b/formatTest/typeCheckedTests/expected_output/attributes.re
@@ -185,7 +185,7 @@ let myObj = {pub p () = {pub z () = 10}};
 
 let result =
   [@onSecondSend]
-  ([@attOnFirstSend] myObj#p(()))#z(());
+  ([@attOnFirstSend] myObj#p())#z();
 
 [@onRecordFunctions]
 type recordFunctions = {
@@ -205,7 +205,7 @@ and unused = ();
 
 let result =
   [@onSecondSend]
-  ([@attOnFirstSend] myRecord.p(())).q(());
+  ([@attOnFirstSend] myRecord.p()).q();
 
 [@onVariantType]
 type variantType =
--- a/formatTest/typeCheckedTests/expected_output/attributes.re
+++ b/formatTest/typeCheckedTests/expected_output/attributes.re
@@ -185,7 +185,7 @@ let myObj = {pub p () = {pub z () = 10}};
 
 let result =
   [@onSecondSend]
-  ([@attOnFirstSend] myObj#p())#z();
+  ([@attOnFirstSend] myObj#p(()))#z(());
 
 [@onRecordFunctions]
 type recordFunctions = {
@@ -205,7 +205,7 @@ and unused = ();
 
 let result =
   [@onSecondSend]
-  ([@attOnFirstSend] myRecord.p()).q();
+  ([@attOnFirstSend] myRecord.p(())).q(());
 
 [@onVariantType]
 type variantType =
--- a/formatTest/typeCheckedTests/expected_output/imperative.re
+++ b/formatTest/typeCheckedTests/expected_output/imperative.re
@@ -63,7 +63,7 @@ let shouldStillLoop = {contents: false};
 
 while (shouldStillLoop.contents) {
   print_string("You're in a while loop");
-  print_newline()
+  print_newline(())
 };
 
 while ({
--- a/formatTest/typeCheckedTests/expected_output/imperative.re
+++ b/formatTest/typeCheckedTests/expected_output/imperative.re
@@ -63,7 +63,7 @@ let shouldStillLoop = {contents: false};
 
 while (shouldStillLoop.contents) {
   print_string("You're in a while loop");
-  print_newline(())
+  print_newline()
 };
 
 while ({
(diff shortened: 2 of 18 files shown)


chenglou:
Can you also paste the repro that shows this caused an actual bug?

Cc @let-def

IwanKaramazow:
```reason
/* before on old syntax*/
let el =
  ReasonReact.element (
    WedstrijdAankomstRow.make
      plaats::curr.plaats
      voornaam::curr.voornaam
      naam::curr.naam
      afstand::curr.afstand
      tijd::curr.tijd
      [||]
  );

/* After refmt to the new syntax */
let el =
  ReasonReact.element(
    WedstrijdAankomstRow.make
      (
        :plaats curr.plaats,
        :voornaam curr.voornaam,
        :naam curr.naam,
        :afstand curr.afstand,
        :tijd curr.tijd
      )
      [||]
  );
```
The latter wouldn't parse, i.e. resulted in a syntax error with `[||]` outside of the parens.

IwanKaramazow:
Added an explicit test case in the tests.

Steps to reproduce:
```
(* errors *)
echo "let x = myFun(1, 2)[||];" | ./refmt_impl.native
(* is correct*)
echo "let x = myFun(1, 2, [||]);" | ./refmt_impl.native
```

let-def:
Indeed, parsing didn't allow arrays in last position (however, `()` and objects are supported).
The fix is good.

formatTest/typeCheckedTests/expected_output/attributes.re
```diff
@@ -185,7 +185,7 @@ let myObj = {pub p () = {pub z () = 10}};
 
 let result =
   [@onSecondSend]
-  ([@attOnFirstSend] myObj#p())#z();
+  ([@attOnFirstSend] myObj#p(()))#z(());
```
chenglou: Is this expected?

formatTest/typeCheckedTests/expected_output/imperative.re
```diff
@@ -63,7 +63,7 @@ let shouldStillLoop = {contents: false};
 
 while (shouldStillLoop.contents) {
   print_string("You're in a while loop");
-  print_newline()
+  print_newline(())
```
chenglou: Same for this

IwanKaramazow: I removed all sugar regarding the last argument getting special printing, do we keep the last argument sugar when there is:
1) one argument
2) its type is unit

chenglou: Definitely! `foo(())` isn't great at all lol

IwanKaramazow:
Implemented the `myFun(())` -> `myFun()` printing

src/reason_pprint_ast.ml
```diff
@@ -6301,27 +6301,14 @@ class printer  ()= object(self:'self)
...
-      | (Nolabel, x) :: xs when is_direct x ->
-        ([self#simplifyUnparseExpr x], List.rev xs)
-      | _ -> ([], xs)
-    in
-    let xs = if xs = [] then [] else
-        [makeTup (List.map self#label_x_expression_param xs)]
+    let xs = (match xs with
+      (* function applications with unit as only argument should be printed differently
+       * e.g. print_newline(()) should be printed as print_newline() *)
+      | [(_, ({pexp_attributes = []; pexp_desc = Pexp_construct ( {txt= Lident "()"}, None)} as x))]
```
let-def: You should restrict the rewriting to the case where the label is `Nolabel`.

IwanKaramazow: Indeed, thx!

formatTest/typeCheckedTests/expected_output/oo.re
```diff
@@ -162,20 +162,19 @@ let acceptsClosedAnonObjAsArg
...
+    {pub x = 0; pub y = 10}
+  );
 
 let res =
-  acceptsOpenAnonObjAsArg{
-    pub x = 0;
-    pub y = 10;
-    pub z = 10
-  };
+  acceptsOpenAnonObjAsArg(
```
chenglou: What happens with bigger objs?

formatTest/unit_tests/expected_output/basicStructures.re
```diff
@@ -643,10 +643,12 @@ let anotherRecord = {
 
 let anotherRecord = {
   ...
-    SomeReally.longFunctionCall{
-      passingRecordField: 0,
-      andThisOtherRecordField: 10
-    },
+    SomeReally.longFunctionCall(
```
chenglou: Is it possible to have this formatted to: 

```reason
foo({
  a: 1
})
```

We can do that in another diff. Just wondering if it's easy.

formatTest/typeCheckedTests/expected_output/oo.re
IwanKaramazow: something not so attractive
```
acceptsOpenAnonObjAsArg(
  {
    /* object stuff */
  }
)
```

formatTest/unit_tests/expected_output/basicStructures.re
IwanKaramazow: Not super easy, probably medium difficulty.
Refmt has a 'unparsing' algorithm which needs to be modified a bit. In essence that would allow us the stick `({` together.

chenglou:
K this is good enough, merged. Thanks!

PR #1413, opened 2017-09-16
work around doc comments in 4.02.3 (fix #1398)

### LINKED ISSUE #1398: Doc comments are different in 4.02.3
There's a CI issue, with doc comments appearing double in 4.02.3.
Doc-comments get filtered out of the comments, because they appear as attributes in the 
parsetree. And are formatted starting from the attribute in the printer, not as comment.
I tracked it down to:
https://github.com/facebook/reason/blob/master/src/reason_toolchain.ml#L348

On 4.02.3 `Hashtbl.add seen (text, loc) ();` properly filters out comments.
While for example on 4.04.2 we need `Hashtbl.add seen ("*" ^ text, loc) ();` to properly filter out comments.

@let-def is there a reason why the behaviour on 4.02.3 is different?

let-def:


chenglou:
I don't see the changes in the CI here: https://travis-ci.org/facebook/reason/jobs/276280707?

let-def:
I have a simpler fix... Assuming the location of a comment is unique, the matching between comment-as-string and comment-as-attribute is done solely based on the location.

chenglou:
I see that the tests are a little bit different now. All three comments files are still failing, but is this the final change for this fix?

PR #1420, opened 2017-09-22
Change JSX to desugar from Foo.createElement to Foo.make

chenglou:
More idiomatic, moves us closer to a future without needing a ppx after the jsx.


cristianoc:
@chenglou this change is for lowercase jsx right?
Some of the comments mention "from Foo.createElement to Foo.make", while this is already the case for uppercsee jsx.

chenglou:
Lowercase jsx doesn't use `make`. `<div />` desugars to `div children::[] [@JSX]`

Issue #1428, opened 2017-09-30
JSX: support `<Foo> ...bar </Foo>`


chenglou:
This would pass a reference. So instead of:

```reason
Foo.createElement children::[bar] ()[@JSX];
```

We'd have

```reason
Foo.createElement children::bar ()[@JSX];
```

This solves a _lot_ of problems. cc @IwanKaramazow

Note: this is not _that_ intuitive, considering that `...` kinda suppose `bar` is a list. It might not be at all.

Let's revisit this after JSX3

PR #1429, opened 2017-09-30
implement spread for jsx3

IwanKaramazow:
https://github.com/facebook/reason/issues/1428

```reason
<Foo> ...bar </Foo>;

<Foo> ...((a) => 1) </Foo>;
```

Is parsed as:
```ocaml
(Foo.createElement ~children:bar ()) [@JSX]);;
(Foo.createElement ~children:(fun a  -> 1) ())[@JSX];;
```


chenglou:
Does this work with not just `...foo` but also `...<div />`? Aka `<Foo> ...<div /></Foo>` -> `Foo.createElement children::(<div />) () [@JSX]`

IwanKaramazow:
Supported:
```reason
<Foo> ...bar </Foo>;
<Foo> ...((a) => 1) </Foo>;
<Foo> ...<div /> </Foo>;
<Foo> ...[|a|] </Foo>;
<Foo> ...(1, 2) </Foo>;
```
Desugars to the following in the old syntax:
```reason
Foo.createElement children::bar () [@JSX];
Foo.createElement children::(fun a => 1) () [@JSX];
Foo.createElement children::<div /> () [@JSX];
Foo.createElement children::[|a|] () [@JSX];
Foo.createElement children::(1, 2) () [@JSX];
```

It **does not ** support: 
```
<Foo> ...[x] </Foo>;
```

Should I add support for `...[]` ?

chenglou:
Ehhh good one, I think we should? What do you think?

jaredly:
Yeah it would be weird to be inconsistent

IwanKaramazow:
Added the `<Navbar> ...[myList] </Navbar>` variant.
Note that this will reformat to a single child because their equivalent.
```ocaml
<Navbar> ...[myList] </Navbar>;
(Navbar.createElement ~children:[myList] ())[@JSX]
<Navbar> myList </Navbar>;
(Navbar.createElement ~children:[myList] ())[@JSX]
```

List syntax inside jsx children without spread operator `<Navbar> [list1] [list2] </Navbar>`, is broken atm & will be addressed in a different PR.

chenglou:
Hah, that's troublesome now. Because from the react jsx ppx perspective, `<Foo> ...[bar] </Foo>` is definitely not the same as `<Foo> bar </Foo>`. Latter's an array...

chenglou:
Non-idempotent formatting isn't great. If this is the only solution then I think we should just format `<Foo> ...[bar] </Foo>` to `<Foo> bar </Foo>`. And then hopefully people don't actually use the former with the reactjs jsx ppx

jaredly:
Hmmm yeah are you sure `<Bar>...[x}</Bar>` would refmt to `<Bar>x</Bar>`?

jordwalke:
This wouldn't be "non-idempotent" as @IwanKaramazow described. I think it would be exactly what you would expect. `<X>...any_expression</X>` means "the children should exactly be the expression `any_expression`. If you happen to supply `<x>...[literal, list]</x>`, then the `...[ ]` was not necessary in the first place and would then be equivalent to `<x>literal list</x>`.

Issue #1432, opened 2017-10-02
Reliance on older topkg version prevents use of several libraries under opam


hcarty:
Reason can't be used with the latest versions of bos (http://erratique.ch/software/bos) among other libraries by dbuenzli.  bos in particular has important bugfixes in more recent versions which are not accessible from Reason project due to Reason's topkg version constraint.

chenglou:
Fixed in #1438

PR #1438, opened 2017-10-03
Tweaks to build on 4.05.0 and latest topkg

hcarty:
Tested with the most recent menhir so the constraints are slightly
relaxed there as well.

I haven't tested against the just-released OCaml 4.06.0 beta so the OCaml version constraint is left on the paranoid side.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ setup_convenient_bin_links:
 
 precompile:
 	cp pkg/META.in pkg/META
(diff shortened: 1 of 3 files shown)


hcarty:
Fixes #1432

chenglou:
Merged! Thanks =D

jordwalke:
ah thank you very much!

hcarty:
Happy to contribute!

Issue #1446, opened 2017-10-04
Make labeled function application use `=`


chenglou:
Current master syntax:

```reason 
let five = add(:x 3, :y 2);
let five = add(:x 3: int, :y 2: int);
let foo = bar(:x? Some(1));
let foo = bar(:x? Some(1): option(int));
```

Goal:

```reason 
let five = add(:x=3, :y=2);
let five = add(:x=3: int, :y=2: int);
let foo = bar(:x=?Some(1));
let foo = bar(:x=?Some(1): option(int))
```

**No syntax opinion debates please**, and don't advertise this too much. Right now, we need to focus on shipping this syntax to unblock people. Syntax bikeshedding are reserved for the next breaking change.

Issue #1444, opened 2017-10-04
Make function type declaration with default value use `=`


chenglou:
Current master syntax:

```reason
type foo = (:first int?, :second int?) => int
```

Goal:

```reason
type foo = (:first: int=?, :second: int=?) => int
```

This is another step toward making function application & declaration & type with labeled default more consistent. Related: #1446

**No syntax opinion debates please**, and don't advertise this too much. Right now, we need to focus on shipping this syntax to unblock people. Syntax bikeshedding are reserved for the next breaking change.

Issue #1443, opened 2017-10-04
Make labeled function declaration renaming use `as`


chenglou:
Current master syntax:

```reason
let add = (:first x, :second y) => x + y;
```

Goal:

```reason
let add = (:first as x, :second as y) => x + y;
```

Make sure this doesn't clash with the other usages of `as`. The keyword will be a bit overloaded now.

**No syntax opinion debates please**, and don't advertise this too much. Right now, we need to focus on shipping this syntax to unblock people. Syntax bikeshedding are reserved for the next breaking change.

chenglou:
It did. You can try them in the playground! https://reasonml.github.io/try/?reason=DYUwLgBA5gTi4QLwQBQD8B2BDAtiCWAzhBgJRIB8EARABIjDAD2NEA1GyQNxA

PR #1448, opened 2017-10-04
Implement #1443, #1444 & #1446

IwanKaramazow:
Implements https://github.com/facebook/reason/issues/1443, Implements https://github.com/facebook/reason/issues/1444 & https://github.com/facebook/reason/issues/1446

```reason
/* 1443 */
let add = (:first as x, :second as y) => x + y;

/* 1444*/
type foo = (:first: int=?, :second: int=?) => int;
type foo = (:first: int, :second: int) => int;

/* 1446 */
let five = add(:x=3, :y=2);
let five = add(:x=3: int, :y=2: int);
let foo = bar(:x=?Some(1));
let foo = bar(:x=?Some(1): option(int))
let explictlyPassed = myOptional(:a?, :b=?None);
```

### DIFF
--- a/formatTest/unit_tests/expected_output/syntax.re
+++ b/formatTest/unit_tests/expected_output/syntax.re
@@ -844,10 +844,10 @@ let explictlyPassedAnnotated: int =
 /*U: Explicitly passing optional with identifier expression */
 let a = None;
 
-let explictlyPassed = myOptional(:a=?, :b=?None);
+let explictlyPassed = myOptional(:a?, :b=?None);
 
 let explictlyPassedAnnotated: int =
-  myOptional(:a=?, :b=?None);
+  myOptional(:a?, :b=?None);
 
 let nestedLet = {
   let _ = 1;
--- a/formatTest/unit_tests/expected_output/syntax.re
+++ b/formatTest/unit_tests/expected_output/syntax.re
@@ -759,9 +759,9 @@ let named (:a, :b) = a + b;
 type named = (:a int, :b int) => int;
 
 /*B*/
-let namedAlias (:a aa, :b bb) = aa + bb;
+let namedAlias (:a as aa, :b as bb) = aa + bb;
 
-let namedAlias (:a aa, :b bb) = aa + bb;
+let namedAlias (:a as aa, :b as bb) = aa + bb;
 
 type namedAlias = (:a int, :b int) => int;
 
@@ -769,7 +769,8 @@ type namedAlias = (:a int, :b int) => int;
 let namedAnnot (:a: int, :b: int) = 20;
 
 /*D*/
-let namedAliasAnnot (:a aa: int, :b bb: int) = 20;
+let namedAliasAnnot
+    (:a as aa: int, :b as bb: int) = 20;
 
 /*E*/
 let myOptional (:a=?, :b=?, ()) = 10;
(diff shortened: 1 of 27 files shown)


chenglou:
Holy crap that was fast... Thanks 

Explicitly passed punning should be `foo(:a?)`, not `foo(:a=?)` I think

IwanKaramazow:
I'll bring `foo(:a=?)` to `foo(:a?)`.

For the second case I tend to follow Ocaml personally `add ~foo:(10 : int)`. Type signature always last, but it might trip people up. What do you think?
Should we ask around?

chenglou:
Type last. It's fine yeah

formatTest/unit_tests/expected_output/syntax.re
```diff
@@ -863,23 +874,26 @@ let nestedLet = {
...
 type typeWithNestedOptionalNamedArgs =
   (
-    :outerOne (:innerOne int, :innerTwo int) =>
-              int
-                ?,
-    :outerTwo int?
+    :outerOne: (:innerOne: int, :innerTwo: int) =>
+               int
+                 =?,
+    :outerTwo: int=?
```
chenglou: This formatting.................

formatTest/unit_tests/input/jsx.re
```diff
@@ -32,15 +32,15 @@ let y =
...
-              ::someOtherAttribute,
+              :width,
+              :height,
+              :color,
+              :backgroundColor,
+              :margin,
+              :padding,
+              :border,
+              :borderColor,
+              :someOtherAttribute,
```
chenglou: How come it's indented like this?

chenglou:
Left two unrelated printing comments. This is great, thanks so much

formatTest/unit_tests/input/jsx.re
IwanKaramazow: Because it's the *input* of the test & we make sure the indentation is correct in the * expected output ⛷  😄

formatTest/unit_tests/expected_output/syntax.re
```diff
@@ -57,59 +57,68 @@ type myRecordType = {
...
-  (:first int => int?, :second list(int)?) => int;
+  (:first: int => int=?, :second: list(int)=?) =>
+  int;
 
 /* The arrow necessitates parens around the next two args. The ? isn't what
  * makes the parens necessary. */
 type firstNamedArgShouldBeGroupedInParensAndSecondNamedArg =
-  (:first int => int?, :second int => int?) =>
+  (
+    :first: int => int=?,
```
bsansouci: Is there an easy way to avoid the newline here?
```
type blablablablablablablabla = 
  (:firstVeryLongNameHere: int => int=?, 
   :secondOtherVeryLongNameBecauseWeNeedToWrap: int => int=?) => int;
```
Something like that?

Not super attached though, don't worry about it if it'll take more than like 10min.

IwanKaramazow: You probably only want this in type signatures?

formatTest/unit_tests/input/jsx.re
chenglou: Oops...

formatTest/unit_tests/expected_output/syntax.re
bsansouci: Mmmh true. I don’t think it’s worth having the type wrap differently than the definition or calls, so nvm forget it ;)
Great job dude

PR #1456, opened 2017-10-05
Record punning for value & pattern for fields with module prefix

### LINKED ISSUE #1431: refmt removes punning when record field is prefixed by a module name
Reason 1.13.7:
```
$ echo 'fun {M.x, y} => x + y' | refmt
fun {M.x: x, y} => x + y;
```

chenglou:
Aka `{M.x: x, y: 1}` puns to `{M.x, y: 1}`.

Rather positive the expectation isn't that it's a shorthand for `{M.x:
M.x, y: 1}`...

Fixes #1431

### DIFF
--- a/formatTest/unit_tests/expected_output/basicStructures.re
+++ b/formatTest/unit_tests/expected_output/basicStructures.re
@@ -691,6 +691,11 @@ type component2 = {
   updater: unit
 };
 
+type component3 = {
+  props: M.props,
+  state
+};
+
 type mutableComponent = {mutable props};
 
 type mutabeleComponent2 = {
@@ -708,6 +713,30 @@ type description('props) = {
 /* Don't pun types from other modules */
 module Foo = {
   type bar = {foo: Baz.foo};
+};
+
+/* record value punning */
+let props = {title: "hi"};
+
+/* no punning available for a single field. Can't tell the difference with a scope + expression */
+let componentA = {props: props};
+
+/* pun for real */
+let componentB = {props, state: ()};
+
+/* pun fields with module prefix too */
+let foo = {Foo.foo: foo};
+
+let bar = {Foo.foo, bar: 1};
+
+let bar = {bar: 1, Foo.foo};
+
+let bar = {Foo.foo, Bar.bar};
+
+({M.x, y}) => 1;
+
+switch (foo) {
+| {y: 1, M.x} => 2
 };
 
 /* Requested in #566 */
--- a/formatTest/unit_tests/input/basicStructures.re
+++ b/formatTest/unit_tests/input/basicStructures.re
@@ -562,6 +562,8 @@ type component = {props};
 
 type component2 = {props, state, updater: unit,};
 
+type component3 = {props: M.props, state};
+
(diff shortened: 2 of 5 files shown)


src/reason_pprint_ast.ml
```diff
@@ -4932,8 +4934,9 @@ class printer  ()= object(self:'self)
         loc_ghost = false;
       } in
       let theRow = match e.pexp_desc with
-        (* Punning *)
-        |  Pexp_ident {txt} when li.txt = txt && shouldPun && allowPunning ->
+        (* record value punning. Turns {foo: foo, bar: 1} into {foo, bar: 1} *)
+        (* also turns {Foo.bar: bar, baz: 1} into {Foo.bar, baz: 1} *)
+        |  Pexp_ident {txt = ident} when Longident.last li.txt = Longident.last ident && shouldPun && allowPunning ->
```
chenglou: @hcarty the `&& shouldPun && allowPunning` part is better lifted up as `match (e.pexp_desc, shouldPun, allowPunning)`, right?

formatTest/unit_tests/expected_output/basicStructures.re
```diff
@@ -691,6 +691,11 @@ type component2 = {
   updater: unit
 };
 
+type component3 = {
+  props: M.props,
```
bsansouci: Should the types also be punned?

formatTest/unit_tests/input/basicStructures.re
```diff
@@ -577,5 +579,24 @@ module Foo = {
   type bar = {foo: Baz.foo};
 };
 
+/* record value punning */
+
+let props = {title: "hi"};
+/* no punning available for a single field. Can't tell the difference with a scope + expression */
+let componentA = {props: props};
+/* pun for real */
+let componentB = {props: props, state: ()};
```
bsansouci: Why is props not punned here

bsansouci: Oh nvm this is input file lel

formatTest/unit_tests/expected_output/basicStructures.re
chenglou: cc @hcarty. What do you think? I didn't do it because `{M.x: x, y: 1}` has the `Foo` part on the left and it was less intuitive punning the type when `M` is on the right

src/reason_pprint_ast.ml
hcarty: I think so - a bit easier to refactor safely in the future.

formatTest/unit_tests/expected_output/basicStructures.re
hcarty: We're referencing the `M.t` type here for use as a field here, not extending the type itself. I agree with @chenglou that it would be better to not allow punning in this case. 

That may be worth considering as syntax for the 4.06.0 feature supporting inherited fields in object types.

bsansouci: Oh I see I see. Good point.

PR #1458, opened 2017-10-05
Fix wrongly failing tests on 4.02

chenglou:


chenglou:
cc @jaredly thanks for pointing this out.

PR #1462, opened 2017-10-05
Add test output - fix one failing test

jaredly:
Looks like some code was added to the input, but we forgot to add it to
the expected.


chenglou:
I made the same change in https://github.com/facebook/reason/commit/e48ad0ea008c8f1c9df21efcfed7afba812f08e4#diff-b8944fab4569d2cb48b6513028e30696

Symbolically merging this one, thanks for pointing it out =D

Issue #1435, opened 2017-10-03
Better dereference sugar


chenglou:
Right now in master `foo.contents` has the sugar `foo^`, which is eeeeh, not ideal. It'd be nice if `*foo` was supported (might clash a lot), or that we at least change it to `^foo`. Not sure what the latter is supposed to evoke though; seems arbitrary.

jaredly:
I'm looking into this

jaredly:
Hmmmmmm I don't think `*` will work actually, because there's probably not a way to distinguish `*foo` (deref) from `(*)foo` (partial application of multiplication) at a language level :(

jordwalke:
I'm not exactly sure if this is launch blocking for the new syntax. I think we can use `.contents` in the mean time and if we come up with something better, we can add it in at that time.

let-def:
I remember this syntax was suggested because it was deferencing syntax from pascal.
It seems much more natural to have postfix syntax for dereferencing (as this works with "method-chaining"), while the prefix syntax from C forces a lot of parentheses.

I don't see why `postfix^` is now a problem? (a part from the implications of having postfix operators).

jaredly:
yeah let's not block launch on this. I agree that postfix seems to make more sense.

chenglou:
Yeah I think this is fine.

PR #1463, opened 2017-10-05
Make deref be a prefix operator

jaredly:
I can see some logic in having it be postfix -- for example
```
foo^.bar^
```
feels a little better than
```
^(^foo).bar
```

(this is on top of https://github.com/facebook/reason/pull/1462)
fixes https://github.com/facebook/reason/issues/1435

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/arityConversion.re.4.02.3
+++ b/formatTest/typeCheckedTests/expected_output/arityConversion.re.4.02.3
@@ -28,6 +28,12 @@ Test.Or((1, 2));
 
 Some(1);
 
+Test.And((1, 2));
+
+Test.Or((1, 2));
+
+Some(1);
+
 module M = {
   type t =
     | TupleConstructorInModule((int, int));
@@ -42,6 +48,16 @@ type t2 =
 
 type t3 =
   | TupleConstructor3((int, int));
+
+M.TupleConstructorInModule((1, 2));
+
+M.TupleConstructor2((1, 2));
+
+TupleConstructor2((1, 2));
+
+M.TupleConstructor3((1, 2));
+
+TupleConstructor3((1, 2));
 
 M.TupleConstructorInModule((1, 2));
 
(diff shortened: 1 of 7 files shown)


hcarty:
That `foo^.bar^` example seems like a very good argument for keeping it postfix.  It's far easier to read than than the equivalent prefix example.

chenglou:
So not `*`?

IwanKaramazow:
Just an idea, but what if we solve this in lexing?
e.g. emit something like `STARLIDENT<string>`  ?

hcarty:
Would parens be required for `v * (*r)`?

IwanKaramazow:
No idea what would happen with the `STARLIDENT<string>` for `v * (*r)` if that even works...
Maybe it's worth a try?

jaredly:
I was able to get `2 * *r` working w/o a hitch, but then I was like "wait does `let x = (*) r` mean 'deref r' or 'partially apply r to multiplication'."

jaredly:
I guess we could make `*r` be an exception, where it literally is a syntax transform to `r.contents`, and refmt turns `r.contents` into `*r`. But that's a little weird (also would be a weirdly polymorphic transformation)

jaredly:
Hmmmmk actually I think I can get "prefix *" to parse into dereference

chenglou:
That's too bad, no `*` then. `^` is still rather weird. Any other suggestion? Maybe even just `foo.contents`?

A bit torn on prefix vs postfix too.

jordwalke:
I just always do `foo.contents` *shrug*

hcarty:
Is it worth removing the sugar for setting the value in a reference too? That may help keep it clear to users that a reference is "just" a pre-defined record type with a single, mutable field. Usage would always look like:

```reason
let foo = ref 0;
foo.contents = 1;
print_int foo.contents;
```

jaredly:
I no longer thing prefix is better, and I'll probably get used to `^`

Issue #1442, opened 2017-10-04
Remove the format `let foo(bar) = 1` from the printer


chenglou:
Context: currently the new master syntax sugars `let foo = (bar) => 1` into `let foo(bar) = 1`. This has tripped up lots of folks testing the new syntax. To save on time, we can first remove the shortcut from the printer for now, and keep accepting the syntax in the parser.

**No syntax opinion debates please**, and don't advertise this too much. Right now, we need to focus on shipping this syntax to unblock people. Syntax bikeshedding are reserved for the next breaking change.

jaredly:
Does this mean we'd refmt `let foo(bar) = 1` into `let foo = (bar) => 1`?

chenglou:
Yes. But the former is only allows because I'd like us to ship a bit faster. If we can remove it from the parser then all the better

PR #1469, opened 2017-10-07
Print functions as javascript

IwanKaramazow:
https://github.com/facebook/reason/issues/1442

```reason
/* before */
let foo (bar) = 1;
let myObj = {pub p () = 10};

/* after */
let foo = (bar) => 1;
let myObj = {pub p = () => 10};
```

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/attributes.re
+++ b/formatTest/typeCheckedTests/expected_output/attributes.re
@@ -120,9 +120,9 @@ let x = [@attrEverything] (true && false);
 /**
  * How attribute parsings respond to other syntactic constructs.
  */
-let add (a) = [@onRet] a;
+let add = (a) => [@onRet] a;
 
-let add (a) = [@onRet] a;
+let add = (a) => [@onRet] a;
 
 let add = [@onEntireFunction] ((a) => a);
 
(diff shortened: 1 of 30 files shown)


chenglou:
Should we do the same for functors?

IwanKaramazow:
Yes, we should, since they're "functions" over modules.

Still struggling with a idempotency problem at the moment with functors.

```bash
Iwans-MBP:reason Iwan$ cat functor.ml
module Increment (M : X_int) : X_int = struct
    let x = M.x + 1
  end;;
Iwans-MBP:reason Iwan$ ./refmt_impl.native --parse ml --print re functor.ml
module Increment = (M: X_int) : X_int => {
  let x = M.x + 1;
};
Iwans-MBP:reason Iwan$ ./refmt_impl.native --parse ml --print re functor.ml | ./refmt_impl.native --parse re --print re
File "", line 1, characters 30-31:
Error: 2149: <UNKNOWN SYNTAX ERROR>
```

```reason
/* valid parse */
module Increment = (M: X_int) => {
  let x = M.x + 1;
};

/* invalid parse */
module Increment = (M: X_int) : X_int => {
  let x = M.x + 1;
};
```

chenglou:
Aaah, is this related to https://github.com/facebook/reason/pull/1409?

IwanKaramazow:
it might be, parser looks pretty ok...
```reason
/* valid parse... */
module Increment  (M: X_int) :X_int {
  let x = M.x + 1;
};
```
Time to call the cavalry?

chenglou:
cc @let-def 

I'm merging this as-is for now then. Let's fix the functor printing and parsing #1472.

Thanks!

Issue #1427, opened 2017-09-30
refmt does not format == correctly in interfaces


hcarty:
Using OCaml 4.04.2 and Reason 1.13.7:
```
$ echo 'val ( = ) : int -> int -> bool' | refmt --parse=ml --print=re --interface=true
let (=): int => int => bool;
```
The output should be `(==)` rather than `(=)`.  This seems to be something parser-related as the following will not compile, saying that `==` is required but not provided:

`foo.re`
```reason
let (==) => (==);
```
`foo.rei`
```reason
let (==): int => int => bool;
```
It makes it impossible to properly expose a `(==)` function through an interface.

chenglou:
Fixed by #1474

PR #1474, opened 2017-10-08
transform ocaml pervasives interfaces to reason correctly

GuillaumeSalles:
Currently on master:

```
$ echo "val ( = ) : 'a -> 'a -> bool" | refmt --parse=ml --print=re --interface=true
```

Actual result
```
let (=): ('a, 'a) => bool;
```

Expected result
```
let (==): 'a => 'a => bool;
```

I think there is 2 issues here :
- `(=)` should mapped to `(==)`
- ~~`'a -> 'a` should not become a tupple~~

This PR fixes only the first one. ~~The second one seems to be a regression.~~

Currently the Pervasives reason documentation page is misleading and I think this PR will fix it. E.g. https://reasonml.github.io/api/Pervasives.html#VAL(=)

And probably that one too #1427

I'm still a new to OCaml / Reason so feel free to tell me if I misunderstood something.

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/pervasive.rei
+++ b/formatTest/typeCheckedTests/expected_output/pervasive.rei
@@ -1,0 +1,5 @@
+let (==): ('a, 'a) => bool;
+
+let (!=): ('a, 'a) => bool;
+
+let (!): bool => bool;
--- a/formatTest/typeCheckedTests/input/pervasive.mli
+++ b/formatTest/typeCheckedTests/input/pervasive.mli
@@ -1,0 +1,5 @@
+val ( = ) : 'a -> 'a -> bool
+
+val ( <> ) : 'a -> 'a -> bool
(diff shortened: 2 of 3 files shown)


hcarty:
This looks like it may fix #1427 - thank you!

The tuple-looking syntax is due to a major syntax change in `master`. See #1299 for more details.

GuillaumeSalles:
> The tuple-looking syntax is due to a major syntax change in master. See #1299 for more details.

Oh I see! Thanks for the info.

chenglou:
Looks good! Thanks, I'll merge it. This indeed fixes #1427.

You can map over signature_item directly btw: http://caml.inria.fr/pub/docs/manual-ocaml/libref/Ast_mapper.html#TYPEmapper

No need to pick `signature` then do `List.map` on it yourself. I'll correct it.

PR #1476, opened 2017-10-09
fix #1422

### LINKED ISSUE #1422: Bring back parentheses-less `switch foo`
@let-def how hard is this? Does this also goes into GLR territory? Lots of people on the current syntax will trip on it I think.

Also, if we can't support it, we should at least give it a good error message. Lots of existing users would be confused by `switch foo` failing

let-def:
A possible solution to https://github.com/facebook/reason/issues/1422

### DIFF
--- a/src/reason_parser.mly
+++ b/src/reason_parser.mly
@@ -2541,11 +2541,11 @@ mark_position_exp
    */
   | FUN match_cases(expr) %prec below_BAR
     { mkexp (Pexp_function $2) }
-  | SWITCH parenthesized_expr LBRACE match_cases(semi_terminated_seq_expr) RBRACE
+  | SWITCH simple_expr_no_constructor LBRACE match_cases(semi_terminated_seq_expr) RBRACE
     { mkexp (Pexp_match ($2, $4)) }
-  | TRY parenthesized_expr LBRACE match_cases(semi_terminated_seq_expr) RBRACE
+  | TRY simple_expr_no_constructor LBRACE match_cases(semi_terminated_seq_expr) RBRACE
     { mkexp (Pexp_try ($2, $4)) }
-  | TRY parenthesized_expr WITH error
+  | TRY simple_expr_no_constructor WITH error
     { syntax_error_exp (mklocation $startpos($4) $endpos($4)) "Invalid try with"}
   | IF parenthesized_expr simple_expr ioption(preceded(ELSE,expr))
     { mkexp(Pexp_ifthenelse($2, $3, $4)) }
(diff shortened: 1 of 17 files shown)


let-def:
This solution fixes the syntax. However it introduces a new variant of `simple_expr`, `simple_expr_no_constructor`, which doesn't allow things like `Some{x,y}`.

The ambiguity is because of the `{`: `switch Some{x,y} { ...`
This has to be written `switch (Some{x,y}) { ...`

A compromise could be `switch Some({x,y}) { ...` which is not enabled.

let-def:
The pretty-printer is not updated yet.

chenglou:
Related: can we print `Some {x: 1}` to `Some({x: 1})` for now? This way if we ever backtrack on the former syntax sugar, it wouldn't be much of a breaking change to upgrade folks.

IwanKaramazow:
Printer already prints it that way on master.
```bash
echo "Some {x: 1};" | ./refmt_impl.native
Some({x: 1});
```

chenglou:
Sigh, thanks. Seems like I've been using the wrong binary again. My environment's a mess

src/reason_parser.mly
```diff
@@ -2797,8 +2777,36 @@ parenthesized_expr:
 
 %inline simple_expr: simple_expr_call { mkexp_app_rev $startpos $endpos $1 };
 
+simple_expr_no_constructor:
+  mark_position_exp(simple_expr_template(simple_expr_no_constructor)) { $1 };
+
+simple_expr_template_constructor:
```
jordwalke: What is meant by "template"?

let-def: We need to use the same definitions in different contexts.
The part that are shared are defined as parametric/higher-order non-terminals. The parameter is used in place of recursive occurrences of the non-terminal. That way the fix point can be tied at the use site.

Thus it is a template of the actual grammar, with the recursion not fixed; the terminology has the same meaning as `templace` in C++.
The parameter plays the same role as a self-type in an OOP language. Actually, OOP inheritance seems appropriate in a grammar specification language.

let-def: A small illustration with an extensible arithmetic grammar.

You start with:
```
expr: INT | LPAREN expr RPAREN | expr PLUS expr.
```
You want to add a `TIMES` operator:
```
expr2: expr TIMES expr
```
Fine, but the recursion is done on `expr`, so `TIMES` can only be used at the top-level.

Second attempt, factor the recursion out of `expr`:
```
open_expr<Y>: INT | LPAREN Y RPAREN | Y PLUS Y.
```
Now the original expr can simply be recovered with `expr: open_expr<expr>` (a fix point of open_expr).

The extension:
```
expr2: expr2 TIMES expr2 | open_expr<expr2>
```

The scheme composes (up to LR), so you can keep the extension open:
```
open_expr2<Y>: Y TIMES Y | open_expr<Y>
```

With an OO specification, Y would be `'self`.

let-def:
Updated with printing of `switch` and `try`.

chenglou:
That sure is a weird test failure on 4.02.3... `switch None` seems to pass fine

chenglou:
This looks good, thanks a lot =D

Merging! I'll update the last test afterward.

Issue #1451, opened 2017-10-04
[feature wish]short-cut for Js.nullable


bobzhang:
following #1450, it  would be nice that we have 

```re
type u = { x : ? int };
```
instead of 
```re
type u = Js.t {x : Js.nullable int}
```

This adds up when dealing with FFI (cc @rickyvetter )

Edit: or alternative
```re
type u = { ?x : int }; 
```

This difference is that such sugar only works for object type

PR #1477, opened 2017-10-09
Js.nullable syntax

IwanKaramazow:
https://github.com/facebook/reason/issues/1451

Applies to object types only.
```
/* before */
type u = {. "x" : Js.nullable(int) };

/* after */
type u  = {. "x"?: int };
```

This will be pretty familiar for people coming from Flow.

Locations are as good as they can get.


chenglou:
Nice! Not yet clear whether we should use `?` for nullable or option. Also, this is a progressive enhancement, so let's wait until next release.

bobzhang:
how about making it more flow like?
```
{ "x"? : int }
```
Then we don't need argue about whether it should be option or nullable, since it only exists in object types

IwanKaramazow:
Moved the question mark to before the colon.
```
type u  = {. "x"?: int };
```

anmonteiro:
What does this actually mean? 

If I'm not mistaken, in Flow `{x?: number}` means that the key may be missing from the object, which would IMO be confusing in OCaml / Reason. 

I'm advocating for the `{. "x": int?}` or ``{. "x": ?int}`` syntax because by defining the type it means the key will always be there, even though its value might be `None`. This actually also works in Flow: https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVAXAngBwKZgBiccYAvGAN5QkBcYA-AHYCuAtgEZ4BOAvgNzoAbgENuYGnHrFSFanTCsYMAUA

IwanKaramazow:
I think the problem is that at one point in the (far :)) future, `?` could be used for `option`.
The purpose of this PR is to explore syntax to help ffi bindings which involve `type u = {. "x" : Js.nullable(int) };`. Personally I can never reminder where I should place the question mark with Flow . I tend to agree that if you follow flow conventions that `key?: type` indicates that the `key`  might be missing. On the other hand, if you bind to this kind of ffi you'll always write `{. "key" : Js.nullable(type) };` 😇 (Unless there's something I don't know about? :D )

bobzhang:
Note Js.nullable could be null/undefined — in that case such key could be missing?

chenglou:
So apparently the diff disappeared...

IwanKaramazow:
Accidentally wiped out everything, new PR opened in https://github.com/facebook/reason/pull/1515

PR #1478, opened 2017-10-10
Improvements to #1474

chenglou:


chenglou:
cc @GuillaumeSalles

GuillaumeSalles:
Make sense to map signatureItem directly. Thanks for the review!

PR #1481, opened 2017-10-10
special case printing of foo(bar)##value

let-def:
This only fixes the printing, parsing already works (as shown by the tests). It seems simple enough to just special case this printing that I didn't look for a more involved solution.


chenglou:
Nice!

chenglou:
@let-def we now have #1507 which is also parsed wrong. Any solution?

PR #1482, opened 2017-10-10
Fix #1472

### LINKED ISSUE #1472: Functor parse & print with arrow
Making functors print as `module Foo = (Bar) => Baz` rather than the current `module Foo(Bar) = Baz`. Related: #1470

let-def:
- add supports for `(Functor_arg : Arg_type) : Functor_body_type => Functor_body` form.
- print functor bindings as `module X = (Args) => Body`

Supported functor forms:
`(A1 : T1) => Body`
`(A1 : T1) : Constraint => Body`
`() => Body`
`() : Constraint => Body`
`(A1 : T1, A2 : T2) => Body`
`(A1 : T1, ()) : Constraint => Body`
...

### DIFF
--- a/formatTest/unit_tests/expected_output/modules.re
+++ b/formatTest/unit_tests/expected_output/modules.re
@@ -168,7 +168,7 @@ module InliningSig: {let x: int; let y: int;} = {
   let y = 20;
 };
 
-module MyFunctor (M: HasTT) = {
+module MyFunctor = (M: HasTT) => {
   type reexportedTT = M.tt;
   /* Inline comment inside module. */
   /** Following special comment inside module. */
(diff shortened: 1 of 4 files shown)


chenglou:
How does `() => Body` work? Doesn't it conflict with function returning a variant?

let-def:
`()` is necessary to preserve soundness for generative functors.

let-def:
See http://caml.inria.fr/pub/docs/manual-ocaml/extn.html#sec253

chenglou:
Just saw on https://blog.janestreet.com/ocaml-4-02-everything-else/ too! Thanks for the info. I'm slightly more confused on how identity works now =P

Thanks. Merging!

Issue #1447, opened 2017-10-04
Change labeled argument from `:` back to `~`


chenglou:
Current master syntax:

```reason
let foo = (:bar) => bar;
```

Goal:

```reason
let foo = (~bar) => bar;
```

Let's reserve this task for last, since this is the bikesheddiest one.

I'm opening up bikeshedding for this **single item**, because this is has been the most divisive one.

👍  **if you're for** the change (you like `~`)

❤️ **if you're against** the change (you like `:`)

Nothing else please! Thanks. **No comment** needed! In case you feel bad for disagreeing with the change: don't worry about it. It's a community decision. Make your upvote/heart seen.

PR #1483, opened 2017-10-10
use ~ for named args

IwanKaramazow:
Fixes https://github.com/facebook/reason/issues/1447


chenglou:
RIP colon. You'll be missed

PR #1485, opened 2017-10-10
Remove ppx_tools again

chenglou:
It was introduced by 26f12520e58461f64b92fcccb161bc64f2c877d5 to patch 1.13.7's release into 1.13.8. We don't need it in master anymore


hcarty:
Sorry for adding that back in!

chenglou:
Np at all =)

PR #1491, opened 2017-10-11
Remove extra parens printed in `type a = Foo((unit => unit))`

### LINKED ISSUE #1490: Extra parentheses printed in `type a = Foo((unit => unit))`
Prints as

```reason
type a =
  | Foo((unit => unit));
```

Same for

```reason
type a =
  [> | `foo((unit => unit))]
```

chenglou:
Fixes #1490 

Summoning @let-def and @IwanKaramazow to review this. I have no idea what I'm doing. Heck, what's the difference between `non_arrowed_non_simple_core_type` and `core_type`?.

This seems to pass the tests though...

### DIFF
--- a/src/reason_pprint_ast.ml
+++ b/src/reason_pprint_ast.ml
@@ -2790,7 +2790,7 @@ class printer  ()= object(self:'self)
       let lst = if print_bar then [atom "|"; sourceMappedName] else [sourceMappedName] in
       makeList ~postSpace:true lst in
     let ampersand_helper i arg =
-      let ct = self#non_arrowed_non_simple_core_type arg in
+      let ct = self#core_type arg in
       let ct = match arg.ptyp_desc with
         | Ptyp_tuple _ -> ct
         | _ -> makeTup [ct]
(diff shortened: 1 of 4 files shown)


jordwalke:
FYI: "non arrowed"  means "a type where all arrows are inside at least one level of parens".

`z => z`: *not* a "non-arrowed" type.
`(a, b)`: a "non-arrowed" type.
`(z=>z)`: a "non-arrowed" type because the arrows are guarded by parens.

src/reason_pprint_ast.ml
```diff
@@ -2804,7 +2804,7 @@ class printer  ()= object(self:'self)
       | Pcstr_record r -> [self#record_declaration r]
       | Pcstr_tuple [] -> []
       | Pcstr_tuple l when polymorphic -> List.mapi ampersand_helper l
-      | Pcstr_tuple l -> [makeTup (List.map self#non_arrowed_non_simple_core_type l)]
+      | Pcstr_tuple l -> [makeTup (List.map self#core_type l)]
```
jordwalke: Here's why this works. With the new syntax, all the `args`, are already inside of a safely guarded place like `Constructor(here, andHere)`. Compare that to the previous syntax `Constructor here andHere`. In the previous syntax, we needed to require that we print "non-arrowed" types for `here`, and `andHere` to avoid something like `Constructor a=>b c=>d`. In the new syntax, we don't care if `here` and `andHere` have unguarded arrow types like `a=>b` because they're safely separated by commas.

jordwalke:
Awesome, @chenglou. We're getting closer!

Issue #1453, opened 2017-10-05
Fix outcome printer


chenglou:
The outcome printer hasn't been updated to reflect the syntax change. This is the last last issue we'd tackle before shipping it.

jaredly:
I'm looking into this -- @let-def have you done anything here yet?

PR #1499, opened 2017-10-12
Add a test setup for oprint!

jaredly:
This way we can verify that oprint's output (for signatures) is valid
reason.

Test Plan:
`node ./formatTest/testOprint.js`

working toward #1453. This doesn't change the printer at all, just makes a way for us to verify when it works.
Note that I haven't hooked this up to our `test` command in the Makefile, as it would be failing.

Current output
```
❯❯❯ node ./formatTest/testOprint.js
Output printed for the signature of "basic.re" not parseable:

## Refmt's error:

File "", line 2, characters 0-3:
Error: 262: <UNKNOWN SYNTAX ERROR>

## Outcome printed:

let run : unit => string
let endOfRangeMustBeSimple : 'a => 'b => int
let theSame : 'a => 'b => 'c => int
type zed = { bar : ref int, }
let foo : ref (ref zed)


## Refmt's interface printed:

let run: unit => string;

let endOfRangeMustBeSimple: ('a, 'b) => int;

let theSame: ('a, 'b, 'c) => int;

type zed = {bar: ref(int)};

let foo: ref(ref(zed));

Done! 1 failures
```


chenglou:
cc @let-def. Might help with the iterations in the future =P

PR #1502, opened 2017-10-12
don't apply sugar to Js.t({.}) and Js.t({..})

IwanKaramazow:
There's a bug in the printer which formats `Js.t({.})` to `{.}`. (Idem for `Js.t({..})`).

This PR gives the following printing.
```reason
type x = Js.t({.});
type y = Js.t({..});
type z = {. "foo": bar}; /* idem for {.. "etc": baz} */
```


chenglou:
Well, I guess it's not too bad not having the sugar `{.. ""}` since we also don't have it for empty js.t obj values...

PR #1515, opened 2017-10-14
Js.nullable syntax: `{. "x"?: int };`

IwanKaramazow:
https://github.com/facebook/reason/issues/1451

Applies to object types only.
```
/* before */
type u = {. "x" : Js.nullable(int) };

/* after */
type u  = {. "x"?: int };
```

This will be pretty familiar for people coming from Flow.

Locations are as good as they can get.

Reopening this one because I magically wiped out everything in https://github.com/facebook/reason/pull/1477


jaredly:
oooh this is quite fancy.

IwanKaramazow:
Can be revisited in the future, not convinced that this would do us any good atm.

Issue #1510, opened 2017-10-13
Upgraded syntax throws error: Tuples must have at least 2 components


jlongster:
I upgraded from beta-v-1.13.3 to master today, and ran the refmt process to upgrade all of my syntax. Previously I was doing some thing like this (I hand-converted this to the old syntax for this issue, hopefully I didn't make any mistakes):

```
Yojson.pretty_to_string(`Assoc [("client_id", `String "hi")]);
```

The new refmt converted it to this:

```
Yojson.pretty_to_string(`Assoc([("client_id", `String("hi"))]));
```

But compiling it throws an error: `Error: broken invariant in parsetree: Tuples must have at least 2 components`.

The full command is: `ocamlfind ocamlc -thread -package core,cohttp.lwt,redis.lwt,yojson,uuidm,kaputt -g -bin-annot -pp "refmt --print binary" -o _build/self/run2 -I _build/self -c -impl src/run2.re`

`run2.re` has only that expression inside of it. refmt is definitely the updated one which I pinned from master:

```
% where refmt
/Users/james/.opam/4.04.2/bin/refmt
```

chenglou:
Small repro:

```reason
`Assoc(1);
```

This is parsed as

```
[
  structure_item (foo.re[1,0+0]..[1,0+10])
    Pstr_eval
    expression (foo.re[1,0+0]..[1,0+9])
      Pexp_variant "Assoc"
      Some
        expression (foo.re[1,0+6]..[1,0+9])
          Pexp_tuple
          [
            expression (foo.re[1,0+7]..[1,0+8])
              Pexp_constant Const_int 1
          ]
]
```

Previously we've parsed

```
`Assoc 1
```

As

```
[
  structure_item (foo.re[1,0+0]..[1,0+9])
    Pstr_eval
    expression (foo.re[1,0+0]..[1,0+8])
      Pexp_variant "Assoc"
      Some
        expression (foo.re[1,0+7]..[1,0+8])
          Pexp_constant Const_int 1
]
```

In contrast, the current `Assoc(1)` and previous `Assoc 1` are both parsed as:

```
[
  structure_item (foo.re[1,0+0]..[1,0+9])
    Pstr_eval
    expression (foo.re[1,0+0]..[1,0+8])
      attribute "explicit_arity"
        []
      Pexp_construct "Assoc" (foo.re[1,0+0]..[1,0+5])
      Some
        expression (foo.re[1,0+5]..[1,0+8])
          Pexp_tuple
          [
[...]

PR #1516, opened 2017-10-15
MyConstructor({x: reallyLong, y: anotherReallyLongThing}) -> hug ({ and })

IwanKaramazow:
Fixes https://github.com/facebook/reason/issues/1510

Parser:

```reason
`Assoc(1)
```
is now parsed as
```
[
  structure_item (x.re[1,0+0]..[1,0+10])
    Pstr_eval
    expression (x.re[1,0+0]..[1,0+9])
      Pexp_variant "Assoc"
      Some
        expression (x.re[1,0+6]..[1,0+9])
          Pexp_constant PConst_int (1,None)
]
```
Unnecessary tuple is gone & matches the previous parser's behaviour.

Printer:

```reason
/* before */
Delete(
  { 
    uuid: imagineSomethingReallyLong 
  }
);
`Delete(
  { 
    uuid: imagineSomethingReallyLong 
  }
);

/* after*/
Delete({
  uuid: imagineSomethingReallyLong
});

`Delete({
  uuid: imagineSomethingReallyLong
});
```
This happens for tuples, lists, spread lists, arrays, bs obj sugar and objects.
Both variants and poly-variants are covered in this PR.

There was another issue with regular function application with a single object which formatted pretty ugly, has also been fixed.
```reason
/* before */
let res =
  acceptsOpenAnonObjAsArg(
    {pub x = 0; pub y = 10}
  );

/* after */
let res =
  acceptsOpenAnonObjAsArg({
    pub x = 0;
    pub y = 10
  });

```

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/mlVariants.re
+++ b/formatTest/typeCheckedTests/expected_output/mlVariants.re
@@ -4,9 +4,9 @@ type polyVariantsInMl = [
   | `StillAnIntTuple(int, int)
 ];
 
-let intTuple = `IntTuple(1, 2);
+let intTuple = `IntTuple((1, 2));
 
-let stillAnIntTuple = `StillAnIntTuple(4, 5);
+let stillAnIntTuple = `StillAnIntTuple((4, 5));
 
 let sumThem =
   fun
--- a/formatTest/typeCheckedTests/expected_output/oo.re
+++ b/formatTest/typeCheckedTests/expected_output/oo.re
@@ -162,19 +162,23 @@ let acceptsClosedAnonObjAsArg =
   o#x + o#y;
 
 let res =
-  acceptsOpenAnonObjAsArg(
-    {pub x = 0; pub y = 10}
-  );
+  acceptsOpenAnonObjAsArg({
+    pub x = 0;
+    pub y = 10
+  });
 
 let res =
-  acceptsOpenAnonObjAsArg(
-    {pub x = 0; pub y = 10; pub z = 10}
-  );
+  acceptsOpenAnonObjAsArg({
+    pub x = 0;
+    pub y = 10;
+    pub z = 10
+  });
 
 let res =
-  acceptsClosedAnonObjAsArg(
-    {pub x = 0; pub y = 10}
-  );
+  acceptsClosedAnonObjAsArg({
+    pub x = 0;
+    pub y = 10
+  });
 
 /* TODO: Unify class constructor return values with function return values */
 class myClassWithAnnotatedReturnType
--- a/src/reason_pprint_ast.ml
+++ b/src/reason_pprint_ast.ml
@@ -2392,6 +2392,28 @@ let is_direct_pattern x = x.ppat_attributes = [] && match x.ppat_desc with
   | Ppat_construct ( {txt= Lident"()"}, None)
   | Ppat_construct ( {txt= Lident"::"}, Some _) -> true
   | _ -> false
+
+(* Some cases require special formatting when there's a function application
+ * with a single argument containing some kind of structure with braces/parens/brackets.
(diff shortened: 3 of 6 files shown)


formatTest/typeCheckedTests/expected_output/mlVariants.re
```diff
@@ -4,9 +4,9 @@ type polyVariantsInMl = [
   | `StillAnIntTuple(int, int)
 ];
 
-let intTuple = `IntTuple(1, 2);
+let intTuple = `IntTuple((1, 2));
```
chenglou: Can this double parens be avoided?

IwanKaramazow: Yes, and what about `OrdinaryVariant((1, 2))` & `myFunctionApplication((1, 2))` ?

chenglou: Oh wait, that can't be avoided righr

chenglou: How did this pass previously?

chenglou: Nvm, see my comment

chenglou:
Discussed offline. We'll proceed with parsing both:

```reason
`Foo(1, 2);
`Foo((1, 2));
```

but printing both to:

```reason
`Foo((1, 2));
```

Poly variant really just accepts a single argument. `Foo(1, 2)` can currently be viewed as sugar for poly variant carrying a tuple. As for why it doesn't take multiple args:

> There's no real great reason for it, but it has to do with boxing. The fact that regular variants accept multiple args is the distinction between additional boxing, and not boxing of a variant, and IIRC, poly variants already required a boxing, so it didn't make sense to allow multiple args which was a means of avoiding boxing

So why not format both to:

```reason
`Foo(1, 2);
```

Because we'd like to future-proof for a potential multi-arity poly variant, and to make that distinction between tuple and n-arity clear from this point on. Last justification:

> Another reason to make it clear:
> These are equivalent:

> `X (1, 2);
> and
> let t = (1, 2); `X(t);

> But these are not equivalent:
> X (1, 2);
> and
> let t = (1, 2); X(t);

Hope this discussion helps for posteriority. Merging! Thanks.

jordwalke:
"Discussed offline. We'll proceed with parsing both:

`Foo(1, 2);
`Foo((1, 2));
"

I proposed that this is a graceful way to teach people that "variants always accept one argument, that may be a tuple". If the current PR only parses \`X((1,2)) and not \`X(1,2) then that's fine too, we can add the graceful parsing/instructive-printing at a later time.

formatTest/typeCheckedTests/expected_output/oo.re
```diff
@@ -162,19 +162,23 @@ let acceptsClosedAnonObjAsArg =
   o#x + o#y;
 
 let res =
-  acceptsOpenAnonObjAsArg(
-    {pub x = 0; pub y = 10}
-  );
+  acceptsOpenAnonObjAsArg({
```
jordwalke: ❤️  thank you thank you thank you.

src/reason_pprint_ast.ml
```diff
@@ -2392,6 +2392,28 @@ let is_direct_pattern x = x.ppat_attributes = [] && match x.ppat_desc with
...
+ * Example: `foo({a: 1, b: 2})` needs to be formatted as
+ *  foo({
+ *    a: 1,
+ *    b: 2
+ *  })
+ *  when the line length dictates breaking. Notice how `({` and `})` 'hug'.
+ *  Also applies to (poly)variants because they can be seen as a form of "function application".
+ *  This function says if a list of expressions fulfills the need to be formatted like
+ *  the example above. *)
+let isSingleArgParenApplication = function
```
jordwalke: Great :D

Issue #1320, opened 2017-06-26
Support record punning + renaming


chenglou:
```reason
fun foo {newSelf: {state: state as prevState}} => 1;
```

It'd be nice to have
```reason
fun foo {newSelf: {state as prevState}} => 1;
```

If possible

keleshev:
Does the later mean the same as the following?

```
fun foo {newSelf: {state: prevState}} => 1;
```

IwanKaramazow:
They have a different meaning:
```
/* pun2.re */
let x = (foo, {newSelf: {state as prevState}}) => 1;
let x = (foo, {newSelf: {state: prevState }}) => 1;

[
  structure_item (pun2.re[1,0+0]..[1,0+52])
    Pstr_value Nonrec
    [
      <def>
        pattern (pun2.re[1,0+4]..[1,0+5])
          Ppat_var "x" (pun2.re[1,0+4]..[1,0+5])
        expression (pun2.re[1,0+8]..[1,0+51])
          Pexp_fun
          Nolabel
          None
          pattern (pun2.re[1,0+9]..[1,0+12])
            Ppat_var "foo" (pun2.re[1,0+9]..[1,0+12])
          expression (pun2.re[1,0+14]..[1,0+51])
            Pexp_fun
            Nolabel
            None
            pattern (pun2.re[1,0+14]..[1,0+45])
              Ppat_record Closed
              [
                "newSelf" (pun2.re[1,0+15]..[1,0+22])
                  pattern (pun2.re[1,0+24]..[1,0+44])
                    Ppat_record Closed
                    [
                      "state" (pun2.re[1,0+25]..[1,0+30])
                        pattern ([0,0+-1]..[0,0+-1])
                          Ppat_alias "prevState" (pun2.re[1,0+34]..[1,0+43])
                          pattern (pun2.re[1,0+25]..[1,0+30])
                            Ppat_var "state" (pun2.re[1,0+25]..[1,0+30])
                    ]
              ]
            expression (pun2.re[1,0+50]..[1,0+51])
              Pexp_constant PConst_int (1,None)
    ]
  structure_item (pun2.re[2,53+0]..[2,53+51])
    Pstr_value Nonrec
    [
      <def>
        pattern (pun2.re[2,53+4]..[2,53+5])
          Ppat_var "x" (pun2.re[2,53+4]..[2,53+5])
        expression (pun2.re[2,53+8]..[2,53+50])
          Pexp_fun
          Nolabel
          None
[...]

PR #1517, opened 2017-10-16
pun record destructuring with renaming

IwanKaramazow:
Fixes https://github.com/facebook/reason/issues/1320

```reason
/* before */
let x = ({ReasonReact.state: state as prevState}) => 1;
let x = ({state: state as prevState}) => 1;

/* after */
let x = ({ReasonReact.state as prevState}) => 1;
let x = ({state as prevState}) => 1;
```


chenglou:
Thanks!

jordwalke:
Nice improvement.

PR #1520, opened 2017-10-17
Fix outcome printer for class type and variant values

chenglou:
For class type: see comment.
For variant values: they weren't updated to the new syntax. We'll put test for them in #1548

### DIFF
--- a/formatTest/oprintTests/basic.re
+++ b/formatTest/oprintTests/basic.re
@@ -17,6 +17,11 @@ type t2 = (a, b) => c;
 type t3 = ((a, b)) => c;
 type t4 = (~x: int, ~y: string) => c;
 type t5 = (~x: a=?) => b;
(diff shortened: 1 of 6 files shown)


jaredly:
Hmm so the parens will become unnecessary once you change the reason parser? :P sounds like they're pretty necessary atm

chenglou:
This currently works:

```reason
let foo: a => b;
```

As well as:

```reason
let foo: a => b => c;
```

They format to what you think they do. But not for class:

```reason
class foo: a => b;
```

This is an error. It absolutely needs to be `class foo: (a) => b` If we comment out the part of the oprint tests for that class type signature, everything else works.

chenglou:
Updated the PR to fix class type outcome printing altogether (nothing to comment out anymore), plus variant value printing

Issue #837, opened 2016-11-09
attribute disappears on external


chenglou:
Repro:

```
external debounce : int => ((unit => unit)[@bs.meth]) = "" [@@bs.module];
```

Turns into:

```
external debounce : int => unit => unit = "" [@@bs.module];
```

cc @bsansouci

jordwalke:
I thought this was fixed. Did we not merge a PR?

chenglou:
There are more than one of these issues.

bobzhang:
Attributes are very easy to be dropped by ppx (not yours maybe others). I would suggest using extension points (but you can design a better syntax).

PR #1523, opened 2017-10-17
Fix printings of attributes on Ptyp_arrow

IwanKaramazow:
Fixes https://github.com/facebook/reason/issues/837

`ptyp_attributes` on `core_type`'s where ignored at some point in the code when formatting `Ptyp_arrow`'s. It resulted in attributes disappearing. The solution is to format `Ptyp_arrow`'s with attributes as a separate core_type instead of recursing through it's tree.

```reason
/* input */
external debounce : (int, [@bs.meth] unit) => unit = "";
external debounce : int => ([@bs.meth] (unit => unit)) = "";
external debounce : (int, ([@bs.meth] (unit => unit))) => ([@bs.meth] (unit => unit)) = "";

/* after refmt */
external debounce : (int, [@bs.meth] unit) => unit = "";
external debounce : int => [@bs.meth] (unit => unit) = "";
external debounce : (int, [@bs.meth] (unit => unit)) => [@bs.meth] (unit => unit) = "";
```

### DIFF
--- a/src/reason_pprint_ast.ml
+++ b/src/reason_pprint_ast.ml
@@ -2561,8 +2561,9 @@ class printer  ()= object(self:'self)
       match (x.ptyp_desc) with
         | (Ptyp_arrow (l, ct1, ct2)) ->
           let rec allArrowSegments acc = function
-            | { ptyp_desc = Ptyp_arrow (l, ct1, ct2); _ } ->
+            | { ptyp_desc = Ptyp_arrow (l, ct1, ct2); ptyp_attributes } when List.length ptyp_attributes == 0 ->
               allArrowSegments ((l,ct1) :: acc) ct2
+
             | rhs ->
               let rhs = self#core_type2 rhs in
               let is_tuple typ = match typ.ptyp_desc with
@@ -2570,7 +2571,7 @@ class printer  ()= object(self:'self)
                 | _ -> false
               in
               match acc with
-              | [Nolabel, lhs] when not (is_tuple lhs) ->
+              | [(Nolabel, lhs)] when not (is_tuple lhs) ->
                 (self#non_arrowed_simple_core_type lhs, rhs)
               | acc ->
                 let params = List.rev_map self#type_with_label acc in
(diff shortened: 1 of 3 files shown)


chenglou:
Humm this doesn't seem right:

```sh
❯ echo 'external debounce : int => ([@bs.meth] (unit => unit)) = "";' | ./refmt_impl.native --parse re --print ml
external debounce : int -> ((unit -> unit)[@bs.meth ]) = ""
```

```sh
❯ echo 'external debounce : int => ([@bs.meth] (unit => unit)) = "";' | ./refmt_impl.native --parse re --print re
external debounce : (int, [@bs.meth] unit) => unit = "";
```

IwanKaramazow:
Sorry, should have tested more thorough.
Pushed the fix, which seems to be a one-liner...

```bash
echo 'external debounce : int => ([@bs.meth] (unit => unit)) = "";' | ./refmt_impl.native --parse re --print ml
external debounce : int -> ((unit -> unit)[@bs.meth ]) = ""
```

```bash
echo 'external debounce : int => ([@bs.meth] (unit => unit)) = "";' | ./refmt_impl.native --parse re --print re
external debounce : int => [@bs.meth] (unit => unit) = "";
```

jordwalke:
Awesome, thank you @IwanKaramazow !

src/reason_pprint_ast.ml
```diff
@@ -2561,16 +2561,17 @@ class printer  ()= object(self:'self)
       match (x.ptyp_desc) with
         | (Ptyp_arrow (l, ct1, ct2)) ->
           let rec allArrowSegments acc = function
-            | { ptyp_desc = Ptyp_arrow (l, ct1, ct2); _ } ->
+            | { ptyp_desc = Ptyp_arrow (l, ct1, ct2); ptyp_attributes } when List.length ptyp_attributes == 0 ->
```
chenglou: Would this be better? `ptyp_attributes = []`

IwanKaramazow: updated

IwanKaramazow:
I have no idea why travis hanged...
Circleci seems to be ok though.

chenglou:
Looks great. Merged! Thanks, as always

PR #1524, opened 2017-10-19
Make rtop integration test syntax-neutral

cristianoc:


chenglou:
Wait, we don't want it to be syntax neutral though? It's also to test whether rtop has picked up the correct new syntax

cristianoc:
This was only to remove the opam pin friction. But since the new syntax release is close, not worth it.

PR #1527, opened 2017-10-20
Missing parens when pattern matching on Ppat_construct containing a list

### LINKED ISSUE #1525: "| Some [x]" does not include parens with refmt in new syntax
Example, using e065e253:
```ocaml
let f x =
  match x with
  | None -> ()
  | Some [] -> ()
  | Some [_] -> ()
  | Some _ -> ()
```
refmts to
```reason
let f = (x) =>
  switch x {
  | None => ()
  | Some([]) => ()
  | Some [_] => () /* Not good */
  | Some(_) => ()
  };
```
which is not valid syntax in the single-element list case.

IwanKaramazow:
Fixes #1525 

```
/* before */
  switch x {
  | Some([]) => ()
  | Some [_] => () /* Not good, parens missing */
  | Some [x] => () /* Not good, parens missing */
  | Some [x, y, z] => () /* Not good, parens missing */
  | Some [x, ...xs] => () /* Not good, parens missing */
  };

/* after */
  switch x {
  | Some([]) => ()
  | Some([_]) => ()
  | Some([x]) => ()
  | Some([x, y, z]) => ()
  | Some([x, ...xs]) => ()
  };
```

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/patternMatching.re
+++ b/formatTest/typeCheckedTests/expected_output/patternMatching.re
@@ -182,3 +182,12 @@ switch (Some(1)) {
 | Some(1) => 1
 | None => 2
 };
+
+switch None {
+| Some([]) => ()
+| Some([_]) => ()
+| Some([x]) => ()
+| Some([x, ...xs]) => ()
+| Some([x, y, z]) => ()
+| _ => ()
+};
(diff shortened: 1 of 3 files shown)


chenglou:
So we support array, record, unit like this:

```reason
switch x {
| Some {a} => ()
};
```

But not list?

That's too bad. I think we should be consistent here and support it. But this can be done in another release. Merging! Thanks

jordwalke:
Thank you!

hcarty:
Thanks!

PR #1528, opened 2017-10-20
Add support for simple pattern direct argument with array, list & record.

IwanKaramazow:
https://github.com/facebook/reason/pull/1527#issuecomment-338124733

This PR streamlines  all "sugar" around `simple_pattern_direct_argument`:
```reason
/* you don't need to surround the constructor's arg with parens */
switch v {
| Foo () => ...
| Foo [a] => ...
| Foo [||] => ...
| Foo {a} => ...
};
```


chenglou:
Good stuff

PR #1534, opened 2017-10-21
Publish refmt.js before `npm publish`

chenglou:
We have an official JS api now!


package.json
```diff
@@ -3,23 +3,23 @@
...
 	"homepage": "https://github.com/facebook/reason",
-	"dependencies": {
-	},
+	"dependencies": {},
 	"devDependencies": {
-		"bs-platform": "^1.9.1"
+		"bs-platform": "^1.9.3"
 	},
 	"scripts": {
+		"prepublish": "cd ./bspacks && ./downloadSomeDependencies.sh && ./reason_bspack.sh"
```
jchavarri: Is this required to run when consumers run `npm install`? If not, would it be better to use [`prepublishOnly`](https://docs.npmjs.com/misc/scripts#prepublish-and-prepare)?

chenglou: Didn't know about that, thanks.

PR #1531, opened 2017-10-21
implement Pexp_letexception

IwanKaramazow:
https://github.com/facebook/reason/issues/1424

Adds support for `Pexp_letexception`.

Ocaml:
```ocaml
let () = let exception E in raise E

```
resulting in parse tree:
```

(* parse tree *)
[
  structure_item (exc.ml[1,0+0]..[1,0+35])
    Pstr_value Nonrec
    [
      <def>
        pattern (exc.ml[1,0+4]..[1,0+6])
          Ppat_construct "()" (exc.ml[1,0+4]..[1,0+6])
          None
        expression (exc.ml[1,0+9]..[1,0+35])
          Pexp_letexception
          extension_constructor (exc.ml[1,0+23]..[1,0+24])
            pext_name = "E"
            pext_kind =
              Pext_decl
                []
                None
          expression (exc.ml[1,0+28]..[1,0+35])
            Pexp_apply
            expression (exc.ml[1,0+28]..[1,0+33])
              Pexp_ident "raise" (exc.ml[1,0+28]..[1,0+33])
            [
              <arg>
              Nolabel
                expression (exc.ml[1,0+34]..[1,0+35])
                  Pexp_construct "E" (exc.ml[1,0+34]..[1,0+35])
                  None
            ]
    ]
]
```

Becomes in Reason:
```reason
let () = {
  exception E;
  raise(E);
};
```
resulting in parse tree:
```
[
  structure_item (local.re[1,0+0]..[4,39+2])
    Pstr_value Nonrec
    [
      <def>
        pattern (local.re[1,0+4]..[1,0+6])
          Ppat_construct "()" (local.re[1,0+4]..[1,0+6])
          None
        expression (local.re[1,0+9]..[4,39+1])
          Pexp_letexception
          extension_constructor (local.re[2,11+12]..[2,11+13])
            pext_name = "E"
            pext_kind =
              Pext_decl
                []
                None
          expression (local.re[3,26+2]..[3,26+11])
            Pexp_apply
            expression (local.re[3,26+2]..[3,26+7])
              Pexp_ident "raise" (local.re[3,26+2]..[3,26+7])
            [
              <arg>
              Nolabel
                expression (local.re[3,26+8]..[3,26+9])
                  Pexp_construct "E" (local.re[3,26+8]..[3,26+9])
                  None
            ]
    ]
]
```

PR #1537, opened 2017-10-21
Re-enable Pexp_letception tests disabled in #1531

### LINKED ISSUE #1535: Re-enable tests from #1531
Those were commented out because 4.02.3 and 4.03 don't support the feature

zploskey:
Re-enables tests that were disabled because this feature was not available on 4.03 and earlier.

Fixes #1535.

Introduces a convention for naming typecheck tests to only run them on the OCaml compiler version in their name or later.

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/attributes.4.04.0.re
+++ b/formatTest/typeCheckedTests/expected_output/attributes.4.04.0.re
@@ -1,0 +1,6 @@
+/* Pexp_letexception with attributes */
+let () = {
+  [@attribute]
+  exception E;
+  raise(E)
+};
--- a/formatTest/test.sh
+++ b/formatTest/test.sh
@@ -5,6 +5,7 @@ set -eu
 
 WARNING='\033[0;31m'
 SUCCESS='\033[0;32m'
+NOTICE='\033[0;33m'
 INFO=''
 DEBUG=''
 RESET='\033[0m'
@@ -62,6 +63,10 @@ function debug() {
     fi
 }
 
+function notice() {
+    printf "${NOTICE}$1${RESET}\n"
+}
+
 function success() {
     printf "${SUCCESS}$1${RESET}"
 }
@@ -72,6 +77,10 @@ function output() {
 
 function warning() {
     printf "${WARNING}$1${RESET}\n"
+}
+
+function version() {
+    echo "$@" | awk -F . '{ printf("%03d%03d%03d\n", $1, $2, $3); }';
 }
 
 function setup_test_dir() {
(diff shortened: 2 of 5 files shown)


zploskey:
Should these be disabled on 4.03 and earlier?

IwanKaramazow:
It's a feature that doesn't run on 4.02.3. Probably needs to be 4.04.2 and higher. Test script needs to be changed to run certain tests on specific versions. They were disabled because of this. Focus is on shipping the new syntax atm.

zploskey:
I added a commit which allows typeCheckedTests to skip the compilation step on certain compiler versions by adding a line to `formatTest/typeCheckedTests/ignore_in_version` indicating the filename and version number separates by a `:` character, i.e.
```
./attributes.re:4.02.3
```
If it is ignored on the compiler version which it is running on then it is marked as `  ☒ IGNORED` in yellow in the output and does not result in a failure.

chenglou:
Does that ignore the whole file?

zploskey:
Currently what it does is ignore only the compilation at the end of the typeCheckedTest. The formatting is still done, and it can fail if something went wrong there. Perhaps the naming of this type of ignore could be expanded upon a bit, but I figured we only wanted to ignore the parts that were failing due to differences in compiler support.

IwanKaramazow:
Not sure if ignoring the test is a good idea 😄 
Can you make it so that all test with the following name scheme work on specific compiler versions and above?
Example:
```
myFileTestName.<version>.re
attributes.4.04.2.re
```
So `attributes.4.04.2.re` only runs on 4.04.2 & 4.05 & 4.06 (e.g. a `>=`)

zploskey:
I was also thinking about whether the stuff that needs to be ignored in earlier compilers should be moved to its own file (so far just these two blocks in question). So I might go ahead and move these two into their own file.

> Can you make it so that all test with the following name scheme work on specific compiler versions and above?
I think that is probably an easier way of specifying which ones not to try to compile. Should they not be run at all or just skip the compilations step?

formatTest/typeCheckedTests/expected_output/attributes.4.04.0.re
```diff
@@ -0,0 +1,6 @@
+/* Pexp_letexception with attributes */
+let () = {
+  [@attribute]
+  exception E;
```
zploskey: Is this what the expected refmt output should look like? It's the current behavior, so I set it to pass with this result, but if the attribute is supposed to be on the same line then it should probably be fixed and the test modified to reflect that.

IwanKaramazow: This looks good for now! Even if we decide it shouldn't be printed this way, it falls outside of the scope of this PR.

IwanKaramazow:
```
 Should they not be run at all or just skip the compilation step?
```
We only care about the typed_checked tests because the unit tests only print the ast. It's the actual compilation with the ocaml compilier that verifies if the ast is correct. Ocaml-migrate-parsetree internally migrates the ast version to make Reason compiler-version agnostic. The ocaml compiler will complain though when you try to use 4.04 features on 4.02.3 for example.

Overall this PR is good for me.

formatTest/typeCheckedTests/expected_output/attributes.4.04.0.re
zploskey: Personally, I kind of like it this way. It jives with my expectations for annotations being on the line immediately preceding the thing they are annotating.

zploskey:
I know we don't build on it in CI, but I have verified that all the tests pass on OCaml 4.04.0, as that was the minimum version set for these tests.

chenglou:
We do run on 4.04 and it's passing?

chenglou:
Thanks, merged!

PR #1538, opened 2017-10-22
Enforce parens in simple pattern match printing

rickyvetter:
This allows for simple pattern matches to be parsed, but will always print with the parens for more consistency with type constructors outside of pattern matches (and the rest of the new syntax).

### DIFF
--- a/formatTest/unit_tests/expected_output/variants.re
+++ b/formatTest/unit_tests/expected_output/variants.re
@@ -400,32 +400,36 @@ let rec atLeastOneFlushableChildAndNoWipNoPending =
   | [] => false
   | [hd, ...tl] =>
     switch hd {
-    | OpaqueGraph {lifecycle: Reconciled(_, [])} =>
+    | OpaqueGraph({lifecycle: Reconciled(_, [])}) =>
       atLeastOneFlushableChildAndNoWipNoPending(
         tl,
         atPriority
       )
-    | OpaqueGraph {
-        lifecycle:
-          ReconciledFlushable(
-            priority,
-            _,
-            _,
-            _,
-            _,
-            _
-          )
-      }
-    | OpaqueGraph {
-        lifecycle:
-          NeverReconciledFlushable(
-            priority,
-            _,
-            _,
-            _,
-            _
-          )
-      }
+    | OpaqueGraph(
+        {
+          lifecycle:
+            ReconciledFlushable(
+              priority,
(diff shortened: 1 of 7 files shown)


src/reason_pprint_ast.ml
```diff
@@ -2388,10 +2388,7 @@ let is_unit_pattern x = match x.ppat_desc with
   | _ -> false
 
 let is_direct_pattern x = x.ppat_attributes = [] && match x.ppat_desc with
-  | Ppat_array _ | Ppat_record _
-  | Ppat_construct ( {txt= Lident"()"}, None)
-  | Ppat_construct ( {txt= Lident"[]"}, None)
-  | Ppat_construct ( {txt= Lident"::"}, Some _) -> true
+  | Ppat_construct ( {txt= Lident"()"}, None) -> true
```
rickyvetter: @IwanKaramazow - you mentioned removing the two cases added in a214071eba5e6ba72946c1655dc63a650f2bf7f1, but I found that Ppat_array and Ppat_record also cause the same inconsistencies. This will print all of the above with parens, but still allow typing without.

formatTest/unit_tests/expected_output/variants.re
```diff
@@ -400,32 +400,36 @@ let rec atLeastOneFlushableChildAndNoWipNoPending =
...
+          lifecycle:
+            NeverReconciledFlushable(
+              priority,
+              _,
+              _,
+              _,
+              _
+            )
+        }
+      )
```
rickyvetter: Printing here could probably be improved in a number of ways. Would be worth testing something like:

```
| OG({
    ...
  })
```
and
```
| OG({
    ...
  }) when ...
```

This would consolidate extra lines and extra indentation.

src/reason_pprint_ast.ml
IwanKaramazow: totally forgot about arrays and records 😅

formatTest/unit_tests/expected_output/variants.re
IwanKaramazow: I'll take a look today.

chenglou: I'll merge this first then

chenglou:
Too much bikeshedding

chenglou:
Opened a follow-up task: #1540

PR #1539, opened 2017-10-22
Keyword swap changes

### LINKED ISSUE #1361: Keyword swapping when using OCaml libraries leads to nonsensical results
Reason's approach of swapping keywords (`switch` <-> `match`, `pub` <-> `method`) can lead to some confusing results when using libraries written in OCaml.

A bad example is the zeromq bindings for OCaml which expose a `pub` value in reference to the pub socket in a pub/sub messaging duo.  `pub` is a keyword in Reason so this value gets translated to `method` which is confusing and undiscoverable to someone who isn't familiar with the details of this keyword swapping.

The relevant interface for the zeromq binding is here: https://github.com/issuu/ocaml-zmq/blob/1867f6a151356f9330e7f7720e7c732a5c1a1b36/src/ZMQ.mli#L35

```ocaml
(* Create a pub/sub socket pair *)
let pub = ZMQ.Socket.create ctx ZMQ.Socket.pub in
let sub = ZMQ.Socket.create ctx ZMQ.Socket.sub in
...
```
becomes
```reason
let method = ZMQ.Socket.create ctx ZMQ.Socket.method;
let sub = ZMQ.Socket.create ctx ZMQ.Socket.sub;
...
```
I've run into this keyword swapping issue a few times but this was by far the most confusing.

Would it be possible to do something less invasive like add a suffix `_` to names from OCaml which are keyword in Reason?

hcarty:
Fixes #1361 -- with the exception that it doesn't update any tests yet and has received minimal testing aside from some quick checks:
```
$ echo 'let pub = 5' | ./refmt_impl.native --parse=ml --print=re
let pub_ = 5;
$ echo 'let pub_ = 5' | ./refmt_impl.native --parse=re --print=ml
let pub = 5
```

### DIFF
--- a/src/syntax_util.ml
+++ b/src/syntax_util.ml
@@ -137,34 +137,88 @@ let syntax_error_extension_node loc message =
  in
  (str, payload)
 
-(** What does this do? Here's an example: Reason code uses `!` for logical not,
-  while ocaml uses `not`. So, for converting between reason and ocaml syntax,
-  ocaml `not` converts to `!`, reason `!` converts to `not`. This table is
-  used later on to do the actual swapping. 
+(** Check to see if the string `s` is made up of `keyword` and zero or more
+    trailing `_` characters. *)
+let potentially_conflicts_with ~keyword s =
+  let s_length = String.length s in
+  let keyword_length = String.length keyword in
+  (* It can't be a match if s is shorter than keyword *)
+  s_length >= keyword_length && (
+    try
+      (* Ensure s starts with keyword... *)
+      for i = 0 to keyword_length - 1 do
+        if keyword.[i] <> s.[i] then raise Exit;
+      done;
+      (* ...and contains nothing else except trailing _ characters *)
+      for i = keyword_length to s_length - 1 do
+        if s.[i] <> '_' then raise Exit;
+      done;
+      (* If we've made it this far there's a potential conflict *)
+      true
+    with
+    | Exit -> false
+  )
 
-  The drawback of this technique is that if you're using an identifier such as
-  `pub` in the ocaml file Foo.ml (say, let pub = 5), for a piece of Reason code
-  to interop with it, said code needs to write `Foo.method` (because of the
-  method <-> pub keyword swap). This is pretty dumb...
+(** Add/remove an appropriate suffix when mangling potential keywords *)
+let string_add_suffix x = x ^ "_"
+let string_drop_suffix x = String.sub x 0 (String.length x - 1)
+
+(** What do these *_swap functions do? Here's an example: Reason code uses `!`
+    for logical not, while ocaml uses `not`. So, for converting between reason
+    and ocaml syntax, ocaml `not` converts to `!`, reason `!` converts to
+    `not`.
+
+    In more complicated cases where a keyword exists in one syntax but not the
+    other these functions translate any potentially conflicting identifier into
+    the same identifier with a suffix attached, or remove the suffix when
+    converting back.
 *)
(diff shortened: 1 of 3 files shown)


hcarty:
Does this seems like a reasonable approach?  I don't have node setup locally so I can't do much in the way of thorough testing until I fix that as it looks like `make test` requires node now.

It would be REALLY nice to have this or an equivalent in before the big release, if that's at all possible.

src/syntax_util.ml
```diff
@@ -137,34 +137,68 @@ let syntax_error_extension_node loc message =
  in
  (str, payload)
 
+let string_prefix ~prefix s =
```
jordwalke: This could use a comment explaining what `string_prefix` does.

hcarty:
And there are bad bugs in the `string_prefix` function which I'll fix as soon as I'm in front of a computer again.

src/syntax_util.ml
jordwalke: And could it ever be the case that `s` is smaller in length than `prefix` (causing runtime error)? For example, an identifier such as `"matc"`. Do you want to do something like:

```
let matched = ref true in
   let n = String.length prefix in
   if String.length s >= n then
     for i = 0 to n - 1 do
       matched := !matched && prefix.[i] = s.[i]
     done;
     !matched
  else
    false
```

hcarty: Yeah - sorry about that! That's the bug I left a comment about. I realized I hadn't handled that case just after I turned off my computer.

This could also short-circuit by raising and immediately catching an exception when a character doesn't match.

I'll fix these issues and add some comments when I'm in front of a computer again, hopefully tomorrow.

hcarty: Thanks for taking the time to review!

hcarty:
Bug fixes, better function names, more comments, short-circuiting and rebased.  This should be in a much better state now.

As @jordwalke noted, the previous version would fail badly if the identifier was shorter than the prefix being tested against.  That should be fixed now.

A more subtle but still quite bad bug was that the `_` suffix meddling was applied for _any_ identifier starting with a keyword.  So both `pub` and `publish` would have been modified when translating between syntaxes even though `publish` does not have a conflict.  This has been fixed so that only keywords with zero or more `_` trailing characters are affected by the name substitution.

hcarty:
Updated tests which aren't run with node.

src/syntax_util.ml
```diff
@@ -137,34 +137,88 @@ let syntax_error_extension_node loc message =
...
+  | "!=" -> "<>"
+  | "!==" -> "!="
+  | x when (
+    potentially_conflicts_with ~keyword:"match" x
+    || potentially_conflicts_with ~keyword:"method" x
+    || potentially_conflicts_with ~keyword:"private" x) -> string_add_suffix x
+  | x when (
+    potentially_conflicts_with ~keyword:"switch_" x
+    || potentially_conflicts_with ~keyword:"pub_" x
+    || potentially_conflicts_with ~keyword:"pri_" x) -> string_drop_suffix x
```
chenglou: This is doing quite a few checks. Shouldn't it be calculated once then have the result pattern-matched on, in a nested match?

hcarty: There are a lot of checks there.  What do you mean by calculated once?

chenglou:
Tested and works. I'll put up some more comments and better tests. The current ones aren't indicative anymore.

Thanks!

jordwalke:
Thanks everyone!

PR #1541, opened 2017-10-23
Produce syntax error messages in some cases

let-def:
This PR extends the build process to analyse the parser at compile-time and produce informations that are used at runtime for generating error messages.

# Features

It detects use of keywords in place of identifiers, use of uppercase identifiers when lowercase is expected, and missing `;`.

```reason
let x = module;
```
```
File "test.re", line 1, characters 8-14:
Error: 1471: module is a reserved keyword, it cannot be used as an identifier. Try `module_' instead
```

```reason
let maker(~SomeModule) => SomeModule.make ();
```
```
File "test.re", line 1, characters 11-21:
Error: 504: variables and labels should be lowercased. Try `someModule'
```

```reason
let foo = (a) => 1

if (foo) { bar }
```
```
File "test.re", line 3, characters 0-2:
Error: 843: syntax error, consider adding a `;' before
```

# Implementation

This PR adds a build-time dependency on menhirSdk. This package is already part of Menhir, so it should not affect existing processes.
menhirSdk is used to process the grammar & parser:
- menhir is invoked with `--cmly` to produce a synthetic description of the grammar
- src/menhir_error_processor.ml extracts relevant informations from the grammar and produces src/reason_parser_explain_raw.ml
- src/reason_parser_explain.ml is linked to reason and is used to produce error messages by matching the parser state with the information stored in src/reason_parser_explain_raw.ml

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,15 +17,19 @@ precompile:
 	cp pkg/META.in pkg/META
 	ocamlbuild -use-ocamlfind -package topkg pkg/build.native
 
-build_without_utop: compile_error setup_convenient_bin_links precompile
+preprocess: precompile
+	./build.native build -r src/reason_parser.ml -r src/menhir_error_processor.native
+	./menhir_error_processor.native _build/src/reason_parser.cmly > src/reason_explain_error.ml
+
+build_without_utop: compile_error setup_convenient_bin_links preprocess
 	./build.native build --utop false
 	chmod +x $(shell pwd)/_build/src/*.sh
 
-build_with_outcome_test: compile_error setup_convenient_bin_links precompile
+build_with_outcome_test: compile_error setup_convenient_bin_links preprocess
 	./build.native build --utop true --outcome_test true
 	chmod +x $(shell pwd)/_build/src/*.sh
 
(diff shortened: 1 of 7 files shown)


chenglou:
Build is failing only because `reason_parser_explain_raw.ml` isn't checked in here, so that's fine. But when I run `rtop` after pinning reason again, I get:

```
File "/Users/chenglou/.opam/4.04.2/bin/rtop_init.ml", line 1:
Error: Reference to undefined global `Reason_parser_explain'
```

Not sure why. The rest seems to be passing!

Makefile
```diff
@@ -17,15 +17,19 @@ precompile:
 	cp pkg/META.in pkg/META
 	ocamlbuild -use-ocamlfind -package topkg pkg/build.native
 
-build_without_utop: compile_error setup_convenient_bin_links precompile
+preprocess: precompile
```
chenglou: I don't get why the CI is failing right now. It says it can't find `Reason_parser_explain_raw` but this rule clearly should have been executed. Maybe if we include `compile_error: update_error precompile`, it'd work

chenglou:
Patched the makefile. M-M-MERGING

PR #1544, opened 2017-10-24
Bump topkg requirements

chenglou:
@let-def's #1541 seems to require a newer version of topkg that provides
a `-r` flag


jordwalke:
Looks good. Is this why the tests were failing in his PR? Feel free to merge this as soon as these tests pass. Thanks, @chenglou.

chenglou:
No, see https://github.com/facebook/reason/pull/1541#issuecomment-338909827

PR #1558, opened 2017-10-27
Proper Esy support

andreypopp:
Now that esy supports `esy.json` manifests parallel to `package.json` we don't need to maintain a separate `package.opam.json` + a special release procedure.

Therefore this renames `package.opam.json` to `esy.json` and reverts the commit which adds a corresponding release procedure.


chenglou:
Thanks! Will merge this in a week or so. In case the new version has some critical flaws we need to patch soon

jordwalke:
Nice, this makes it so the stock git repo can be used as an npm package or an opam/esy package without conflict.

jordwalke:
Actually, merging this will speed up my dev cycle while trying to remove our `utop` dependency. Merging - feel free to revert if it ends up causing any problems.

chenglou:
Wait we're not removing utop before hot patching the current version...

jordwalke:
> Wait we're not removing utop before hot patching the current version...

Sorry, what do you mean by that?

This diff should have no perceivable effect on anyone except me while I try to remove `utop` dependencies (I merged it only because it helps me iterate faster in the repo).

On the subject, what would you suggest for the order that we remove `utop/rtop` into its own package? It's *technically* a breaking change, but one that shouldn't effect anyone's build processes - only the CLI experience. If we can get away with not bumping a major version for it, it will simplify a lot of things. So the two things that would be required:

A. Factor `rtop` out into its own package.
B. Decide whether or not to bump a major version.
C. Cut a new release with a major/minor version bump depending on what we decide.

What would you suggest for ordering/versioning?

Issue #1559, opened 2017-10-27
compilation fails on doc comments inside of records


zploskey:
Repro:
```reason
type thing = { 
  .
  /**comment**/
  "name": string
};
```
results in the following compile-time error on Reason 1.13.7 w/ bs-platform@1.10.3 compiled (refmt gives the same error):
```
File "/home/user/project/src/doubleast_comment_in_obj.re", line 4, characters 2-8:
Error: 320: <UNKNOWN SYNTAX ERROR>
File "/home/user/project/src/doubleast_comment_in_obj.re", line 1:
Error: Error while running external preprocessor
```
On 3.0.0, refmt on this file gives
```
Error: 291: <UNKNOWN SYNTAX ERROR>
```

Note that when the non-sugared syntax is used it will compile but `refmt` fails, i.e.
```reason
type thing = Js.t {
  .
  /**comment**/
  name: string
};
```

```
$ refmt --in-place src/doubleast_comment_in_obj_desugared.re 
File "src/doubleasc_comment_in_obj_desugared.re", line 1, characters 18-19:
Error: 434: syntax error, consider adding a `;' before
```

Comment placement does not seem to matter as long as it's within the curly braces, it gives an error. Comments with a single opening `*` work fine.

cristianoc:
+ @andreypopp

chenglou:
Likewise for record:

```reason
type bla = {
  /** foo */
  foo: bar
}
```

zploskey:
On 3.0.1, doc comments in a Js.t object gives:
```
Error: 3053: <UNKNOWN SYNTAX ERROR>
```
Using a doc comment in a record gives:
```
Error: 2813: <UNKNOWN SYNTAX ERROR>
```
Can someone more familiar with the codebase tell me where the error codes are coming from and where might be a good place to look at to track down the problem? I guess this would probably be in the parser but I could waste a lot of time trying understand what's going on in there. I'm currently looking at places where Reason_parser_explain.message (the function that prints `<UNKNOWN SYNTAX ERROR>`) gets called and trying to trace back what's going on from there.

zploskey:
[...]

PR #1565, opened 2017-10-28
WIP: fixing #1559, doc comments inside record

zploskey:
Adds a test which triggers #1559, doc comments inside a record.

### DIFF
--- a/src/reason-parser/reason_parser.mly
+++ b/src/reason-parser/reason_parser.mly
@@ -3111,15 +3111,21 @@ expr_optional_constraint:
 record_expr:
   | DOTDOTDOT expr_optional_constraint lnonempty_list(preceded(COMMA,lbl_expr))
     { (Some $2, $3) }
-  | as_loc(label_longident) COLON expr
-    { (None, [($1, $3)]) }
+  | attribute as_loc(label_longident) COLON expr
+    { let rec_expr = {$4 with pexp_attributes = [$1] @ $4.pexp_attributes} in
+      (None, [($2, rec_expr)])
+    }
   | lseparated_two_or_more(COMMA, lbl_expr) COMMA?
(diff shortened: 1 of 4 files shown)


src/reason-parser/reason_parser.mly
```diff
@@ -3111,15 +3111,21 @@ expr_optional_constraint:
...
 
 lbl_expr:
-  | as_loc(label_longident) COLON expr { ($1, $3) }
-  | as_loc(label_longident)            { ($1, exp_of_label $1) }
+  | attribute as_loc(label_longident) COLON expr
+    { ($2, {$4 with pexp_attributes = [$1] @ $4.pexp_attributes} ) }
+  | attribute as_loc(label_longident)
+    { let exp = exp_of_label $2 in
+      ($2, {exp with pexp_attributes = [$1] @ exp.pexp_attributes})
+    }
```
zploskey: This compiles but broke attributes like this
```reason
[@bs.config {foo, jsx: 2}];
```
with this error that also appears many times on other tests:
```
Error: 2823: <UNKNOWN SYNTAX ERROR>
```
and also gave
```
File "./formatTest/unit_tests/input/./basicStructures.re", line 210, characters 25-28:
Error: 304: fun is a reserved keyword, it cannot be used as an identifier. Try `fun_' instead
```
It may be worthwhile trying a different approach unless someone has other ideas on how to make this work. Trying to use `item_attributes` instead introduces shift reduce conflicts that I don't yet really know how to deal with. Since we have the 4.04 AST available potentially we could use the [Docstrings](https://caml.inria.fr/pub/docs/manual-ocaml/libref/Docstrings.html) module to handle docstrings.

zploskey:
Fixed in #1817.

Issue #1567, opened 2017-10-28
[printer bug] a constructor w/ a type-constrained argument prints invalid syntax


jaredly:
old syntax
```
Some (x: int)
```
new syntax
```
Some((x: int))
```
refmt prints:
```
Some(x:int)
```
which fails to parse under the new syntax

PR #1560, opened 2017-10-27
Print`foo(()) as `foo() + update parser

IwanKaramazow:
Fixes https://github.com/facebook/reason/issues/1550

PR #1576, opened 2017-10-30
Allow parsing of constraint expressions without parens inside constructor expr

IwanKaramazow:
Fixes https://github.com/facebook/reason/issues/1567

```reason
/* before */
Some((x: string), (y: int), (b: float));

/* after */
Some(x: string, y: int, b: float);
```
Parser didn't support the form without parens.


chenglou:
Rebasing after having merged #1560. Let's see if this works

chenglou:
Seems working. Merged!

PR #1578, opened 2017-10-30
Prefer git diff over normal diff for test results

### LINKED ISSUE #1575: Use git diff to output diffs for tests
Right now in order to compare expected and actual test results we use `icdiff`, if available, and fall back to normal `diff`. In my experience, having `icdiff` installed makes running the test suite take about twice as long.

The `git diff` tool is fast and produces nice colored diffs. It works very well even for files not checked in to version control as long as you pass the `--no-index` flag. How would folks feel about using git's diff by default if git is available?

zploskey:
Fixes #1575.

### DIFF
--- a/formatTest/test.sh
+++ b/formatTest/test.sh
@@ -27,7 +27,7 @@ fi
 if hash icdiff 2> /dev/null; then
     DIFF="icdiff"
 elif hash git 2> /dev/null; then
-    DIFF="git diff --no-index"
+    DIFF="git --no-pager diff --no-index"
 else
     DIFF="eval diff --unchanged-line-format='' --new-line-format=':%dn: %L' --old-line-format=':%dn: %L'"
 fi
--- a/formatTest/test.sh
+++ b/formatTest/test.sh
@@ -24,10 +24,12 @@ then
 fi
 
 # for better visual diffing in the terminal, try https://github.com/jeffkaufman/icdiff
-if hash icdiff 2>/dev/null; then
-  DIFF="icdiff"
+if hash icdiff 2> /dev/null; then
+    DIFF="icdiff"
+elif hash git 2> /dev/null; then
+    DIFF="git diff --no-index"
 else
-  DIFF="eval diff --unchanged-line-format='' --new-line-format=':%dn: %L' --old-line-format=':%dn: %L'"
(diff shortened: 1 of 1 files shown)


chenglou:
What's that `>` thing on CircleCI diff? https://circleci.com/gh/facebook/reason/964?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

zploskey:
Weird. It's showing `>` before `  ☑ PASS`, but only on Circle, and only on tests other than the typeCheckedTests. Travis doesn't show these characters at all.

zploskey:
The thing that the tests it happens on (unit and idempotent) have in common is they call `stdin_test()`. `typecheck_test()` doesn't.

zploskey:
Likely the real reason is that `typecheck_test()` doesn't do a call to `$DIFF` but the others do. Well that makes some sense then, but why does only shows up on Circle. :man_shrugging:

chenglou:
I just pulled this locally and now it's showing a screen for every test that I have to manually dismiss. Not sure what I misconfigured in git?

zploskey:
Do you currently have `icdiff` installed? It will still be used as a first choice.

There is an an interesting side effect to switching to using git for diffs. Every person can configure the git diff command to display however they want it to, including using an external program. It's possible some people have configured their diffs to use a pager, which might explain why you're having to dismiss a screen for each diff. I think when I tested I probably did not have large enough diffs in expected/actual results to trigger a pager, and instead it just outputs the diff and moves on. It is pretty normal for it to have to display at least 1 terminal screen full to trigger paging. I think this warrants some more testing.

zploskey:
I made a big diff for myself by indenting most of one of the test files and got a pager interrupting the running of the tests. Using `git diff --no-pager` should prevent this.

chenglou:
I get these now:

```
Idempotent Test: ./comments.mli
fatal: invalid diff option/value: --no-pager
⊘ FAILED
```

Yeah, I think that's because I have an external diff tool configured... Any way to override it?

zploskey:
It helps if I put the `--no-pager` flag in the right place.

zploskey:
That got rid of the weird characters. How does your external diff tool fare?

chenglou:
Tested, great! Thanks again =)

PR #1582, opened 2017-11-01
[Reason] Cut Build Time In Half

jordwalke:
Summary:
1. One thing that contributors report as being a major point of friction to contributing
to the `Reason` repo is the long build times. This diff ports the build to `jbuilder` which
reduces the clean build times by around 50-60%!  Incremental build time is also reduced
by 50% when making changes to the parser. I'm a lot more excited about fixing/improving
Reason when the build times are much faster. This brings cold builds down from 24seconds
to 11seconds and incremental builds (when changing parser/printer) from 15seconds to
8seconds.

2. Since this uses `jbuilder`, it also allows easily removing the `utop`
dependency in the next diff. (I don't know how to refactor `ocamlbuild`
projects (`_tags`, `myocamlbuild`, wat.), but here, I just guessed how
to do stuff with `jbuilder` and most of the time it worked.) I pulled apart [Rudi's initial port](https://github.com/facebook/reason/pull/1261) and fixed some of the remaining issues. I also refactored the diff to be a little less aggressive and fixed the remaining issues.

3. Completeness:
 - `make test` works.
 - Documentation in `./src/README.md` updated.
 - `opam pin add reason .` works.
 - I tested `rebuild` on a sample project.
 - I disabled one test momentarily (the oprint test) because it requires
 a modern version of node with `async` which I don't have. I'll reenable
 and test it before we merge this.

Test Plan:make test

Reviewers:@chenglou, @cristianoc, @IwanKaramazow, @rickyvetter

CC: @rgrinberg

Thank you to @rgrinberg for showing how to do the port in your initial diff and for answering questions!


esy.lock
```diff
@@ -0,0 +1,349 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
chenglou: Can you mark this as binary please

rgrinberg:
How come you didn't factor things into multiple opam packages like I did? This will allow downstream users to depend on much less packages if they're publishing reason code on opam.

esy.lock
jordwalke: Thanks for the reminder.

jordwalke:
@rgrinberg: that’s a good idea and I didn’t do it because:
1. I just want to do one step at a time. The next diff will rip out the rtop dependency.
2. Currently the process for submitting multiple dependent opam version releases is really really slow (days). I don’t want to break off anything more than rtop until we get a more streamlined release process.

TODO.md
```diff
@@ -0,0 +1,9 @@
+# Do not merge this pull request until the following is done:
+[x] Make sure rebuild actually works.
+[.] Separate rtop repo/package.
+   [x] created repo:  https://github.com/reasonml/rtop
+   [ ] Move rtop portion there.
+[x] Compare old .install file to new one to make sure we're not missing anything.
```
rgrinberg: I'm surprised to see this comment. I thought people at Facebook understood the value mono repos bring.

Let me just note that jbuilder easily supports multiple packages out of 1 repo. So you're only getting a second git history and no extra modularity from such a split.

Makefile
```diff
@@ -22,8 +22,10 @@ clean-tests:
 	rm -f ./formatTest/failed_tests
 	rm -f ./miscTests/reactjs_jsx_ppx_tests/*.cm*
 
+# Not all versions of jbuilder have the clean command.
+# jbuilder clean
 clean: clean-tests
-	jbuilder clean
+	rm -rf ./_build
```
rgrinberg: Note that this clean command is incomplete. In particular, it will not clean generated .merlin files in the source.

TODO.md
jordwalke: Side note: I don't really think "monorepo" is a meaningful term with the way it's often used, but for any definition, having multiple packages under development in one repo breaks down when it comes time to consumer or produce code for people outside of that organization. But that aside - for my particular grievance here, it doesn't matter if we do use a "monorepo", because we still need to push these individual packages to opam and that particular process is the painful part - not creating multiple packages, or using them in a "monorepo" - but submitting the physical packages to opam when they depend on each other. It takes one day to submit an opam-repository request, and I can't submit the one that depends on the first until the first is merged otherwise the second package's build will fail. The reverse matrix is super cool don't get me wrong, but it typically ends up taking *two* entire days to release two packages. I understand the value of having a stable opam universe, and so I'm okay with that - but we need tools that let us rapidly release our own stack, (and get others to instantly consume that stack) bypassing opam-repository, and then gradually sync those multiple released packages to opam.

rgrinberg:
The grievances with the opam release process are indeed valid. But I just
don't see how maintaining more git histories directly solves this problem.
We should make suggestions that solve the problem at the root.

But I'm also not sure why you think that more opam packages out of 1 repo
will complicate anything. When you make a release, you will submit all
packages in a single PR. So you will not have the problem of chasing
multiple dependent PRs. As you've pointed out, that would be hell with the
current system.

As for your point about monorepos breaking down for consumers outside the
org. I don't see it as breaking down, rather it complicates things when
your tooling is bad, but it's a much lesser evil than maintaining a
compatibility matrix between your own packages and maintaining accurate
constraints between them. This is a task that nobody in the opam world has
yet to figure out how to do effectively. Not to mention all the problems
associated with having a split git history: reverts are hard, collaborating
on large refactoring, fixing bugs that span multiple repos, etc.

Anyway I won't push beyond this. Theres definitely a subjective aspect to
this issue.

Makefile
jordwalke: Thanks.

jordwalke:
Those are definitely some great points. I think the thing we were missing was the workflow to submit multiple sub-packages to opam-repository in one pull request. For our next release, perhaps we should setup something that does so.
The one downside, is that a one-repo to one-package relationship makes `git` dependencies very natural:
It allows us to point "opam dependencies" to specific git commits of packages in their original repos.

(Note that our solution for managing monorepos with external dependencies at Facebook is yarn "lock files").

IwanKaramazow:
I'm so excited about this PR! 😃
Seems like everything looks good, we can defer decisions about mono-repos to a later point in time.
There are a lot of improvements to be made in refmt that might sooner land with the faster build times.

jordwalke:
Confirmed that bspacking works correctly now as well. Merging!

TODO.md
avsm: >  It takes one day to submit an opam-repository request, and I can't submit the one that depends on the first until the first is merged otherwise the second package's build will fail

Just submit them both in the same PR, and they will be batch tested individually with the right dependency checks.

jordwalke: @avsm Thanks, we're investigating a couple of approaches to doing that. It would be great to be able to keep multiple opam packages in one repo.

"and they will be batch tested individually with the right dependency checks."

Do you mean they will be batch tested simultaneously?

PR #1583, opened 2017-11-01
Remove useless jbuilder rule

jordwalke:
Lets see what this breaks if anything!


jordwalke:
Not so useless after all!

PR #1584, opened 2017-11-01
Fix bspack

jordwalke:


jordwalke:
I also added a separate `make test-ci` which does an opam pin, so that we don't have to use opam to test it locally. But CI is a good way to make sure we don't break the opam install.

jordwalke:
All green!

PR #1586, opened 2017-11-02
Add all-supported-ocaml-versions target

rgrinberg:
This target will build & test the entire project against a workspace that
defines a set of switches where reason is supported.

We use this little trick to make sure that jbuilder works on every version of OCaml. I think you might find it useful too.


jordwalke:
Sure! Thanks.

PR #1588, opened 2017-11-02
don't print fun in record expressions with Pexp_fun values

### LINKED ISSUE #1587: Annotated return values cause refmt to print "fun" keyword.
> `echo "{contents: () => (() : unit) }" | refmt`
Outputs:
> `{contents: fun () => ((): unit)};`

Thankfully it still parses with the `fun` keyword, but there should be nothing wrong with omitting the fun keyword. The original thinking might have been that to add an annotation to the return value of the lambda, you'd need a `fun` keyword for this form `fun (args) : retType => retVal`. There is no  difference in AST between `fun (x) : unit => ()` and `() => (() : unit)`, but `fun` is only required for the first form. The bug is that when `refmt` tries to print the second form, it adds the unnecessary `fun` keyword. Easy fix, marking GoodFirstTask.

IwanKaramazow:
fixes #1587 

```reason
/* before */
{contents: fun () => (() : unit)};

/* after */
{contents: () => (() : unit)};
```

### DIFF
--- a/formatTest/unit_tests/expected_output/basicStructures.re
+++ b/formatTest/unit_tests/expected_output/basicStructures.re
@@ -777,3 +777,6 @@ let () = {
   exception E;
   raise(E)
 };
+
+/* # 1587: don't print fun keyword when printing Pexp_fun in a record expression  */
+{contents: () => ((): unit)};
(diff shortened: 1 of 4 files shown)


chenglou:
Awesome as always

Issue #1556, opened 2017-10-27
Nested Records/objs printed on the same line


chenglou:
`let person = {"name": {"first": "Bob", "last": "Zhmith"}, "age": 32};`

Should probably break onto many lines because of the nested record/obj

PR #1593, opened 2017-11-03
Force breaks for nested records

IwanKaramazow:
Fixes https://github.com/facebook/reason/issues/1556 (partially) 

```
/* before */
let person = {"name": {"first": "Bob", "last": "Zhmith"}, "age": 32};

/* after */
let person = {
  "name": {
    "first": "Bob",
    "last": "Zhmith"
  },
  "age": 32
};
```
Nested records/bs-obj sugar will always be printed with breaks.
Improves readability/extensibility of the code a lot.

Note that it only works with record or bs-obj sugar which contains another record/bs-obj/objects.
Force breaking for nested objects hasn't been implemented yet, because of some printing challenges.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,9 @@ clean-tests:
 	rm -f ./formatTest/failed_tests
 	rm -f ./miscTests/reactjs_jsx_ppx_tests/*.cm*
 
+testFormat: build clean-tests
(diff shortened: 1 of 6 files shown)


chenglou:
Thanks!

PR #1596, opened 2017-11-04
Always break object def with two or more rows

### LINKED ISSUE #1595: Object and record line breaking inconsistent
Repro:

```reason
type same_length = {
  title: string,
  render: unit => ReasonReact.reactElement
};

type length_same = {. title: string, render: unit => ReasonReact.reactElement};
```

These two should break in the same way

IwanKaramazow:
Fixes #1595 
```reason
/* before */
type length_same = {. title: string, render: unit => ReasonReact.reactElement};

/* after */
type length_same = {
  .
  title: string,
  render: unit => ReasonReact.reactElement
};
```


chenglou:
Looks good, thanks!

PR #1598, opened 2017-11-05
[Reason] Make ++, +++, ++(any operator char) right associative.

jordwalke:
Summary:This maintains compatibility with `^`, which means when
converting from old Reason to new Reason, we don't need to manually edit
any parenthesis.

The downsides:
- people have already upgraded their code to be grouped in the other
manner.
- Depending on whether we consider the left associativity a bug or not,
this could be a "breaking change" worthy of version bumping.
- If people upgraded to this change without running a conversion script,
they could have bugs in their code - but only if they overloaded ++ to
be something other than string concat, or where order matters.

Test Plan:

Reviewers:

CC:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ test-ci: install test
 
 test: build clean-tests
 	# I don't have modern enough node to test. brb.
-	node ./formatTest/testOprint.js
+	# node ./formatTest/testOprint.js
 	./miscTests/rtopIntegrationTest.sh
 	./miscTests/jsxPpxTest.sh
 	cd formatTest; ./test.sh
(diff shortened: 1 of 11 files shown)


jordwalke:
I propose that we consider the previous behavior a bug, and consider this the fix. The downside is that anyone who overloaded the `++`/`+++` operators to represent functions where associativity matters, will experience different behavior (bugs).

People who manually removed parens from string concatenation when upgrading to v3 don't need to do anything to their code - this version will just work for them, but new people who convert to v3 will not have those extra parens on their string concats.

jordwalke:
Thanks to @rgrinberg this now passes on 4.02.3! Merging.

PR #1609, opened 2017-11-08
Fix build instructions

rgrinberg:
This will make them resilient against adding multiple opam packages to the repo.
And also, it will use the number of jobs as set by OPAM.

### DIFF
--- a/reason.opam
+++ b/reason.opam
@@ -10,8 +10,9 @@ bug-reports: "https://github.com/facebook/reason/issues"
 dev-repo: "git://github.com/facebook/reason.git"
 tags: [ "syntax" ]
 build: [
-  ["jbuilder" "build"]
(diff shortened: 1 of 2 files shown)


jordwalke:
Looks like the build is failing because lambda-term deprecated support for `4.02.3`. Do you think we could convince the maintainer to continue to support `4.02.3` for us? Here's the error:

```
Error: This expression has type bytes but an expression was expected of type
#          string
```

I think we should still merge this diff as is because it's clear the failure isn't related to your change.

rgrinberg:
Lambda-term supports 4.02.3 it just requires a new release of camomile to
do that. I believe we already discussed that

jordwalke:
Ah, I think I just misunderstood previously. Thank you. I'll merge this.

Issue #1608, opened 2017-11-07
(Less Hacky) OCaml 4.06.0 support


hcarty:
It almost works!  A constructor changed in part of the compiler internals - not sure what exactly this comes from - but this crude hack is enough to make Reason build under 4.06.0.

```diff
diff --git a/src/reason-parser/reason_lexer.mll b/src/reason-parser/reason_lexer.mll
index 8e057ca..4a83b40 100644
--- a/src/reason-parser/reason_lexer.mll
+++ b/src/reason-parser/reason_lexer.mll
@@ -315,8 +315,9 @@ let preprocessor = ref None
 (* Warn about Latin-1 characters used in idents *)
 
 let warn_latin1 lexbuf =
-  Location.prerr_warning (Location.curr lexbuf)
-    (Warnings.Deprecated "ISO-Latin1 characters in identifiers")
+  let loc = Location.curr lexbuf in
+  Location.prerr_warning loc
+    (Warnings.Deprecated ("ISO-Latin1 characters in identifiers", loc, loc))
 ;;
 
 (* Error report *)
```
I'm sure that there is a cleaner way to fix this.  I'm not sure, however, how to fix this and keep compatibility with previous OCaml versions without using something like cppo.

jordwalke:
Thank you, @hcarty. I think the simplest solution is to just include the implementation of 4.06.0's `prerr_warning` at the top of the `reason_lexer.mll` file. Then it should work in all versions.

PR #1629, opened 2017-11-11
OCaml 4.06.0 support

ncihnegn:
Code copied from OCaml 4.06.0

#1608

### DIFF
--- a/esy.json
+++ b/esy.json
@@ -5,7 +5,7 @@
     "@esy-ocaml/substs": "^0.0.1",
     "@esy-ocaml/esy-installer": "^0.0.0",
     "@opam/ocamlfind": "",
-    "@opam/menhir": " >= 20170418.0.0  <= 20170712.0.0",
+    "@opam/menhir": " >= 20170418.0.0  <= 20171013.0.0",
     "@opam/utop": " >= 1.17.0",
     "@opam/merlin-extend": " >= 0.3.0",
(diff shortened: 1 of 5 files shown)


jordwalke:
Thanks! It looks like it's failing on (at least) 4.04 with "Location.deprecated" not being found.

ncihnegn:
Not sure how to make it work for pre-4.06 versions. It is due to Warnings.Deprecated type change.

chenglou:
cc @hcarty

rgrinberg:
Using cppo comes to mind...

jordwalke:
Note that if you rebase, you'd pick up some commented code from OCaml that I copy/pasted into the repo, which would be useful for making the tests pass on all OCaml versions perhaps. I'm looking to add 4.06 support myself, so if there's anything else I can do to help, please feel free to reach out to me on Discord and I can help get this through.

jordwalke:
Is there a reason why you bumped the menhir version?

ncihnegn:
Just want to use the latest. My bad. I will refrain from adding unrelated commits.

hcarty:
If it works with the latest menhir then it's probably worth reflecting that in the version constraints.

jordwalke:
I've made this work with all ocaml compiler versions in [this pr](https://github.com/facebook/reason/pull/1663) and I actually had to eventually bump menhir as you have in this diff. Thank you!

PR #1631, opened 2017-11-11
Remove extra parens wrapping for type applications that contain arrow

### LINKED ISSUE #1616: Printer: extra parentheses around function in parametric types
```reason
type foo = option(int => int);
```

Turns into

```reason
type foo = option((int => int));
```

chenglou:
Fixes #1616

### DIFF
--- a/formatTest/unit_tests/expected_output/wrappingTest.re
+++ b/formatTest/unit_tests/expected_output/wrappingTest.re
@@ -2514,11 +2514,10 @@ type wrappingGadt(_) =
   | ThisIsLongSoTypeWillWrap(int): wrappingGadt(
                                      int
                                    )
-  | Add: wrappingGadt(((int, int) => int))
-  | App(
-         wrappingGadt(('b => 'a)),
-         wrappingGadt('b)
-       ): wrappingGadt('a);
+  | Add: wrappingGadt((int, int) => int)
+  | App(wrappingGadt('b => 'a), wrappingGadt('b)): wrappingGadt(
+                                        'a
+                                        );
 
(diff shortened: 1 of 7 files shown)


jordwalke:
This is probably okay, but can you copy/paste the test cases you added, and add attributes at various parts on the types (both inside and outside the parens etc)? The code you're calling into now deals with attributes and I want to make sure this didn't break any of those cases.

formatTest/unit_tests/expected_output/wrappingTest.re
```diff
@@ -2514,11 +2514,10 @@ type wrappingGadt(_) =
...
                                    )
-  | Add: wrappingGadt(((int, int) => int))
-  | App(
-         wrappingGadt(('b => 'a)),
-         wrappingGadt('b)
-       ): wrappingGadt('a);
+  | Add: wrappingGadt((int, int) => int)
+  | App(wrappingGadt('b => 'a), wrappingGadt('b)): wrappingGadt(
+                                        'a
+                                        );
```
IwanKaramazow: 🙈

chenglou: Unrelated right

jordwalke: Yeah what’s up there.

IwanKaramazow: I can fix this later today

chenglou:
I've put some attributes on everything

IwanKaramazow:
Looks solid, should be ok.

PR #1633, opened 2017-11-11
Fix refmt.js example and add important information about avoiding mis…

glennsl:
…sing module error wiht webpack


chenglou:
Thanks!

Issue #1630, opened 2017-11-11
Printer: extra parentheses around variants with payload


chenglou:
```reason
let x = Some(1);

let y = Some(-1);
```

Formats to

```reason
let x = Some(1);

let y = Some((-1));
```

Likewise for `foo(-1)`

@jordwalke I see `method constant` in the printer. Does that mean we need to pass down an optional `wrap` all the way through?

PR #1634, opened 2017-11-11
remove unnecessary parens around float & integer constants in constructor expr/fun app

IwanKaramazow:
fixes https://github.com/facebook/reason/issues/1630

```reason
/* before */
Some((-1));

funAppl((-1));

/* after */
Some(-1);

funAppl(-1);
```


formatTest/typeCheckedTests/expected_output/reasonComments.re
```diff
@@ -268,8 +268,8 @@ let result =
  */
 let res =
   switch (
-    /* Retain this */
-    X(2, 3)
+    X
+      (2, 3) /* Retain this */
```
IwanKaramazow: not sure why this happening

src/reason-parser/reason_pprint_ast.ml
```diff
@@ -3674,6 +3688,12 @@ class printer  ()= object(self:'self)
       if ensureExpr then
         makeList ~wrap:("(", ")") children
       else makeList children
+    | { pexp_attributes = [] ; pexp_desc = Pexp_constant c } ->
+      (* When we have Some(-1) or someFunction(-1, -2), the arguments -1 and -2
+       * pass through this case. In this context they don't need to be wrapped in extra parens
+       * Some((-1)) should be printed as Some(-1). This is in contrast with
+       * 1 + (-1) where we print the parens for readability. *)
+      self#constant ~parens:(false || ensureExpr) c
```
chenglou: Can remove false

formatTest/unit_tests/expected_output/syntax.re
```diff
@@ -1072,3 +1072,35 @@ Some(x: int);
...
+
+Some(-0.1, -0.1, -0.1);
+
+Some(-0.1, -0.1, -0.1);
+
+Some(-0.1G, -0.1x, -0.1H);
+
+Some(-0.1G, -0.1x, -0.1H);
+
+Some([@foo] (-1), [@foo] (-1), [@foo] (-1));
```
IwanKaramazow: It's possible to not print the parens in this case, should we remove them?

chenglou: I think so? The fewer the better =D

jordwalke:
Did you see there was a high pri bug someone found with unary negation and named arguments? Does this make those matters worse? Worth adding a test case (even if it has to be disabled)

IwanKaramazow:
Hmm I've seen it, this PR seems like a good opportunity to fix the high pri bug too.

chenglou:
Should I wait til you add more tests?

IwanKaramazow:
Yes, with a bit of a luck I can fix it too.

IwanKaramazow:
@chenglou this is ready,  added extra tests & removed the extra parens when printing with attributes

src/reason-parser/reason_pprint_ast.ml
```diff
@@ -2334,17 +2334,31 @@ let paren: 'a . ?first:space_formatter -> ?last:space_formatter ->
...
-    paren (i.[0]='-') (fun f (i, m) -> pp f "%s%c" i m) f (i,m)
-  | Pconst_float (i, None) -> paren (i.[0]='-') (fun f -> pp f "%s") f i
-  | Pconst_float (i, Some m) -> paren (i.[0]='-') (fun f (i,m) ->
+      let parens = if parens then (i.[0]='-') else false in
+      paren parens (fun f (i, m) -> pp f "%s%c" i m) f (i,m)
+  | Pconst_float (i, None) ->
+      let parens = if parens then (i.[0]='-') else false in
+      paren parens (fun f -> pp f "%s") f i
+  | Pconst_float (i, Some m) ->
+      let parens = if parens then (i.[0]='-') else false in
```
chenglou: This can be simplified to `parens && i.[0]='-'`

IwanKaramazow: done

chenglou:
Merging! Thanks.

Issue #1635, opened 2017-11-11
BS Js.t obj value with field with attribute doesn't print correctly


chenglou:
Repro:

```ml
type y = <
  height : int [@bs.set no_get] (1)
> Js.t
```

This ocaml code should print into this reason code:

```reason
type y = {.
  [@bs.set no_get] "height" : int
};
```

Instead it prints into

```reason
type y = {.
  [@bs.set no_get] height : int
};
```

Then fails to parse.

cc @IwanKaramazow

IwanKaramazow:
Will get it in later today, high pri.

PR #1637, opened 2017-11-12
parse and print attributes on object type rows

IwanKaramazow:
Fixes https://github.com/facebook/reason/issues/1635

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/attributes.re
+++ b/formatTest/typeCheckedTests/expected_output/attributes.re
@@ -381,10 +381,10 @@ module Js = {
 
 type classAttributesOnKeys = {
   .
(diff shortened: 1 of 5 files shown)


chenglou:
Too bad the attribute can carry more info like `[@@bs.set no_get]`. Otherwise we can have `{mutable "foo": bar}`. cc @bobzhang

bobzhang:
Do you find set only attributes useful?

chenglou:
Haven't used any of these much yet

chenglou:
Aaaalright, is this ready?

IwanKaramazow:
Ready 🙌

chenglou:
Good stuff

Issue #1639, opened 2017-11-14
External with annotation line break logic wrong now


chenglou:
Sorry for the vague title. Can't describe it well.

```reason
[@bs.splice] [@bs.val] [@bs.module "React"]
external createElement : (Class.t, ~props: Js.t({..})=?, array(t)) => t = "createElement";
```

This formats to:

```reason
[@bs.splice] [@bs.val] [@bs.module "React"]
external createElement : (Class.t, ~props: Js.t({..})=?, array(t)) => t =
  "createElement";
```

Though if we remove the annotations, it formats to:

```reason
external createElement : (Class.t, ~props: Js.t({..})=?, array(t)) => t = "createElement";
```

The first example shouldn't break; the last one certainly doesn't.

I think the annotations might be playing a role in breaking the line here, though they shouldn't. Maybe they are because of some legacy logic when they were attached at the end? Or maybe something else.

cc @IwanKaramazow

PR #1640, opened 2017-11-14
fix breaking primitive declaration with attributes

IwanKaramazow:
fixes https://github.com/facebook/reason/issues/1639

```reason
/* before */
[@bs.val] external imul : (int, int) => int =
  "Math.imul";

/* after */
[@bs.val] external imul : (int, int) => int = "Math.imul";
```

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/attributes.re
+++ b/formatTest/typeCheckedTests/expected_output/attributes.re
@@ -372,8 +372,8 @@ let myFun =
    let myFun = fun ((X(hello) | Y(hello)) [@onOrPattern]) => hello;
    */
 /* Bucklescript FFI item attributes */
-[@bs.val] external imul : (int, int) => int =
-  "Math.imul";
+[@bs.val]
+external imul : (int, int) => int = "Math.imul";
 
 module Js = {
   type t('a);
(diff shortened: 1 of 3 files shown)


chenglou:
Aha, so did I diagnose right? Is it because it thought attributes were part of "first part"?

Ping me if you think it's ready to merge

formatTest/typeCheckedTests/expected_output/attributes.re
```diff
@@ -372,8 +372,8 @@ let myFun =
    let myFun = fun ((X(hello) | Y(hello)) [@onOrPattern]) => hello;
    */
 /* Bucklescript FFI item attributes */
-[@bs.val] external imul : (int, int) => int =
-  "Math.imul";
+[@bs.val]
+external imul : (int, int) => int = "Math.imul";
```
jordwalke: This is exactly how I'd want it to break.
Next, imagine if we drop the `[ ]` for "simple enough" attributes:

```reason
@bs.val
external imul : (int, int) => int = "Math.imul";
```

IwanKaramazow:
@chenglou Indeed, attributes where attached to the first part (all stuff up to `=`), which resulted in the printing going the wrong road.

PR #1642, opened 2017-11-15
Fix: refmt3 with unary -. in labelled arguments

### LINKED ISSUE #1580: refmt3 with unary -. in labelled arguments
the following code works,

```reason
ReasonJs.Canvas2d.translate(~x=(-. actualSize /. 2.0), ~y=(-. actualSize /. 2.0), context);
```

but refmt reformats as

```reason
ReasonJs.Canvas2d.translate(~x=-. actualSize /. 2.0, ~y=-. actualSize /. 2.0, context);
```

which throws a syntax error

anmonteiro:
fixes #1580

### DIFF
--- a/formatTest/unit_tests/expected_output/syntax.re
+++ b/formatTest/unit_tests/expected_output/syntax.re
@@ -1070,3 +1070,5 @@ Some(x: int);
 Some(x, y: int, b);
 
 Some(x, y: int, b);
+
+foo(~x=-. bar);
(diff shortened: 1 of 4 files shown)


chenglou:
cc @IwanKaramazow @jordwalke

jordwalke:
Great fix! Thank you.

IwanKaramazow:
Looks good, nice job!
There might be another trick to solve this without modifying the parser:
[set_lexeme_length](https://github.com/facebook/reason/blob/master/src/reason-parser/reason_lexer.mll#L186)

If you add a rule:
```ocaml
| "=-"  {
  set_lexeme_length lexbuf 1;
  EQUAL
}
```
This trick lets you implement look-ahead: you match a longer string which is not completely consumed. The lexer will resume lexing at "-" and should return MINUS or whatever token is matched in combination with MINUS.

Issue #1292, opened 2017-06-06
Refmt nested jsx syntax error


rickyvetter:
Input:

```
<Foo bar=(<Baz />) />;
```

Output:

```
<Foo bar=<Baz /> />;
```

Which is a syntax error.

Issue #1181, opened 2017-03-31
`<div asd=1></div>` gives bad error


chenglou:
```
Error: 2661: This is a reserved keyword. Consider using a different one. For BuckleScript, add an underscore at the end (http://bloomberg.github.io/bucklescript/Manual.html#_object_label_translation_convention).
```

In reality the error is a missing space between the tags.

From #1018

PR #1645, opened 2017-11-15
Fix subtle jsx parsing issues.

IwanKaramazow:
Fixes https://github.com/facebook/reason/issues/856 https://github.com/facebook/reason/issues/904 https://github.com/facebook/reason/issues/1181 https://github.com/facebook/reason/issues/1263 https://github.com/facebook/reason/issues/1292

Allow parsing of following "problematic jsx":
```reason
let listy = [<div />, <div />];
let x =[|<div />|];
let x=<div />;
let y = (<div />, <div />);
<Foo bar=<Baz /> />;
foo#=<bar />;
<div asd=1></div>;
```

### DIFF
--- a/src/reason-parser/reason_lexer.mll
+++ b/src/reason-parser/reason_lexer.mll
@@ -459,18 +459,6 @@ rule token = parse
       }
   | blank +
       { token lexbuf }
-  | "::" lowercase identchar *
-      {
-        let delim = Lexing.lexeme lexbuf in
-        let delim = String.sub delim 2 (String.length delim - 2) in
-        COLONCOLONLIDENT delim
-      }
-  | lowercase identchar * "::"
-      {
-        let delim = Lexing.lexeme lexbuf in
-        let delim = String.sub delim 0 (String.length delim - 2) in
-        LIDENTCOLONCOLON delim
-      }
   | "_"
       { UNDERSCORE }
   | "~"
@@ -538,6 +526,11 @@ rule token = parse
         let esc = String.sub l 1 (String.length l - 1) in
         raise (Error(Illegal_escape esc, Location.curr lexbuf))
       }
+  | "#=<" {
+    (* Allow parsing of foo#=<bar /> *)
+    set_lexeme_length lexbuf 2;
+    SHARPOP("#=")
+  }
   | "#" operator_chars+
       { SHARPOP(lexeme_operator lexbuf) }
   | "#" [' ' '\t']* (['0'-'9']+ as num) [' ' '\t']*
@@ -567,7 +560,6 @@ rule token = parse
   | ";"  { SEMI }
   | ";;" { SEMISEMI }
   | "<"  { LESS }
-  | "<-" { LESSMINUS }
   | "="  { EQUAL }
   | "["  { LBRACKET }
   | "[|" { LBRACKETBAR }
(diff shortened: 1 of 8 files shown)


chenglou:
Holy crap, cc @jordwalke @let-def

IwanKaramazow:
There's one "dramatic" case left:
```
let listy = [<div />, <div />];
```
it clashes with poly-variant types. If we change the syntax for poly-variant types, we can pretty much remove all the very weird stuff in the parser.

IwanKaramazow:
^ nevermind, it's fixed without crazy modifications.

This parses too:
```reason
let listy = [<div />, <div />];
```

src/reason-parser/reason_lexer.mll
```diff
@@ -538,6 +526,11 @@ rule token = parse
         let esc = String.sub l 1 (String.length l - 1) in
         raise (Error(Illegal_escape esc, Location.curr lexbuf))
       }
+  | "#=<" {
```
jordwalke: Could this never otherwise be a valid infix/ppx extension token on its own? Do we want to require that what follows the `<` is an identifier without any white space in between?

src/reason-parser/reason_lexer.mll
```diff
@@ -595,6 +587,25 @@ rule token = parse
   *)
   | "}"  { RBRACE }
   | ">}" { GREATERRBRACE }
+  | "=<" {
```
jordwalke: What would happen if someone defines an infix operator `=<`?

```reason
let (=<) = (a, b) => a + b;
let x = 10;
let y = 20;
let result = x =< y;
```
Today, the result is `30`. After this change, what happens?

I think it would parse it as `LET ident EQUAL ident EQUAL LESSTHAN ident`, which is not what we want.

I think what you can do is be more specific here, and only do this lexing backtracking in the case that you have `=<ident` without any whitespace. If you do that, then it would mean that after the change, we could still parse:

```reason
let result = x =< y;
```

But we couldn't parse:

```reason
let result = x=<y;
```

Because there's no white space between the `<` and the identifier `y`.

I think it's fine to say "=<` as an infix operator must have white space after the `<`, but not fine to say "`=<` is not longer a valid infix operator".


If I'm wrong about this diff, perhaps we could have some test cases demonstrating that it's not an issue right now.

IwanKaramazow:
I added a few extra test cases to make sure that existing user-defined infix operators stay valid.
The following convention is introduced for `=<`.
If it's followed by some form of whitespace, newline or `)`, it's an infix op.
```reason
let x = y =< z;
let (=<) = (a, b) => a + b;
let x = y

=<
z;
```
^ lexes as infix op.
Apparently `=<` is the only case which needs this logic. Operators like `[<` or `[|<`, weren
t/aren't even valid infix operators to begin with. (Same in Ocaml)

src/reason-parser/reason_lexer.mll
```diff
@@ -587,7 +608,7 @@ rule token = parse
   *)
   | "}"  { RBRACE }
   | ">}" { GREATERRBRACE }
-  | "=<" {
+  | "=<" is_jsx_not_infix_operator  {
```
jordwalke: What about the infix identifier ` =</>` ?

`/` is not white space, not a newline and not a `)`, so won't this rule incorrectly fire, and then back up the lexer?

Would it be simpler to have a rule like `"=<" uppercase_or_lowercase+` that backs up the lexer? We have a similar rule for:

```
"></" uppercase_or_lowercase+
```

IwanKaramazow:
Added an extra test and removed `is_jsx_not_infix_operator` in favour of `uppercase_or_lowercase+`. Wasn't the best idea to begin with...

src/reason-parser/reason_lexer.mll
```diff
@@ -608,7 +587,7 @@ rule token = parse
   *)
   | "}"  { RBRACE }
   | ">}" { GREATERRBRACE }
-  | "=<" is_jsx_not_infix_operator  {
+  | "=<" uppercase_or_lowercase+ {
```
jordwalke: Cool, we might run into a few more cases like this and if we do, the fix could be similar to this one.

src/reason-parser/reason_lexer.mll
```diff
@@ -595,6 +587,25 @@ rule token = parse
...
+    set_lexeme_length lexbuf 2;
+    SLASHGREATER
+  }
+  | "[|<" {
+    set_lexeme_length lexbuf 2;
+    LBRACKETBAR
+  }
+  | "></" uppercase_or_lowercase+ {
+    (* allow parsing of <div asd=1></div> *)
+    set_lexeme_length lexbuf 1;
```
jordwalke: There's probably a few other jsx parsing hacks I could have avoided if I knew about htis `set_lexme_length` trick.

PR #1650, opened 2017-11-18
[WIP] Implement different treatment for empty children and disallowed children

wokalski:
This PR is work in progress. It will be finished if you like the idea. The remaining part is `typeCheckedTests/expected_output/jsx.re` test.

It just shows an idea how to make components which disallow children compatible with JSX.

I'm waiting for feedback 👍.

The behaviour has only changed for upper cased components:

`<Div> </Div>` is pretty printed as `<Div> </Div>`. And after desugaring it becomes `((Div.createElement ~children:[] ())[@JSX ])`.

However, `<Div />` is desugared as `((Div.createElement ~children:() ())[@JSX ])` (take a look at `children` arg)

### DIFF
--- a/miscTests/reactjs_jsx_ppx_tests/expected_v2_1.re
+++ b/miscTests/reactjs_jsx_ppx_tests/expected_v2_1.re
@@ -27,6 +27,8 @@ let divRef = ReactDOMRe.createElement("div", [||]);
 
 ReactDOMRe.createElement("div", [||]);
 
+ReactDOMRe.createElement("div", [||]);
+
 ReactDOMRe.createElement("div", ~props=ReactDOMRe.props(~className="hello", ()), [||]);
 
 ReactDOMRe.createElement(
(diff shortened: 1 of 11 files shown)


jordwalke:
Interesting idea! I'm wondering if it's possible to achieve the best of all of these.

- Components can disallow children entirely.
- Components can support children optionally.
- You can use `<Div />` or `<Div> </Div>` with both kinds.

wokalski:
Good idea. The consequence would be that children is always a labeled argument followed by unit which is very explicit. Probably not v useful but another thing would be forcing a component to have children.

jordwalke:
> One downside is that the dotdotdot child always would be labelled as children which is not ideal but ok with me.

Doesn't every proposal have that behavior as well?

wokalski:
When you go with the naive approach I suggested it's not the case - i.e. the last arg - children is not labelled so it can be named anything, for example `callback`. Alternatively the `dot dot dot child` could desugar to the last arg (instead of unit) which would cause the old behaviour but it becomes a bit complicated then.

Basically the approach I suggested is optimising for component definition whereas yours optimises for the call site. I'd personally like to optimise for the former but it introduces too much noise in render functions in comparison to the one you suggested.

jordwalke:
I see. So do you feel that it's somewhat negative to require that children be a named argument?

wokalski:
Only in the case of a callback for instance passed as a child but it's an edge case, i.e. something I never use.

If there're no more doubts I'm gonna go with implementing the labelled children approach.

PR #1651, opened 2017-11-19
Enforce final semicolon for sequences.

jordwalke:
Summary:

First step:  Simplify detection of M.(P.(foo)).
Second step: Enforce the final semicolon for multi-line statements/sequences. (First step doesn't have much to do with it, but it helped me understand what was going on).

Reviewers:chenglou, iwankaramazow

### DIFF
--- a/src/reason-parser/reason_pprint_ast.ml
+++ b/src/reason-parser/reason_pprint_ast.ml
@@ -2150,7 +2150,7 @@ let makeLetSequence letItems =
     ~newlinesAboveComments:0
     ~newlinesAboveItems:0
     ~newlinesAboveDocComments:1
-    ~renderFinalSep:false
+    ~renderFinalSep:true
     ~postSpace:true
     ~sep:";"
     letItems
@@ -2181,7 +2181,7 @@ let makeUnguardedLetSequence letItems =
     ~indent:0
     ~newlinesAboveItems:0
     ~newlinesAboveDocComments:1
-    ~renderFinalSep:false
+    ~renderFinalSep:true
     ~postSpace:true
     ~sep:";"
     letItems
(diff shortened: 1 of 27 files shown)


src/reason-parser/reason_pprint_ast.ml
```diff
@@ -1759,9 +1765,11 @@ let consolidateSeparator = preOrderWalk (function
        when listConfig.sep <> ""
          && listConfig.sepLeft
     ->
+     (* TODO: (perf) cache List.length *)
      let layoutsWithSepAndComment =
        List.mapi (fun i layout ->
            (* Do not render the final separator *)
+           (* TODO: Support !sepLeft, and this should apply to the *first* separator if !sepLeft.  *)
```
jordwalke: Thought I'd leave a couple of notes about some quadradic running times I've found while refactoring.

src/reason-parser/reason_pprint_ast.ml
```diff
@@ -1759,9 +1765,11 @@ let consolidateSeparator = preOrderWalk (function
        when listConfig.sep <> ""
          && listConfig.sepLeft
     ->
+     (* TODO: (perf) cache List.length *)
```
jordwalke: Thought I'd leave a couple of notes about some quadradic running times I've found while refactoring.

jordwalke:
Some other benefits:

- This will reduce the amount of times people confuse `{x}` for a single field punned record because  people will be trained to know that sequences end with `;`.

This doesn't at all make it harder for us to later remove more semicolons - it only makes the current dev experience better while semicolons are generally required. For example, right now, since semis are mandatory in between let statements, people end up introducing errors when adding a line after the final item in a let sequence because `refmt` would remove that semicolon.

In a world where we relax the need for semicolons, this diff wouldn't be helpful - but we're not in that world - yet.

ubsan:
A sequence is at the module, not expression, level, no?

jordwalke:
Does anyone see any issues with this? It seems to address common feedback we've heard in Discord, and I don't *think* it has any downsides. In the future, I think we can do better by only printing the final semi if it precedes a newline, and therefore allow more things to be expressed inline.

PR #1663, opened 2017-11-27
[Reason] Support OCaml 4.06

jordwalke:
Summary: As titled.

Test Plan:Tested with esy ocaml 4.06.

Reviewers:

CC:

### DIFF
--- a/jbuild-workspace.dev
+++ b/jbuild-workspace.dev
@@ -1,5 +1,0 @@
-;; This file is used by `make all-supported-ocaml-versions`
-(context ((switch 4.02.3)))
-(context ((switch 4.03.0)))
-(context ((switch 4.04.2)))
-(context ((switch 4.05.0)))
--- a/src/reason-parser/reason_lexer.mll
+++ b/src/reason-parser/reason_lexer.mll
@@ -348,10 +348,13 @@ let preprocessor = ref None
 *)
 
 
+let latin1IdentifierMessage =
+  "ISO-Latin1 characters in identifiers - This error is being incorrectly reported as " ^
+  "'unused variable warning' but it's really an issue with the characters in your identifier. "
+
 let warn_latin1 lexbuf =
   Location.prerr_warning (Location.curr lexbuf)
-    (Warnings.Deprecated "ISO-Latin1 characters in identifiers")
-;;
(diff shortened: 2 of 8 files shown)


src/reason-parser/reason_lexer.mll
```diff
@@ -348,10 +348,13 @@ let preprocessor = ref None
...
 
+let latin1IdentifierMessage =
+  "ISO-Latin1 characters in identifiers - This error is being incorrectly reported as " ^
+  "'unused variable warning' but it's really an issue with the characters in your identifier. "
+
 let warn_latin1 lexbuf =
   Location.prerr_warning (Location.curr lexbuf)
-    (Warnings.Deprecated "ISO-Latin1 characters in identifiers")
-;;
+    (Warnings.Unused_var latin1IdentifierMessage)
```
hhugo: Why not just use `Location.deprecated` and some conditional compilation.
This is how the issue is handled in ppx_ast https://github.com/janestreet/ppx_ast/blob/master/src/jbuild#L27

jordwalke: We'll probably do something like that eventually - I just wanted to quickly unblock 4.06. Pull requests welcome (but we're trying to avoid bash in the build (and remove any remaining bash-isms) so we'd need to tweak that example a bit).

jbuild-workspace.dev
```diff
@@ -1,5 +0,0 @@
-;; This file is used by `make all-supported-ocaml-versions`
```
jordwalke: @rgrinberg We had to remove this because of [this problem](https://github.com/janestreet/jbuilder/issues/318) with `jbuilder`. Having a workspace.dev file makes it impossible to build sandboxed dependencies in a local cache if those dependencies build with jbuilder because they will climb the file system searching for a `jbuild-workspace(.*)` file.

src/reason-parser/reason_lexer.mll
hhugo: You don't need bash to replicate what's done in `ppx_ast`. It's just an ocaml program printing ocaml source code on stdout. 
I'll let you deal with people complaining about weird error messages ;)

jordwalke: Yeah we will refine this. Thanks for the suggestion.

Issue #1373, opened 2017-08-26
Callback-style printing


IwanKaramazow:
```
/* original */
describe("App", () => {
  test("math", () => {
    Expect.expect(1+2) |> toBe(3)
  }); 
});

/* becomes */
describe("App", () => test("math", () => Expect.expect(1 + 2) |> toBe(3)));
```

Printing based on line-length isn't very human-friendly in above kind of code.
Ideally we want to preserve the breaking here for readability.
This might be possible we if take the `Location.t` of every node into account.

jordwalke:
This just didn't make sense in the old syntax, so it never got any attention at the time. You can see why:

<img width="240" alt="screen shot 2017-11-13 at 2 42 36 pm" src="https://user-images.githubusercontent.com/977348/32753287-211d2bb8-c881-11e7-968b-095d5b30b1dc.png">

For the new syntax, we can just match a little deeper to see if the final function application argument is a lambda, and of so, break apart the lambda into `L => R`, and dock the lambdas `L`.

PR #1664, opened 2017-11-27
Callback printing

IwanKaramazow:
fixes https://github.com/facebook/reason/issues/1373

Example:
```reason
test("my test", () => {
  let x = a + b;
  let y = z + c;
  x + y
});

describe("App", () => {
  let x = 1;
  test("math", () => {
    let y = 1;
    Expect.expect(1 + 2) |> toBe(3)
  })
});

describe("App", () =>
  test("math", () =>
    Expect.expect(1 + 2) |> toBe(3)
  )
);

Thing.map(foo, bar, baz, (abc, z) =>
  MyModuleBlah.toList(argument)
);

Thing.map(
  foo,
  bar,
  baz,
  foo2,
  bakjlksjdf,
  okokokok,
  (abc, z) => {
    let x = 1;
    MyModuleBlah.toList(argument);
  }
);
```

### DIFF
--- a/formatTest/unit_tests/expected_output/syntax.re
+++ b/formatTest/unit_tests/expected_output/syntax.re
@@ -1133,7 +1133,23 @@ foo(
 /* Smooth formatting of functions with callbacks as arguments */
 funWithCb("text", () => doStuff());
 
+funWithCb(~text="text", ~f=() => doStuff());
+
+funWithCb(~text="text", ~f=?() => doStuff());
+
 test("my test", () => {
+  let x = a + b;
+  let y = z + c;
+  x + y
+});
+
+test(~desc="my test", ~f=() => {
+  let x = a + b;
+  let y = z + c;
+  x + y
+});
+
+test(~desc=?"my test", ~f=?() => {
   let x = a + b;
   let y = z + c;
   x + y
@@ -1145,7 +1161,27 @@ describe("App", () =>
   )
 );
 
+describe(~text="App", ~f=() =>
+  test(~text="math", ~f=() =>
+    Expect.expect(1 + 2) |> toBe(3)
+  )
+);
+
+describe(~text=?"App", ~f=?() =>
+  test(~text=?"math", ~f=?() =>
+    Expect.expect(1 + 2) |> toBe(3)
+  )
+);
+
 Thing.map(foo, bar, baz, (abc, z) =>
+  MyModuleBlah.toList(argument)
+);
+
+Thing.map(~a=foo, ~b=bar, ~c=?baz, ~f=(abc, z) =>
+  MyModuleBlah.toList(argument)
+);
+
+Thing.map(~a=foo, ~b=bar, ~c=?baz, ~f=?(abc, z) =>
   MyModuleBlah.toList(argument)
 );
 
@@ -1160,4 +1196,38 @@ Thing.map(
     let x = 1;
     MyModuleBlah.toList(x, argument);
   }
+);
+
+Thing.map(
+  ~a=foo,
+  ~b=bar,
+  ~c=baz,
+  ~d=foo2,
+  ~e=bakjlksjdf,
+  ~f=okokokok,
+  ~cb=(abc, z) => {
+    let x = 1;
+    MyModuleBlah.toList(x, argument);
+  }
+);
+
+Thing.map(
+  ~a=?foo,
+  ~b=?bar,
+  ~c=?baz,
+  ~d=?foo2,
+  ~e=?bakjlksjdf,
+  ~f=?okokokok,
+  ~cb=?(abc, z) => {
+    let x = 1;
+    MyModuleBlah.toList(x, argument);
+  }
+);
+
+Thing.map(
+  foo,
+  bar,
+  baz,
+  (abc, z) => MyModuleBlah.toList(argument),
+  (abc, z) => MyModuleBlah.toList(argument)
 );
--- a/formatTest/unit_tests/expected_output/syntax.re
+++ b/formatTest/unit_tests/expected_output/syntax.re
@@ -1129,3 +1129,35 @@ foo(
   ~y=[@foo] -0.1n,
   ~z=[@foo] -0.1p
 );
+
+/* Smooth formatting of functions with callbacks as arguments */
+funWithCb("text", () => doStuff());
+
+test("my test", () => {
+  let x = a + b;
+  let y = z + c;
+  x + y
+});
+
+describe("App", () =>
+  test("math", () =>
+    Expect.expect(1 + 2) |> toBe(3)
+  )
+);
+
+Thing.map(foo, bar, baz, (abc, z) =>
+  MyModuleBlah.toList(argument)
+);
+
+Thing.map(
+  foo,
+  bar,
+  baz,
+  foo2,
+  bakjlksjdf,
+  okokokok,
+  (abc, z) => {
+    let x = 1;
+    MyModuleBlah.toList(x, argument);
+  }
+);
--- a/formatTest/unit_tests/expected_output/wrappingTest.re
+++ b/formatTest/unit_tests/expected_output/wrappingTest.re
@@ -766,7 +766,8 @@ echoTheEchoer(
       o,
       p
     )
-  ) => (
+  ) =>
+  (
     a,
     b,
     c,
--- a/src/reason-parser/reason_heuristics.ml
+++ b/src/reason-parser/reason_heuristics.ml
@@ -1,4 +1,4 @@
-let is_punned_labelled_expression e lbl = 
+let is_punned_labelled_expression e lbl =
   let open Ast_404.Parsetree in
   match e.pexp_desc with
   | Pexp_ident { txt; _ }
@@ -7,7 +7,14 @@ let is_punned_labelled_expression e lbl =
     -> txt = Longident.parse lbl
   | _ -> false
 
-let estimateLineLengthForCallback ~args ~funExpr () =
+(* We manually check the length of `Thing.map(foo, bar, baz`,
+ * in `Thing.map(foo, bar, baz, (a) => doStuff(a))`
+ * because Easyformat doesn't have a hook to change printing when a list breaks
+ *
+ * we check if all arguments aside from the final one are either strings or identifiers,
+ * where the sum of the string contents and identifier names are less than the print width
+ *)
+let funAppCallbackExceedsWidth ~printWidth ~args ~funExpr () =
   let open Ast_404.Parsetree in
   let open Ast_404.Asttypes in
   let funLen = begin match funExpr.pexp_desc with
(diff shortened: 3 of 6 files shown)


formatTest/unit_tests/expected_output/syntax.re
```diff
@@ -1129,3 +1129,25 @@ foo(
...
+  test("math", () =>
+    Expect.expect(1 + 2) |> toBe(3)));
+
+List.map((aankomst) =>
+  renderUitslagenAankomst(
+    ~aankomst,
+    ~leden,
+    ~updateAankomst,
+    ()
+  ), Aankomsten.toList(aankomsten));
```
jordwalke: I like how Prettier formats callbacks. Here's an example:

```javascript
List.map(foo, bar, baz, (abc, z) =>
  MyModuleBlah.toList(argument)
);

List.map(
  (abc, z) => MyModuleBlah.toList(argument),
  foo,
  bar,
  baz
);
```

I notice it does two things that are worth emulating:
1. It only "docks" the callback if it is the last arg.
2. The final closing paren is aligned with the opening line (in refmt implementation that would be parens formatted with `~inline:(true, false)` I believe.

src/reason-parser/reason_pprint_ast.ml
```diff
@@ -2570,6 +2570,22 @@ let formatComputedInfixChain infixChainList =
   print [] [] "" infixChainList
 
 
+let rec injectClosingRParen = function
```
jordwalke: I think if we match Prettier's style, nothing special would be needed for injecting the closing paren, right?

jordwalke:
```
Thing.map(foo, bar, baz, (abc, z) =>
  MyModuleBlah.toList(argument)
)

```

I think the Prettier style could be accomplished by detecting if the final
argument to a function application is a lambda, and if so, special casing the
formatting as follows:

```
<Label>
<left>Thing.map(foo, bar, baz, (abc, z)</left><right>=>
  MyModuleBlah.toList(argument)
)</right>
</Label>
```

Then the "right" a list with wrap:("=>", ")") break:IfNeed and a single element in the list
(the contents of right hand side of arrow):

```
Thing.map(foo, bar, baz, (abc, z) <List>=>
  MyModuleBlah.toList(argument)
)</List>
```

Then the `<left>` part of the outermost label could itself be a label like:


```
<Label><left>Thing.map</left></right>(foo, bar, baz, (abc, z) </right></Label>=>
  MyModuleBlah.toList(argument)
)
```

Where the `<right>` part of *that* label could be a `<List>` with `wrap:("(", "
") break:IfNeed inline:(true, true)` with items: "foo", "bar", "baz", "(abc, z)", separated by commas.
Notice the empty space as second part of `wrap:` and notice `inline:(true, true)` to
make sure that the final `" "` space doesn't ever go onto its own newline.

```
Thing.map<List>(foo, bar, baz, (abc, z) </List>=>
  MyModuleBlah.toList(argument)
)
```

I think that would make sure that if needed, it would break like:

```
Thing.map<List>(
  foo,
  bar,
  baz,
  (abc, z) </List>=>
    MyModuleBlah.toList(argument)
)
```
which would always have the proper space before the `=>`.

Do you think this would work?

jordwalke:
@IwanKaramazow Pointed out that using a `<List>` with `wrap:("=>", ")")`, has one problem:

The following "bracey" expression on RHS of arrow:

```
<Label>
<left>Thing.map(foo, bar, baz, (abc, z)</left><right>=>
  {

  }
)</right>
</Label>
```

Formats as:

```
<Label>
<left>Thing.map(foo, bar, baz, (abc, z)</left><right>=> {

})</right>
</Label>
```

jordwalke:
Looks like it's getting closer!

IwanKaramazow:
I'm also looking into getting the following formatting 
```
List.map(
  item => doStuff(item.x, item.y, item.z, item.p),
  list
);

List.map(render => {
  let x = 1;
  x;
}, renderList);
```
Prettier seems to have super smooth inline printing when the expression returned by the callback has braces. I'm wondering if we should do something similar for when the expression isn't braced:
```
List.map(item =>
  doStuff(item.x, item.y, item.z, item.p),
  list
);
```

jordwalke:
Honestly if supporting first-item-is-callback ends up introducing complexity in the printer, I think we could live without it. I'm hoping that we can greatly simplify and refactor the printer and I think the main value of callback formatting is when the final arg is a callback.

formatTest/unit_tests/expected_output/syntax.re
```diff
@@ -1129,3 +1129,67 @@ foo(
...
+/* Smooth formatting of functions with callbacks as arguments */
+test("my test", () => doStuff());
+
+test("my test", () => {
+  let x = a + b;
+  let y = z + c;
+  x + y;
+});
+
+describe("App", () =>
```
jordwalke: This looks great.

formatTest/unit_tests/expected_output/syntax.re
```diff
@@ -1129,3 +1129,67 @@ foo(
...
+);
+
+Thing.map(
+  foo,
+  bar,
+  baz,
+  foo2,
+  bakjlksjdf,
+  okokokok,
+  (abc, z) =>
```
jordwalke: Is this the only remaining thing that could be improved?

IwanKaramazow: It is

IwanKaramazow:
Ok, managed the print everything as smooth as Prettier after introducing a heuristic suggested by @jordwalke. When manually calculating the line length upfront (func identifier + length of args), we know if the whole thing is gonna break or not. Based on that we select a different printing style...

This will need to be refactored once we're able to push this kind of logic down into our formatting algorithm.

chenglou:
Awesome. This format can be better:

```reason
Thing.map(foo, bar, baz, (abc, z) => MyModuleBlah.toList(argument), (abc, z) =>
  MyModuleBlah.toList(argument)
);
```

Maybe print it like this?

```reason
Thing.map(
  foo, 
  bar, 
  baz, 
  (abc, z) => MyModuleBlah.toList(argument), 
  (abc, z) => MyModuleBlah.toList(argument)
);
```

Rationale: I'd print it like so even if it was short:

```reason
Thing.map(
  (abc, z) => z,
  (abc, z) => z
);
```

Callback in the front seems to print well (except for the above case as well: `Thing.map((abc, z) => z, (abc, z) => z, 1);`)

chenglou:
Here's another. This is the new one:

```reason
Thing.map(foo, bar, baz, (abc, z) =>
  MyModuleBlah.toList(argumentaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa)
);
```

Maybe I was expecting:

```reason
Thing.map(
  foo,
  bar,
  baz,
  (abc, z) => MyModuleBlah.toList(argumentaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa)
);
```

Not sure. Also, we don't seem to automatically break when we see a callback, so the following looks like this:

```reason
let jsComponent = ReasonReact.wrapReasonForJs(~component, (jsProps: jsProps) => make());
```

When I think I'm expected either this:

```reason
let jsComponent = ReasonReact.wrapReasonForJs(
  ~component, 
  (jsProps: jsProps) => make()
);
```

Or this:

```reason
let jsComponent = ReasonReact.wrapReasonForJs(~component, (jsProps: jsProps) => 
  make()
);
```

jordwalke:
I think we could/should refine the printing of this one that you found:

```reason
Thing.map(foo, bar, baz, (abc, z) => MyModuleBlah.toList(argument), (abc, z) =>
  MyModuleBlah.toList(argument)
);
```

The other two that you found are exactly how Prettier prints them and I'm not really opposed to them. I don't think we *always* want to break when there's a callback because I'd expect to be able to do things like the following in a lightweight manner (which is probably why Prettier doesn't force a break on every callback).

```reason
let x = Lst.map(myList, x => x + 1);
```

src/reason-parser/reason_pprint_ast.ml
```diff
@@ -6640,6 +6644,159 @@ class printer  ()= object(self:'self)
...
+       lost *)
+    let maybeJSXAttr = List.map self#attribute jsxAttrs in
+
+    let categorizeFunApplArgs args =
+      let reverseArgs = List.rev args in
+      match reverseArgs with
+      | (Nolabel, {pexp_desc = Pexp_fun _})::_ -> `LastArgIsCallback reverseArgs
+      | _ -> `NormalFunAppl args
+    in
+    begin match categorizeFunApplArgs args with
```
jordwalke: This printer is getting pretty long. Can we move this to another module? Something like "style_classifier.ml". We can probably move a bunch of stuff there. It would make Merlin more responsive while editing the pretty printer.

formatTest/unit_tests/expected_output/syntax.re
```diff
@@ -1129,3 +1129,35 @@ foo(
   ~y=[@foo] -0.1n,
   ~z=[@foo] -0.1p
 );
+
```
jordwalke: Worth adding Cheng's finding to the test cases to track how they format:

```
Thing.map(foo, bar, baz, (abc, z) => MyModuleBlah.toList(argument), (abc, z) =>
  MyModuleBlah.toList(argument)
);
```

src/reason-parser/reason_pprint_ast.ml
```diff
@@ -6640,6 +6644,159 @@ class printer  ()= object(self:'self)
...
     | [x] -> x
     | xs -> makeBreakableList xs
+
+  method formatFunAppl ~jsxAttrs ~args ~funExpr () =
+    (* If there was a JSX attribute BUT JSX component wasn't detected,
+       that JSX attribute needs to be pretty printed so it doesn't get
+       lost *)
+    let maybeJSXAttr = List.map self#attribute jsxAttrs in
+
+    let categorizeFunApplArgs args =
```
jordwalke: To address @chenglou's finding, I propose that we change this to detect when the final argument is a callback and no other argument is a callback.

src/reason-parser/reason_pprint_ast.ml
```diff
@@ -6640,6 +6644,159 @@ class printer  ()= object(self:'self)
...
+      | (Nolabel, {pexp_desc = Pexp_fun _})::_ -> `LastArgIsCallback reverseArgs
+      | _ -> `NormalFunAppl args
+    in
+    begin match categorizeFunApplArgs args with
+    | `LastArgIsCallback reverseArgs ->
+        (* This is the following case:
+         * Thing.map(foo, bar, baz, (abc, z) =>
+         *   MyModuleBlah.toList(argument)
+         *)
+        let (args, callbackArg) = begin match reverseArgs with
```
jordwalke: If you store the last arg and remaining args at the time you create`LastArgIsCallback`, then you already have them split and can remove this `raise`.

src/reason-parser/reason_pprint_ast.ml
```diff
@@ -6640,6 +6644,159 @@ class printer  ()= object(self:'self)
...
+        | callback::args -> (List.rev args, callback)
+        | [] -> raise (NotPossible "can't partition 0 function expression args")
+        end in
+        (* callback isn't a labelled argument *)
+        let (_, cb) = callbackArg in
+        let (_sweet, cbArgs, retCb) = self#curriedPatternsAndReturnVal cb in
+
+        (* major hack; we manually check the length of `Thing.map(foo, bar,baz`,
+         * because Easyformat doesn't have a hook to change printing when a list breaks *)
+        let estimatedLineLength =
```
jordwalke: This might be a good candidate for placing in `heuristics.ml` or something.

src/reason-parser/reason_pprint_ast.ml
```diff
@@ -6640,6 +6644,159 @@ class printer  ()= object(self:'self)
...
+               let identList = Longident.flatten ident.txt in
+               let lengthOfDots = List.length identList - 1 in
+               let len = List.fold_left (fun acc curr ->
+                 acc + (String.length curr)) lengthOfDots identList in
+               len
+           | _ -> 0
+          end in
+          let argsLength = List.fold_left (fun acc curr ->
+            match acc with
+            | None -> None
```
jordwalke: Is it cleaner/simpler to just use an integer where `0` functions as `None`?

IwanKaramazow:
@chenglou I addressed your comments.

I don't follow 100% on 
```reason
Thing.map(
  (abc, z) => z,
  (abc, z) => z
);

Thing.map(
  (abc, z) => z,
  (abc, z) => z,
  1
);
```
You want them to always break because there are two function expressions as arguments?

formatTest/unit_tests/expected_output/wrappingTest.re
```diff
@@ -766,7 +766,8 @@ echoTheEchoer(
       o,
       p
     )
-  ) => (
+  ) =>
```
IwanKaramazow: Collateral damage of Easy_format's limitations, will be addressed in the future.

jordwalke:
> You want them to always break because there are two function expressions as arguments?

My understanding was that at least we shouldn't try to do any final-argument docking when > 1 argument is a callback. I'm not sure if we should force break every time, but at least don't do any docking.

src/reason-parser/reason_heuristics.ml
```diff
@@ -0,0 +1,64 @@
...
+       let identList = Longident.flatten ident.txt in
+       let lengthOfDots = List.length identList - 1 in
+       let len = List.fold_left (fun acc curr ->
+         acc + (String.length curr)) lengthOfDots identList in
+       len
+    | _ -> -1
+  end in
+  (* eats an argument & substract its length from the printWidth
+   * as soon as the print width reaches a sub-zero value,
+   * we know the print width is exceeded & returns *)
```
jordwalke: nice.

src/reason-parser/reason_pprint_ast.ml
```diff
@@ -6649,6 +6652,128 @@ class printer  ()= object(self:'self)
...
+       lost *)
+    let maybeJSXAttr = List.map self#attribute jsxAttrs in
+
+    let categorizeFunApplArgs args =
+      let reverseArgs = List.rev args in
+      match reverseArgs with
+      | ((_, {pexp_desc = Pexp_fun _}) as callback)::args 
+          when let otherCallbacks = 
+            List.filter (fun (_, e) -> match e.pexp_desc with Pexp_fun _ -> true | _ -> false) args
+          in List.length otherCallbacks == 0
```
jordwalke: cool I think this takes care of Cheng's request.

src/reason-parser/reason_pprint_ast.ml
```diff
@@ -6649,6 +6652,128 @@ class printer  ()= object(self:'self)
...
+           *  describe("App", () =>
+           *    test("math", () =>
+           *      Expect.expect(1 + 2) |> toBe(3)
+           *    )
+           *  );
+           *)
+          let forceBreak = match funExpr.pexp_desc with
+          | Pexp_ident ident when
+              let lastIdent = Longident.last ident.txt in
+              List.mem lastIdent ["test"; "describe"; "it"; "expect"] -> true
```
jordwalke: okay, nice touch. No one will complain.

PR #1665, opened 2017-11-28
More liberal menhir version constraint.

jordwalke:
The [opam release](https://github.com/ocaml/opam-repository/pull/10869)
was failing in a way that seemed related to Menhir. Only on linux though
- and it wasn't being caught in our own CI - only OPAM's. Must be for
some configuration we don't test. Making menhir dep more liberal to see
if it will fix our attempted release.

Creating pull request to see if CI passes.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,4 @@
 *.messages binary
-*.lock binary
 
 docs/highlightJs/** binary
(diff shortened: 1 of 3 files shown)


jordwalke:
Merging and will try submitting to opam.

PR #1670, opened 2017-11-29
Fixes #1668

### LINKED ISSUE #1668: Super ugly printing of variant constructor with attribute
Repro:

```reason
type fruit =
  | Apple
  | [@bs.as 10] Orange
  | [@bs.as 100] Kiwi
  | Watermelon;
```

prints as:

```reason
type fruit =
  | Apple
  [@bs.as 10] | Orange
  [@bs.as 100] | Kiwi
  | Watermelon;
```

chenglou:
Took the occasion to clean up some logic

### DIFF
--- a/src/reason-parser/reason_pprint_ast.ml
+++ b/src/reason-parser/reason_pprint_ast.ml
@@ -2859,12 +2859,6 @@ class printer  ()= object(self:'self)
   method type_variant_leaf1 opt_ampersand polymorphic print_bar x =
     let {pcd_name; pcd_args; pcd_res; pcd_loc; pcd_attributes} = x in
     let {stdAttrs} = partitionAttributes pcd_attributes in
-    let prefix = if polymorphic then "`" else "" in
-    let sourceMappedName = SourceMap (pcd_name.loc, atom (prefix ^ pcd_name.txt)) in
-    let nameOf = makeList ~postSpace:true [sourceMappedName] in
-    let barName =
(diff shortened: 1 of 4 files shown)


src/reason-parser/reason_pprint_ast.ml
```diff
@@ -2859,12 +2859,6 @@ class printer  ()= object(self:'self)
   method type_variant_leaf1 opt_ampersand polymorphic print_bar x =
     let {pcd_name; pcd_args; pcd_res; pcd_loc; pcd_attributes} = x in
     let {stdAttrs} = partitionAttributes pcd_attributes in
-    let prefix = if polymorphic then "`" else "" in
-    let sourceMappedName = SourceMap (pcd_name.loc, atom (prefix ^ pcd_name.txt)) in
-    let nameOf = makeList ~postSpace:true [sourceMappedName] in
-    let barName =
```
chenglou: This logic is a duplicate from the `makeList` in `add_bar` below. Unified both into a new `add_bar`. Named the old `add_bar` into `add_bar_2`

jordwalke:
Cool, let's just make sure the tests pass on all versions, then merge away!

IwanKaramazow:
Can you add a test case with a gadt?

chenglou:
Argh good point. And I thought they Just Worked.

Doesn't work for GADT and extensible variants, as parsing `| [@a] B` fails for these two. This should be addressed in a different issue though

PR #1675, opened 2017-12-01
change print-width default from 100 to 80

kennetpostigo:
This PR will change the default print-width of refmt to be 80 columns. This matches prettiers print-width default. 

cc @IwanKaramazow


IwanKaramazow:
I'm ok with this change. @chenglou what do you think?

chenglou:
Yeah let's merge this

PR #1688, opened 2017-12-08
Oprint: print `Js.t({. foo: bar})` as `{. "foo": bar}`

### LINKED ISSUE #1682: Outcome printer doesn't print `Js.t({. foo: int})` as `{. "foo": int}`

chenglou:
Just like for our regular printer. Fixes #1682


chenglou:
Seems good

PR #1692, opened 2017-12-09
Make identifier lambdas parenless

jordwalke:
This partially addresses the issue that @bobzhang and @glennsl brought up about the new syntax having too many parens. I tried formatting a large project and noticed that the most jarring use of parens was:

- Single identifiers in lambdas that had parens included.
- Records arguments to lambdas that had parens included.

We can't really remove the parens around record arguments to lambdas at the moment because then it opens up this precedent which we cannot uphold for tuples.
But still, single identifier lambda arguments were the biggest offender of paren-fatigue in my example project.

cc @IwanKaramazow @chenglou @rickyvetter

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/jsx.re
+++ b/formatTest/typeCheckedTests/expected_output/jsx.re
@@ -205,7 +205,7 @@ let selfClosing3 =
     b="cause the entire thing to wrap"
   />;
 
-let a = <Foo> <Bar c=((a) => a + 2) /> </Foo>;
+let a = <Foo> <Bar c=(a => a + 2) /> </Foo>;
 
 let a3 = <So> <Much> <Nesting /> </Much> </So>;
 
@@ -268,7 +268,7 @@ let jsxInList7 = [<Foo />, <Foo />];
 
 let jsxInList8 = [<Foo />, <Foo />];
 
-let testFunc = (b) => b;
+let testFunc = b => b;
 
 let jsxInFnCall = testFunc(<Foo />);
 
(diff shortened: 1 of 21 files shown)


formatTest/unit_tests/expected_output/functionInfix.re
```diff
@@ -19,42 +19,42 @@ fff >>= xx(yy) >>= aa(bb);
 fff >>= (xx(yy) >>= aa(bb));
```
jordwalke: @glennsl This is the kind of stuff this diff cleans up.

formatTest/typeCheckedTests/expected_output/jsx.re
```diff
@@ -205,7 +205,7 @@ let selfClosing3 =
     b="cause the entire thing to wrap"
   />;
```
jordwalke: And jsx callback arguments.

formatTest/unit_tests/expected_output/infix.re
```diff
@@ -1084,7 +1084,7 @@ let server = {
         )
     )
     >>= (
-      (body) =>
+      body =>
```
IwanKaramazow: Hmm forgot this case in the callback PR.
The `body =>` should dock on the same line as `>>= (`, shouldn't it?

jordwalke: I'm not sure - printing these have bigger issues - like knowing when exactly we can omit the parens. We're fairly certain that determining when the () can be omitted is such a hard problem that it requires integration with the parser.

PR #1703, opened 2017-12-13
Pretty print extensions

let-def:
With current version, Reason accepts:
```
%lwt
let x = Lwt.return(5);
```

But pretty print it as:
```
[%%lwt let x = Lwt.return(5)];
```

This PR implements pretty-printing for extension sugars (in structures and in expressions).

### DIFF
--- a/formatTest/unit_tests/expected_output/extensions.re
+++ b/formatTest/unit_tests/expected_output/extensions.re
@@ -22,3 +22,34 @@ let x = {
   %extend
   return("hi");
 };
+
+let x = {
+  %extend
+  if (true) {1} else {2};
+  %extend
+  switch None {
+  | Some(x) => assert false
+  | None => ()
+  };
+  %extend
+  try (raise(Not_found)) {
+  | Not_found => ()
+  | Invalid_argument(msg) => prerr_endline(msg)
+  };
+};
+
+let x = [%extend if (true) {1} else {2}];
+
+let x = [%extend
+  switch None {
+  | Some(x) => assert false
+  | None => ()
+  }
+];
+
+let x = [%extend
+  try (raise(Not_found)) {
+  | Not_found => ()
+  | Invalid_argument(msg) => prerr_endline(msg)
+  }
+];
--- a/formatTest/unit_tests/expected_output/extensions.re
+++ b/formatTest/unit_tests/expected_output/extensions.re
@@ -8,48 +8,40 @@ module M = {};
 %extend
 module type M = {};
 
-%extend
-let x = "hi";
+let%extend x = "hi";
 
 let x = {
-  %extend
-  let x = ();
+  let%extend x = ();
   ignore();
   %extend
   ignore();
-  %extend
-  let x = ();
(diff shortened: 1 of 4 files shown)


hcarty:
Very cool!  Does this work for `match`, `if` and `try`?  Those are all used/supported by Lwt and `match` and `if` are supported by ppx\_let.

let-def:
Yes, these are supported, I added them to the test.

However, the sugar triggers only in "sequences" (between `{`/`}`, so either with a `...; ...` or `let ... = ...; ...`).

So when there is only one expression, it falls back to `[%lwt ...]`.
Compare:
- https://github.com/facebook/reason/pull/1703/commits/4816f921a2230eb5cfe99fb17ec934c10178792c#diff-ffb9c2b127d7d439f65f425eda9d3a28R58
and
- https://github.com/facebook/reason/pull/1703/commits/4816f921a2230eb5cfe99fb17ec934c10178792c#diff-b7c46d6c2761df27d2ef15c37fe786e1R50

let-def:
So the question remains what should be the printing for a single expression:

```reason
let x = {
  %extend
  switch None {
    | Some(x) => assert(false)
    | None => ()
  };
};
```

```reason
let x = [%extend
  switch None {
    | Some(x) => assert(false)
    | None => ()
  };
];
```


```reason
let x = (%extend
  switch None {
    | Some(x) => assert(false)
    | None => ()
  };
);
```

```reason
let x = %extend
  switch None {
    | Some(x) => assert(false)
    | None => ()
  };
```

(I am not sure all cases are unambiguous, I can investigate patching the parser).

chenglou:
Awesome! Though can we do like ocaml and have `let%lwt`, etc?

let-def:
@chenglou: I agree it would be nice, I will investigate!

jordwalke:
The convenient thing about OCaml's `let%ppx` form is that it conveys that the whole `let` is a ppx extension - completely replacing an AST node, as opposed to a "decorator" attached to another otherwise normal AST node.

For that Reason, I'm thinking that the prefix syntax you have here would make perfect sense for `@` attributes:

```
@foo
let anotherThing = bar;
```

Even if ppx extensions use the more "embedded" syntax:

```
let%foo = bar;
```

For single expressions, the same lean bracketless ppx attributes would also be awesome:

```
let something = @attr myFunc();
```

let-def:
@chenglou: I agree it would be nice, I will investigate!

let-def:
Ok, the last update add supports for `fun%extend, function%extend, let%extend, switch%extend, try%extend, while%extend, for%extend` (I hope I didn't forget any :P?).

Plus non negligible simplification to the parser (more to come).

hcarty:
@let-def Looks like it covers `if%extend` too, correct?

This is really cool! ocamlformat just got extension point sugar support too so pleasantly formatted extension points for everyone!

let-def:
@hcarty indeed, thanks I forgot to list that one :).

hcarty:
This looks amazing - I can't wait to try it out!

PR #1704, opened 2017-12-13
Don't install menhir_error_processor

let-def:
This tool is used during compilation and should not be installed.


jordwalke:
Thank you!

Issue #1364, opened 2017-08-19
`let f thing => thing##x` doesn't format to ocaml correctly


chenglou:
It turns into `let f thing = ## thing x` in ocaml

hcarty:
I'm not sure how refmt uses pprintast, but if it's inlined rather than pulled from compiler-libs the latest git version should have a fix.

jaredly:
this is an ocaml printer problem :/

chenglou:
We'll solve it in around 3 months

PR #1708, opened 2017-12-14
Fix wrong ## and #= output by refmt

wkwkes:
Fixes [#1364](https://github.com/facebook/reason/issues/1364)

Refmt converts `##` and `#=` in Reason incorrectly.
For example, `let f(thing) = thing##x#=42` becomes `let f thing = #= (## thing x) 42`.

Refmt looks rely on OCaml for printing the parsed AST [(here).](https://github.com/facebook/reason/blob/master/src/reason-parser/reason_toolchain.ml#L459)
OCaml compiler, however, does not treat `##` and `#=` as infix operators.
So, `##` and `#=` are printed as normal functions.

To fix this problem, it seems that we have to modify `Pprintast` module.
I replaced `Pprintast` with `Reason_ml_pprintast` that was copied from `pprintast.ml` in OCaml compiler, adding `##` and `#=` to `special_infix_strings` in it.

This works fine: `let f thing = (thing ## x) #= 42`

Of course, copying `pprintast.ml` is not elegant, and I'm not sure this is the right solution.

### DIFF
--- a/src/reason-parser/jbuild
+++ b/src/reason-parser/jbuild
@@ -45,7 +45,8 @@
   (public_name reason)
   (wrapped false)
   (modules
-   (syntax_util
+   (reason_ml_pprintast
+    syntax_util
     reason_heuristics
     reason_toolchain
     reason_config
--- a/src/reason-parser/reason_ml_pprintast.ml
+++ b/src/reason-parser/reason_ml_pprintast.ml
@@ -1,0 +1,1446 @@
+(**************************************************************************)
+(*                                                                        *)
+(*                                 OCaml                                  *)
+(*                                                                        *)
+(*                      Thomas Gazagnaire, OCamlPro                       *)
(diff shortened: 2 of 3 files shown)


Octachron:
Beyond the problem of the duplication of the pprintast file (which is under a LGPL license), this is the wrong fix: `#` should be added to the list of infix symbols. There are no reason to special case `##` and `#=` since `#` is a valid prefix for operators. Note also that this bug was fixed upstream in OCaml 4.06.

wkwkes:
Thank you for your comment. I did not know this was fixed. 
This PR contains many problems. I close it.

chenglou:
@wkwkes Thanks for your PR though! It'll take a little longer (next year) until the web version of refmt gets upgraded to 4.06.

Keep the PRs coming though. Welcome =)

PR #1709, opened 2017-12-15
Better way to support OCaml 4.06 by using janestreet/ppx_ast/generate

### LINKED ISSUE #1608: (Less Hacky) OCaml 4.06.0 support

ncihnegn:
Follow the approach suggested by @hhugo in comments of #1663 
Hopefully this will close #1608


src/generate/select.ml
```diff
@@ -0,0 +1,35 @@
+let dump_file file =
```
jordwalke: If this unmodified from Jane Street's library, I think it needs to retain the copyright header, and have a mention in `ORIGINS.md`, specifying what license it was originally written under.

jordwalke:
I'm happy to see people cleaning up some of the less elegant pieces of the toolchain, thanks!

src/generate/select.ml
ncihnegn: It was slightly modified for better compatibility. There was no copyright header but I has added to ORIGINS.md.

jordwalke: Thank you.

src/reason-parser/jbuild
```diff
@@ -2,6 +2,12 @@
 
 (ocamllex (reason_lexer))
 
+; Use select.exe from janestreet/ppx_ast to choose lexer_warning.ml based on ocaml version
+(rule
+ ((targets (lexer_warning.ml))
+   (deps (../generate/select.exe lexer_warning.ml-4.06 lexer_warning.ml-default))
+   (action (bash "../generate/select.exe lexer_warning.ml-4.06 lexer_warning.ml-default > ${@}"))))
```
jordwalke: @rgrinberg Do we need to prefix this with `bash` here?

rgrinberg: Since it's doing shell redirection, it's necessary. But the the use of bash itself is not justified. This can be with the `with-stdoutout-to` stanza instead:

```
(with-stdout-to ${@} (run ../generate/select.exe lexer_warning.ml-4.06 lexer_warning.ml-default))
```

jordwalke: That's what I'm looking for! Thanks @rgrinberg!
@ncihnegn Care to try that and then we can merge?

jordwalke:
Thank you, again!

PR #1719, opened 2017-12-23
Fix PSig + various simplifications to the parser

let-def:
Note: this PR is on top of https://github.com/facebook/reason/pull/1703, it should be merged after.

Implements handling of PSig attributes payload (`[@attr: <signature>]`), which is provided by OCaml but not yet supported by Reason.

Then refactor the parser:
- some important simplifications: remove `only_core_type` trick, simpler treatment of `record`/`object` labels
- more cosmetic stuff: replace custom ast constructors (`mkexp`, `mkclass`, ...) by `Ast_helper`, replace `/*` comments by `(*` comments.

I can remove or adjust this last part if people don't feel comfortable with it (in particular it might conflict with other open PRs).
Using Ast_helper allows to reduce the number of definitions that one needs to know to work with the parser code. The current code has a lot of subtly different variants for constructing pieces of Ast (for instance with handling of ghost locations), this refactoring reduce the mental overhead a bit (however one still needs to know about Ast_helper) and provide at least as precise and sometime more precise locations.
Using `(*` make the comments more uniform (same rules at menhir level and at OCaml level, with the nice side effect of fixing the highlighting in vim).


jordwalke:
Your pretty printing of ppx extensions is merged, and your other diff which implements PSig is now also merged. Awesome work. Now this diff will look like less of an intimidating change on top of master.

let-def:
Rebased on current master (and I dropped the patch that rewrote code to make use of Ast_helper, remaining ones are more self-contained).

PR #1720, opened 2017-12-23
parser/pprint_ast: force parentheses around switch (...)/try (...)

let-def:
This PR force scrutinees in `switch` and `try` expressions to be wrapped with parentheses: `switch (a) { | None => ... }`.


chenglou:
Is it possible to only do the printing change for now? This way after folks all refmt their files we can release the parser change, which wouldn't break (many) things

jordwalke:
How does this change the parsing of switching on tuples?

let-def:
It doesn't change: `switch (a,b) { | (Some(one), Some(body)) => ... }`

PR #1724, opened 2017-12-28
Add missing Unix dep to menhir_error_processor

rgrinberg:
Should fix errors like this: https://travis-ci.org/janestreet/jbuilder/jobs/321452625#L790


jordwalke:
Thank you very much! We really appreciate these pull requests.

PR #1725, opened 2017-12-28
Move rebuild to own opam package

rgrinberg:
FIrst step in slimming down the dependency code for people who just want to build reason code.

### DIFF
--- a/src/rtop/jbuild
+++ b/src/rtop/jbuild
@@ -2,6 +2,7 @@
 
 (install
  ((section bin)
+  (package reason)
   (files (rtop_init.ml))))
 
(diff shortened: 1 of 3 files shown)


src/rtop/jbuild
```diff
@@ -2,6 +2,7 @@
 
 (install
  ((section bin)
+  (package reason)
```
jordwalke: Was this needed? Why?

rgrinberg: This is needed because when you have multiple packages in one opam repository, jbuilder needs to know which binary belongs to which package.

PR #1726, opened 2017-12-28
Remove reason.opam.in

rgrinberg:
There's no need to generate reason.opam and to check it in because none of the
generated information should be included in the opam file regardless.


rgrinberg:
I spent some time to make sure nothing is broken but I'm not 100% sure. Any breakage should be trivial to fix however. Since it's simply removing a step.

Issue #1728, opened 2017-12-30
Negative numbers print unnecessary parens


jordwalke:
```reason
(-10, -10)
```

Print as:

```reason
((-10), (-10))
```

We're on a hunt to reduce as many unnecessary parens as possible to avoid paren-fatigue.

aweary:
@jordwalke I see there's [some logic that explicitly wraps negative numbers in expressions for clarity](https://github.com/facebook/reason/blob/master/src/reason-parser/reason_pprint_ast.ml#L2376-L2392) like `1 + (-2) + 3`. I imagine you want to keep doing that right? Are there other cases where parens are desired/required?

jordwalke:
I don't mind it in that situation where you have infix `+`. However, for tuples, `(x, y)`, the `x` and `y` should be treated the same as if they were to the right side of an `=` sign in `let _ = x`. They are protected by their surrounding parens/commas.

aweary:
Is this issue specifically about removing the parens when the negative numbers are part of a tuple? I'm wondering how you want to handle the other cases where negative numbers are currently wrapped. AFAICT they're always wrapped in parens with the exception of constructor and function arguments.

jordwalke:
I think if we fixed whatever is causing the parens in tuples, it would also fix some other times they are wrapped in superfluous parens. I can't recall if there's any case where they need to be wrapped in parens right now in the new syntax.

PR #1731, opened 2018-01-06
Adding optional param to know when to wrap neg prefix symbol in parens

RobertPaul01:
#1728

Passing along this extra parameter should allow the printer to know when it's correct to wrap a negative prefix. I found that this would work by backtracking what functions are called for these different cases.


chenglou:
Thanks for the contribution!
Though this fix seems to special-case negative number printing a bit too much imo. Is there a way to pattern-match on a leaf node and print it specially, rather than carrying a labeled argument all the way through?

RobertPaul01:
Sure I can look more into it, I think pattern matching should be possible. Which cases are not wanted to be printed without the parentheses?

PR #1732, opened 2018-01-06
Split reason into more packages

rgrinberg:
* rtop - utop comment
* reason-merlin - merlin plugin


jordwalke:
I think that the merlin reason adapter should stay in the reason plugin. Why? Because otherwise people will have to specify two packages to develop Reason with merlin. They'll have to specify Reason, and the Reason Merlin adapter. Isn't the Reason Merlin adapter so small that it should be okay to have it included in Reason?

rgrinberg:
I get the concern but I'm not sure if keeping the two components in one package is the correct way to address it. Keeping the merlin component inside is quite heavy for users who just want to run to compile/publish some reason code on opam.

I think the split that I've proposed here still works best, but we could also do something a bit more complicated with virtual packages. For example, we can have `refmt`, `reason-merlin`, etc. packages and a `reason` virtual package that still depends on all of those.

kennetpostigo:
Not sure if worth-while or possible, but couldn't the inverse be true to? Couldn't we have something like `reason-compiler` that will compile only compile reason code?

rgrinberg:
> Couldn't we have something like reason-compiler that will only compile reason code?

Isn't this the same as my proposed `refmt` package?

jordwalke:
This would require people updating their project workflow, and all the docs just to support editing in merlin. I'm okay with having people go back end edit their workflow/installs to include a REPL, especially since utop is so huge - I'm not sure it's worth it for this relatively small dependency primarily because it's so unlikely to cause any package version conflicts (Reason is the only one that depends on it).

rgrinberg:
I see. I mistakenly thought that merlin-extend pulls in all of merlin. If it's so small, then I'm not sure why don't you just vendor it, or even just develop it in the reason repo. Don't see how this situation is different from easy-format for example.

jordwalke:
I'm not sure there's much of a difference between vendoring and not vendoring for such a small library that is unlikely to cause package manager conflicts.

rgrinberg:
Easy format is just as small. I don't think there's a difference either.
But it doesn't seem like the policy is applied consistently

jordwalke:
One difference is that Easy Format has many more packages that depend on it so it could increase the probability of package conflicts by having it as a proper dependency instead of vendored. The merlin-extend dependency is only depended on by reason.
The difference isn't very large in this case because even easy format only has a handful of dependencies.

Here's one benefit to having merlin-extend, (and even easy format) as external deps. If I upgrade Reason, merlin-extend is in my global package build cache and so it's quicker to only recompile Reason and not the vendored dependencies.

rgrinberg:
> Here's one benefit to having merlin-extend, (and even easy format) as external deps. If I upgrade Reason, merlin-extend is in my global package build cache and so it's quicker to only recompile Reason and not the vendored dependencies.

I don't think that matter as they're still separate opam packages. Esy does the caching on the opam package level, right?

Anyway, I still remained unconvinced by the explanation and puzzled about how the vendoring decision is made. But since I agree that the impact is minimal either way, I just got rid of the reason-merlin package. I'll re-raise this issue if it ever causes a problem for me.

jordwalke:
Thanks for your patience @rgrinberg. I'm working through pull requests now. The tests on this one seem to be failing. I think the PR as is should be merged except for those failures.

jordwalke:
The Makefile has a call to `./miscTests/rtopIntegrationTest.sh` that is run for Reason. That might be failing. I'll also have to create a publishing workflow that will let us release either of these packages.

jordwalke:
Due to the failing tests, I think I'll just have to manually reapply your change and figure out what's going on. Hope that's okay, and thanks for your patience. The next opam release will not have rtop or rebuild within it.

jordwalke:
I took this and fixed the failing test cases in [this diff](https://github.com/facebook/reason/pull/1790/files).

Thanks for sending this PR! I'll close it in favor of the updated version I sent.

PR #1733, opened 2018-01-06
Temporarily revert #1720

chenglou:
See https://github.com/facebook/reason/pull/1720#issuecomment-353732887

This way of printing with parens while parsing both lets userland
gradually use the enforced parens syntax without breaking their app
(especially, and mainly, if they have third-party dependencies).

We'll re-revert this patch in the future.


chenglou:
cc @let-def.

Feel free to re-revert this diff (aka basically build on top of 1720) for your experiment on the percent-less extension point syntax

let-def:
Thanks! :)

PR #1734, opened 2018-01-06
Fix PSig

let-def:
Just the PSig part of https://github.com/facebook/reason/pull/1719.

Add support for signatures in attribute payloads and update the printing of attributes (no more space between the attribute identifier and the token used to distinguish signatures and pattern cases).


jordwalke:
Thank you, @let-def ! Merging!

PR #1735, opened 2018-01-07
Percent less

let-def:
Allow extension sugar without having to put `%`:

```reason
let async f = ...
/* as well as */
switch async / try async / for async / if async / ...
```

The sugar applies for any lowercase identifier that is not a keyword.

The `let f() = ...` / `let f() { ... }` syntax had to be removed (functions are now always bound like any other expressions, `let f = () => ...`).

### DIFF
--- a/formatTest/typeCheckedTests/input/attributes.re
+++ b/formatTest/typeCheckedTests/input/attributes.re
@@ -87,7 +87,7 @@ let x = [@attrEverything] (true && false);
 /**
  * How attribute parsings respond to other syntactic constructs.
  */
-let add(a) { [@onRet] a };
+let add = (a) => { [@onRet] a };
 let add = fun(a) => [@onRet] a;
 let add = [@onEntireFunction] (fun(a) => a);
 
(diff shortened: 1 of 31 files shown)


hcarty:
OCaml supports `;%extend` - is there a way to support a Reason equivalent with this syntax?  It would be useful for operations used purely for side effects like printing/logging.

jordwalke:
It's a very nice surprise that Reason's syntax enabled us to omit the `%` without any ambiguities. This looks really approachable.

A couple of thoughts: We would probably want to start with make omitting the `%` opt-in, just in case we for some reason need to use a different solution in the next release. In that way people could try it out that it's subject to change.
We also might want to only add a few whitelisted ppx names to be printed with the `%` omitted because it might look confusing for other ppx names.

For some reason it's failing on Mac OS only with " Error: 8 states have shift/reduce conflicts."

let-def:
The conflicts are due to the patch forcing parentheses around `switch`/`try` scrutinee being reverted.

ELLIOTTCABLE:
Wow. This would *really* be the killer feature of the Reason syntax, IMO.

Well, assuming the BuckleScript team relaxes its position on AST modification. 🤣

jaredly:
ummm I have strong reservations about elevating ppx extension points to this level of "prominence" in the syntax. I believe it should be fairly "loud" when a syntax transform is happening.
@chenglou and I have talked about an alternate proposal that would allow monads to be handled gracefully without relying on ppx's (basically through first-class CPS support).
it would involve introducing a `let!foo x = y; ...etc` form that would be translated to `foo(y, x => ...etc)`
This allows us to handle async, optionals, results, etc, while *also* allowing you to cmd-click into `foo` to see what's happening. It feels much less magic than ppxs, which I like a lot.

hcarty:
@jaredly There are a few relevant active PRs upstream which may be of interest and worth chiming in on to make sure the Reason community's needs are taken into account.

Monadic `let` operators - https://github.com/ocaml/ocaml/pull/1947

`match` operators - https://github.com/ocaml/ocaml/pull/1955

jaredly:
@hcarty wow, that's a very cool proposal!

PR #1739, opened 2018-01-09
Explicit long identifiers on parser

haskellcamargo:
According to a `TODO` comment on parser, in order to achieve best maintainability, All `open` declarations should be removed one by one.
I've removed all except `Migrate_parsetree.OCaml_404.Ast`, which seems a lot more complex than others and will be in another PR.
I've specified long identifiers for the expressions OCaml wasn't able to find explicity, thus fixing all these compilation issues. `make build` is working and all tests via `make test` are passing. I've also replaced the `TODO` by the next task.

### DIFF
--- a/src/reason-parser/reason_parser.mly
+++ b/src/reason-parser/reason_parser.mly
@@ -49,7 +49,7 @@
 
 %{
 open Migrate_parsetree.OCaml_404.Ast
-open Syntax_util
+(*open Syntax_util*)
 open Location
 open Asttypes
 open Longident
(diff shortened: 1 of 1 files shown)


src/reason-parser/reason_parser.mly
```diff
@@ -94,20 +84,20 @@ open Ast_mapper
...
+       {Parsetree.ptyp_desc = d; Parsetree.ptyp_loc = loc; ptyp_attributes = attrs}
      ..
    end
 
    parse_tree.mli
    --------------
    and core_type = {
-     ptyp_desc: core_type_desc;
-     ptyp_loc: Location.t;
+     Parsetree.ptyp_desc: core_type_desc;
```
jordwalke: You only need to do this on one of the record labels. I notice you've done it twice in several cases.

jordwalke: Want to make sure you just do it at most once per record? (It will help reduce merge conflicts).

Issue #1736, opened 2018-01-08
Put JSX in a JSX attribute


thangngoc89:
Refmt formats this 

```reason
<Description term=(<Text text="Age" />)> ... </Description>
```

into this 

```reason
<Description term=<Text text="Age" />> ... </Description>
```

And it's a syntax error.

Refmt gets it right with self-closing tag: 

```reason
<Description term=<Text text="Age" /> />
```

PR #1745, opened 2018-01-11
Fix1736

IwanKaramazow:
Fixes https://github.com/facebook/reason/issues/1736

This PR adds printing of braces around JSX as JSX-arguments:
```reason
<Description term={<Text text="Age" />}> ... </Description>;
<Foo bar={<Baz />} />
```
Error messages for ambiguous cases are also provided:
captura de pantalla 2018-01-12 a las 21 35 02
captura de pantalla 2018-01-12 a las 21 35 57


ubsan:
This seems unfortunate. Could we make it context dependent instead?

jordwalke:
Great! So, if we convert some code from OCaml to Reason that uses `/>>`, how is it printed? I think we need a way to express  the `/>>`identifier with some test cases showing that it is possible.

IwanKaramazow:
Per discussion with @chenglou, braces are printed around JSX as jsx-arguments:
```reason
<Description term={<Text text="Age" />}> ... </Description>;
<Foo bar={<Baz />} />
```
The infix operators are preserved this way, also added tests, and it's a bit more familiar for JS-developers.
Error messages for ambiguous cases are also provided:
captura de pantalla 2018-01-12 a las 21 35 02
captura de pantalla 2018-01-12 a las 21 35 57

thangngoc89:
Why don’t we use normal () braces? Those curly braces look ugly

chenglou:
@thangngoc89 mostly for familiarity with existing JSX, and to reduce the amount of parentheses.

jordwalke:
To reduce the number of parens, we could print them as `{ }` in any other JSX cases that require parens.

src/reason-parser/reason_pprint_ast.ml
```diff
@@ -2452,6 +2452,26 @@ let is_direct_pattern x = x.ppat_attributes = [] && match x.ppat_desc with
   | Ppat_construct ( {txt= Lident"()"}, None) -> true
   | _ -> false
 
+let isJSXComponent loc args =
```
jordwalke: Probably worth pulling out JSX logic into its own module in the future.

PR #1747, opened 2018-01-15
Make lexer not crash on `//`.

cristianoc:
Currently the lexer crashes on `//`.
With this, it likely gives a syntax error if used as a comment.

### DIFF
--- a/formatTest/unit_tests/expected_output/basicStructures.re
+++ b/formatTest/unit_tests/expected_output/basicStructures.re
@@ -127,6 +127,8 @@ let x = arr^[0] = 1;
  *============================================================================
  */;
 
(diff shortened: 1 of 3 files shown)


jordwalke:
Thanks!

We can make it not crash the lexer by landing this. Then, separately, how would you feel about (in addition to not crashing the lexer), making `//` be used for end-of-line comments like in JS/C/Java etc?

@let-def has a prototype of it he's been working on.

cristianoc:
I'd love to have full support for `//` for end-of-line comments. Is @let-def's prototype already in one of the pull requests? I thought I had seen something but can't find it right now.

jordwalke:
I don't think it's in an active pull request - working out some details.

PR #1761, opened 2018-01-21
Break up into multiple packages

jordwalke:
Make 'rebuild' package testable w esy workflow

Now, rebuild is in its own opam package, and more of the internals of the reason repo are being broken into their own packages (rtop is next).
That means that to build them with `esy` (such that the tests succeed)  we need to tell each package to be built and installed.

This also restores the ability to depend on master branch of Reason from another esy project.


jordwalke:
Landing as this only affects the esy workflow and Circle passed.

PR #1763, opened 2018-01-21
Add more source maps

jordwalke:
Summary: This is required in order to enable intelligent printing of trailing commas, but probably improves some other printings too. The SourceMap was just lost at some point on these AST structures.

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/reasonComments.re
+++ b/formatTest/typeCheckedTests/expected_output/reasonComments.re
@@ -268,8 +268,8 @@ let result =
  */
 let res =
   switch (
-    X
-      (2, 3) /* Retain this */
+    /* Retain this */
+    X(2, 3)
   ) {
   /* Above X line */
   | X(
--- a/formatTest/unit_tests/expected_output/wrappingTest.re
+++ b/formatTest/unit_tests/expected_output/wrappingTest.re
@@ -252,62 +252,56 @@ let defOptionalAliasAnnot =
   10;
 
 /* Invoking them */
-named
+named(
   /* a::a */
-  (
-    ~a,
-    /* b::b; */
-    ~b
-  );
+  ~a,
+  /* b::b; */
+  ~b
+);
 
-named
+named(
   /* a::a */
-  (
-    ~a,
-    /* b::b; */
-    ~b
-  );
+  ~a,
+  /* b::b; */
+  ~b
+);
 
-optional
+optional(
   /* a::a */
-  (
-    ~a,
-    /* b::b; */
-    ~b
-  );
+  ~a,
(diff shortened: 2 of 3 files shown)


formatTest/typeCheckedTests/expected_output/reasonComments.re
```diff
@@ -268,8 +268,8 @@ let result =
  */
 let res =
   switch (
-    X
-      (2, 3) /* Retain this */
+    /* Retain this */
```
jordwalke: Fixes this issue that happened since v3.

formatTest/unit_tests/expected_output/wrappingTest.re
```diff
@@ -252,62 +252,56 @@ let defOptionalAliasAnnot =
   10;
 
 /* Invoking them */
-named
+named(
```
jordwalke: Definitely improves this case too. Now the comment on the first arg is in the correct location when printed.

formatTest/unit_tests/expected_output/wrappingTest.re
```diff
@@ -252,62 +252,56 @@ let defOptionalAliasAnnot =
...
     /* None */
-    (
-      ~a=?None,
-      /* b::? */
-      /* None; */
-      ~b=?None
-    );
+    ~a=?None,
+    /* b::? */
+    /* None; */
```
jordwalke: This one is a little worse, but it's much less common that the other things that this diff improves. In future formatting fixes, we hope that even this one will be improved.

formatTest/unit_tests/expected_output/wrappingTest.re
```diff
@@ -2588,13 +2582,12 @@ let callMeWithComments =
 
 let result =
   /* Comment before function to invoke */
-  callMeWithComments
+  callMeWithComments(
```
jordwalke: another improvement.

src/reason-parser/reason_pprint_ast.ml
```diff
@@ -6888,8 +6889,19 @@ class printer  ()= object(self:'self)
     | `NormalFunAppl args ->
         let theFunc = SourceMap (funExpr.pexp_loc, self#simplifyUnparseExpr funExpr) in
         (*reset here only because [function,match,try,sequence] are lower priority*)
+        let syntheticArgLoc = match args with
```
jordwalke: Here's where I synthesize a location for the arguments list `call(a, b)`. There is not actually a tuple, so there wasn't a location - so I constructed one from the information available.

jordwalke:
Travis mac builds are unbelievably slow. We're talking days!

Issue #1767, opened 2018-01-22
[refmt] for-loop functions with comments that have single unit arg have weird formatting


kyldvs:
This [(try reason link)](https://reasonml.github.io/en/try.html?reason=GYewTgBAFAlhMDsIAYIBcQQIwEoIG8AoCCAegCoIBjEAW1oFME0JzTiIAHMRNAfQQMA7gBtEDKDkIBfANyEgA):

```
for (i in 0 to 1) {
  /* comment */
  print_newline()
};
```

Gets formatted to:

```
for (i in 0 to 1) {
  /* comment */
  print_newline
    ()
};
```

All of these are fine:

```
for (i in 0 to 1) {
  ();
  /* comment */
  print_newline()
};

for (i in 0 to 1) {
  print_newline()
};

for (i in 0 to 1) {
  /* comment */
  print_int(1)
};
```

jordwalke:
Remaining (more rare) issue tracked here: #1781

Issue #1768, opened 2018-01-22
[refmt] Formatting of trailing comments in for-loops is not idempotent


kyldvs:
(what happens to the comment is also probably a bug)

Start here: [(try reason link)](https://reasonml.github.io/en/try.html?reason=GYewTgBAFAlhMDsIAYIBcQQIwEoIG8AoCCAegCoIBjEAW1oFME0JzTiIAHMRNAfQQMA7gBtEDKDgDcHCtTqNmrdiW68BwsYMmzKNekxZtCAXxmFJUoA)

```
for (i in 0 to 1) {
  /* comment */
  print_newline();
  /* comment */
  print_newline()
  /* comment */
};

();
```

After first refmt:

```
for (i in 0 to 1)
  {
    /* comment */
    print_newline();
    /* comment */
    print_newline()
  };
  /* comment */

();
```

After second refmt:

```
for (i in 0 to 1) {
  /* comment */
  print_newline();
  /* comment */
  print_newline()
};

/* comment */
();
```

jordwalke:
Remaining (more rare) issue tracked here: https://github.com/facebook/reason/issues/1781

PR #1770, opened 2018-01-23
Fix interleaving of comments imperative structures.

jordwalke:
Fix interleaving of comments imperative structures.


jordwalke:
Fixes #1767 and #1768

jordwalke:
Merging! Let me know if this caused any problems, and we'll quickly revert.

kyldvs:
Thanks :)

PR #1772, opened 2018-01-23
Don't include semicolon in structure item loc. Makes it consistent wi…

jordwalke:
Don't include semicolon in structure item loc. Makes it consistent with other separators like tuples.

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/reasonComments.re
+++ b/formatTest/typeCheckedTests/expected_output/reasonComments.re
@@ -40,7 +40,8 @@
  **
  */
 module JustString = {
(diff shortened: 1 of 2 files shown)


formatTest/typeCheckedTests/expected_output/reasonComments.re
```diff
@@ -40,7 +40,8 @@
  **
  */
 module JustString = {
-  include Map.Make(Int32); /* Comment eol include */
+  include
+    Map.Make(Int32); /* Comment eol include */
```
jordwalke: One casualty of this change - but it will be improved by future diffs.

Issue #1549, opened 2017-10-26
Big issue for trailing comma


chenglou:
Gathering a bunch of syntaxes where we don't allow trailing commas. Feel free to add yours here

rostero1:
```
let animal: NotHuman.animal = {
  id: None,
};
```

`Error: 2539: <UNKNOWN SYNTAX ERROR>`

chenglou:
```
type foo = {
  "a": string,
};

type foo = {
  "a": string,
  "b": string,
};
```

jordwalke:
Option 1: Better error messages:
Options 2: Leniently parse them, even if removed upon print.

Any other ideas?

bsansouci:
@jordwalke is there any way to allow and enforce trailing commas?

chenglou:
Definitely option 2. No error message is better than having error + manual correction.

jordwalke:
@bsansouci Yes, we can allow and enforce them. You'd typically only want the trailing comma if it immediately preceeds a line break. Our formatting utilities don't provide facilities for those kinds of conditional separators, but the solution is surprisingly easy. You insert some dummy text that you replace in a final post-processing stage. However, I'd like to make sure that doesn't slow down `refmt` printing significantly by adding another pass over the output. Either way, first step would be to parse it (and print it as it currently prints).

zploskey:
I didn't see function calls yet.
```reason
let foo = baz(
  a,
  b,
);
```

IwanKaramazow:
Does it make sense to allow them in `ConstructorExpr(1,2,3,)`?

ckknight:
If they're allowed in function calls, I'd want them to also be allowed in Ctor instantiation. It's least surprising that way.

TheSpyder:
Lists would be nice too

```
let foo = [
  'a',
  'b',
  'c',
];
```

chenglou:
All solved.

PR #1775, opened 2018-01-23
🎉 Trailing commas 🎉 

jordwalke:
Intelligent printing and parsing of trailing commas on basically everything!

Please see the stack of diffs here. Each commit accomplishes some purpose and it will be much easier to understand the entire set of changes by looking at the individual commits.

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/patternMatching.re
+++ b/formatTest/typeCheckedTests/expected_output/patternMatching.re
@@ -1,6 +1,6 @@
 type point = {
   x: int,
-  y: int
+  y: int,
 };
 
 let id = x => x;
@@ -17,19 +17,19 @@ let computeTuple = (a, b, c, d, e, f, g, h) => (
   a + b,
   c + d,
   e + f,
-  g + h
+  g + h,
 );
 
 let res =
   switch (TwoCombos(Unused, Unused)) {
   | TwoCombos(
       HeresTwoConstructorArguments(x, y),
-      HeresTwoConstructorArguments(a, b)
+      HeresTwoConstructorArguments(a, b),
     ) => (
       x,
       y,
       a,
-      b
+      b,
     )
   | TwoCombos(_, _) => (0, 0, 0, 0)
   | Short
@@ -37,7 +37,7 @@ let res =
       100000,
       100000,
       100000,
-      100000
+      100000,
     )
   | AlsoHasARecord(firstItem, two, {x, y}) =>
     computeTuple(
@@ -48,7 +48,7 @@ let res =
       firstItem,
       two,
       two,
-      two
+      two,
     )
   };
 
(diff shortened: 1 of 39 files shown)


chenglou:
For reference: https://github.com/facebook/reason/issues/1549

formatTest/typeCheckedTests/expected_output/patternMatching.re
```diff
@@ -116,13 +116,13 @@ let res =
     id(
       fun
       | Some(x) => x + 1
-      | None => 0
+      | None => 0,
```
arnihermann: Is this trailing comma expected?

jordwalke: This is a trailing comma on the function argument. I’m pretty sure this is what we want. The alternative is to not print trailing commas for functions accepting only one argument.

ubsan: I think this would be a lot nicer to look at with different `function` syntax

jordwalke: Yeah, if `fun`s had braces, it would be nicer. There was discussion about adding them. In the mean time, the comma is kind of nice in that it tells you pretty clearly where to add another argument. We could also special case it so that single argument functions whose argument is a `fun |` do not print the trailing comma.

hcarty: > In the mean time, the comma is kind of nice in that it tells you pretty clearly where to add another argument

This seems like a compelling argument for keeping the trailing comma.

jordwalke:
I've found that trailing commas help with a bit of paren-fatigue as well, because it helps you distinguish closing parens for arguments/tuples/variants from parens whose only purpose is to resolve precedence - at least in the case where things break onto multiple lines. It's a small help, but it is a help.

formatTest/unit_tests/expected_output/bucklescript.re
```diff
@@ -90,7 +90,9 @@ let c = {
 let d = {
   "a": a2,
   "b": b,
-  "func": a => {"a": (arg1, arg2) => arg1 + arg2}
+  "func": a => {
+    "a": (arg1, arg2) => arg1 + arg2
+  }
```
chenglou: Seems missing here

formatTest/typeCheckedTests/expected_output/patternMatching.re
arnihermann: It felt a bit odd when I saw it, but it makes sense.

I really like this change and I agree it helps with paren-fatigue 👍

formatTest/unit_tests/expected_output/bucklescript.re
jordwalke: Yes JS objects were the only ones I didn’t get to. I’ll try to add them before landing.

jordwalke:
Going to merge:

Suggestions for anyone trying to extend this to other features in the future:
**It's very important to make sure your printer doesn't write a check that your parser can't cash.**
Meaning, if you're going to print trailing commas (or anything), make sure it can be parsed. It's tough to catch all the bugs because many times trailing commas (or other features we want to add in the future) will be omitted most of the time. To catch these cases, go into the printer and disable the part that strips out these trailing commas when lines don't break. Then run the tests. Search for idempotent failures to parse. It will catch all of them. Apply that same technique to features you add in the future that are "intelligently omitted". (Such as leading bars on switch cases etc).

Issue #1574, opened 2017-10-30
[proposal] `(. )` checked application


bobzhang:
```re
[@bs] f (a,b,c)
-->
f (. a, b,c) 

[@bs]
let f = (b,c) => {b + c }
-->
let f = (. b, c) => {  b + c}
```

chenglou:
I think this makes sense now, under the new application syntax

bobzhang:
@chenglou  `f (.)`

chenglou:
cc @IwanKaramazow 

I'm ok with this I think. Though it might not be the time to onboard newcomers and introduce two application syntax yet. Not sure

IwanKaramazow:
Ok, working on it

bobzhang:
could this be prioritized? we will add a few libs, having a native syntax for uncurried callback will help people consume such libs

IwanKaramazow:
Yes, I'll see what I can do this week.

chenglou:
I believe the snippet should have been:

```reason
let f = [@bs] ((a, b) => a + b);
```

ubsan:
Would `let f = (. a, b) => a + b;` have type `(int * int) -> int`? or would it be `int -> int -> int`?

chenglou:
Nono, it has nothing to do with tuples. It's a BuckleScript-specific optimizations that annotates the type with some info that the BS compiler understands.

If you wanted tuple you can do it today (two layers of parens)

ubsan:
ah, okay. So it's typed the same way as a normal function, but with some annotations?

PR #1779, opened 2018-01-24
Uncurried syntax

IwanKaramazow:
Fixes https://github.com/facebook/reason/issues/1574

Feedback welcome
cc @bobzhang @chenglou @jordwalke 

PS: documentation is coming

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/attributes.re
+++ b/formatTest/typeCheckedTests/expected_output/attributes.re
@@ -312,9 +312,11 @@ and anotherClassType = {
   pub bar: int
 };
 
-class type _x = {.
-  pub height: int
-};
+class type _x =
+  [@bs]
+  {
+    pub height: int
+  };
 
 class type _y = {
   [@bs.set]
--- a/formatTest/typeCheckedTests/expected_output/attributes.re
+++ b/formatTest/typeCheckedTests/expected_output/attributes.re
@@ -312,18 +312,16 @@ and anotherClassType = {
   pub bar: int
 };
 
-class type _x =
-  [@bs]
-  {
-    pub height: int
-  };
+class type _x = {.
+  pub height: int
+};
 
 class type _y = {
   [@bs.set]
   pub height: int
 };
 
-[@bs]
+[@attr]
 class type _z = {
   pub height: int
 };
--- a/formatTest/typeCheckedTests/expected_output/attributes.rei
+++ b/formatTest/typeCheckedTests/expected_output/attributes.rei
@@ -29,8 +29,7 @@ external createClassInternalHack :
   "createClass";
 
 [@bs.send.pipe: array('a)]
-external map : [@bs] (('a => 'b) => array('b)) =
-  "";
+external map : (. ('a => 'b)) => array('b) = "";
 
 [@bs.val] [@bs.module "react"]
 external createClassInternalHack :
--- a/formatTest/unit_tests/expected_output/uncurried.re
+++ b/formatTest/unit_tests/expected_output/uncurried.re
@@ -10,22 +10,11 @@ f(. a, b, c);
 [@attr]
 f(. a, b, c);
 
-f(. ~a, ~b, ~c);
-
-[@attr]
-f(. ~a, ~b, ~c);
-
 let f = (. a, b) => a + b;
-
-let f = (. ~a, ~b) => a + b;
 
 let f = [@attr] ((. a, b) => a + b);
 
-let f = [@attr] ((. ~a, ~b) => a + b);
-
 let f = (. a, b) => a + b;
-
-let f = (. ~a, ~b) => a + b;
 
 let f = ("hello", (. b, c) => b + c);
 
@@ -37,8 +26,6 @@ let obj: tesla = {
   pub drive = (. speed, safe) => (speed, safe);
   pub drive2 =
     [@attr] ((. speed, safe) => (speed, safe));
-  pub drive3 =
-    (. ~speed, ~safe) => (speed, safe);
   pub park = (.) => ();
   pub park2 = [@attr] ((.) => ())
 };
@@ -48,10 +35,6 @@ type f = (. int, int) => int;
 type f = [@attr] ((. int, int) => int);
 
 type f = (. int, int) => int;
-
-type f = (. ~a: int, ~b: int) => int;
-
-type f = (. ~a: int, ~b: int) => int;
 
 type z = (. unit) => unit;
 
@@ -85,21 +68,15 @@ funWithCb("text", (.) => doStuff());
 
 funWithCb("text", (. test) => doStuff());
 
-funWithCb("text", [@attr] (. test) => doStuff());
+funWithCb("text", [@attr] (. arg) => doStuff());
 
-test(~desc="my test", ~f=(.) => {
+test(~desc="my test", (.) => {
   let x = a + b;
   let y = z + c;
   x + y;
 });
 
-test(~desc="my test", ~f=(. a, b, c) => {
-  let x = a + b;
-  let y = z + c;
-  x + y;
-});
-
-test(~desc="my test", ~f=[@attr] (. a, b, c) => {
+test(~desc="my test", [@attr] (. a, b, c) => {
   let x = a + b;
   let y = z + c;
   x + y;
@@ -112,10 +89,10 @@ Thing.map(
   ~d=?foo2,
   ~e=?bakjlksjdf,
   ~f=?okokokok,
-  ~cb=?[@attr] (. abc, z) => {
+  ~cb=[@attr] (. abc, z) => {
     let x = 1;
     MyModuleBlah.toList(x, argument);
-  }
+  },
 );
 
 type f = int => (. int) => unit;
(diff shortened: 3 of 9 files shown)


src/reason-parser/reason_parser.mly
```diff
@@ -132,6 +132,7 @@ open Ast_mapper
 
 *)
 
+let bs_payload loc = ({loc; txt = "bs"}, PStr [])
```
jordwalke: I think that this should be a general purpose ppx attribute called `@uncurried` which isn't specific to buckleScript - at least at parse time. Thoughts? Then in the reactjs jsx transform which is where most of the BS logic is located, we can turn these into `@bs` attributes.

formatTest/typeCheckedTests/expected_output/attributes.re
```diff
@@ -311,18 +311,16 @@ and anotherClassType = {
   pub bar: int
 };
 
-class type _x =
-  [@bs]
-  {
-    pub height: int
-  };
+class type _x = {.
```
chenglou: lol! Not sure we should be giving this a sugar... but that's interesting

formatTest/unit_tests/expected_output/uncurried.re
```diff
@@ -0,0 +1,115 @@
+f(.);
```
chenglou: Is there an example for `foo => (. bar => baz)`?

jaredly:
Oohh I had an idea about this the other day. What if we use `#`, mirroring the javascript attribute-access?
like: `some#()` is `[@bs] some()`. Also for declaring a function, you can do `let x = #(a, b) => a + b`

formatTest/typeCheckedTests/expected_output/attributes.re
jordwalke: @IwanKaramazow Can you explain how the `.` here impacts the perceived semantics of class types? Does it somehow "close" them in the same way that `x(.1, 2)` makes function application "closed" and how `{. x : int}` makes object types closed?

formatTest/unit_tests/expected_output/uncurried.re
IwanKaramazow: Will add one in the next commit.
The current implementation output `type f = (foo, . bar) => baz;`

formatTest/typeCheckedTests/expected_output/attributes.re
IwanKaramazow: Hmm I saw a `[@bs]` attribute & automatically implemented the syntax sugar without thinking. It felt more consistent. Anything {. } or (. ) feels more like [@bs]  now.
I can omit it.

src/reason-parser/reason_parser.mly
IwanKaramazow: I actually started out by parsing an `@uncurried` attribute and transforming it to `@bs`.
One thing that kept me from going that route is existing code:
Imagine having
```reason
let f = [@bs] (a, b) => a + b;
```
One of the current "strengths" of our printer is the automatic transformation to
```reason
let f = (. a, b) => a + b;
```
Refmt teaches developers that `[@bs] (a, b) => ...` can be written in a much shorter way as `(. a, b) => ...`.

I'm actually in favour of parsing the `@uncurry` and delegating to `React ppx`, as a way to make Reason not Bs dsl, but the "teachable moment" kept it from here.
Am I missing something?
Should we just parse `@uncurry` and risk people typing `[@bs] (. a, b) => a + b`?
@chenglou What do you think?

formatTest/typeCheckedTests/expected_output/attributes.re
jordwalke: The OO types already kind of throw a wrench into that. I think we should make dot consistently mean “closed”, not “@bs”.

src/reason-parser/reason_parser.mly
jordwalke: If printing were to be lossy, you could print either form with the leading dot. It would be reparsed as uncurried. That would provide the teachable moment, but at the expense of removing the desirable properties of refmt: that formattting is never lossy with respect to the AST. At least it would be idempotent.
Perhaps we could make an exception for lossinees this one time and then in the next version stop auto converting ‘@bs’ to ‘@uncurried’

chenglou:
@jaredly hummm might be a little late, not sure. But I like it.
@jordwalke let's wait for this PR before release

jordwalke:
We can hold off on release until this is in. Releases are pretty lightweight now though.

jaredly:
Sorry I didn't have the idea earlier! I've always thought the `(.` syntax was super weird, but didn't have a better alternative.

src/reason-parser/reason_parser.mly
IwanKaramazow: Sounds good, I'll get right to it.

formatTest/typeCheckedTests/expected_output/attributes.re
IwanKaramazow: Consider it done. 😄

src/reason-parser/reason_parser.mly
chenglou: Discussed a little. This shouldn't be in the ReactJS JSX PPX as it has nothing to do with it. Not just because of the name, but because not everyone uses the ReactJS JSX PPX. And it'd be really confusing if folks want to have another JSX transform then suddenly have the @ uncurry -> @ bs logic change.

The clean way to solve it would be for BuckleScript to also support @ uncurry as an alias to @ bs. We shouldn't try to mingle here.

In the meantime, since no one uses @ bs for anything other than the BuckleScript uncurrying logic, I think we should keep it this way. Heck, touching @ uncurry might actually be bad right now by the same logic.

jordwalke: "In the meantime, since no one uses @ bs for anything other than the BuckleScript "

That wasn't my concern. My concern was that we want to use the uncurry syntax outside of bucklescript. I don't think there should be hard coded assumptions on BuckleScript in the Reason parser. If you want to have this use "@bs" attributes it will likely break in the future. Right now, nothing else is using the uncurried syntax, but if it does, it must be okay to move Reason to use a different ppx attribute "@uncurried" (or something like that) without anyone getting upset that their BuckleScript "@bs" uncurrieds broke. If that's not okay, then the solution is to make it so that no one is unhappy by having Reason parser parse into some bs-agnostic attribute such as "@uncurried", and either have BS pay attention to "@uncurried", or apply some post-processing to turn it into "@bs".

chenglou: Are you saying people are using the dot notation while using another ppx that also looks for @ bs tag and transforms it in some custom way, while also not use bucklescript? Even if that's the case, it's easy for _those_ ppxes to handle both @ bs and @ uncurry

formatTest/typeCheckedTests/expected_output/attributes.re
bobzhang: indeed, maybe we can delay this change (over class type) a bit.

formatTest/typeCheckedTests/expected_output/attributes.rei
```diff
@@ -29,8 +29,7 @@ external createClassInternalHack :
   "createClass";
 
 [@bs.send.pipe: array('a)]
-external map : [@bs] (('a => 'b) => array('b)) =
-  "";
+external map : (. ('a => 'b)) => array('b) = "";
```
bobzhang: the dot position is a bit weird, should not it be 
```reason
external map : ((. 'a) => 'b ) => array ('b)
```

formatTest/unit_tests/expected_output/uncurried.re
bobzhang: `(foo) => (. bar) => baz` looks  more consistent?

formatTest/unit_tests/expected_output/uncurried.re
```diff
@@ -0,0 +1,115 @@
...
+
+f(.);
+
+f(. a, b, c);
+
+[@attr]
+f(. a, b, c);
+
+f(. ~a, ~b, ~c);
```
bobzhang: currently, uncurried application does not support labels, maybe you can make an error earlier?

formatTest/typeCheckedTests/expected_output/attributes.rei
IwanKaramazow: The dot represents a `[@bs]` payload. Here the `[@bs]` sits on the `('a => 'b)`
```reason
external map : ([@bs] ('a => 'b)) => array('b) = "";
external map : (. ('a => 'b)) => array('b) = "";
```

formatTest/unit_tests/expected_output/uncurried.re
IwanKaramazow: Currently Reason uses `type f = (foo, bar) => baz;` which represents Ocaml's `type f = foo -> bar -> baz`. I agree that `foo => (. bar => baz)` is easier to read, will change.

IwanKaramazow:
@jordwalke @chenglou Just rebased against latest master, should be ready.

cristianoc:
With nested uncurrying, it looks like
```
add(. 2)(. 3)
```
reformats to
```
add(. 2, 3)
```
which loses the inner uncurry.

Instead this stays unchanged:
```
(add(. 2))(. 3)
```

IwanKaramazow:
@cristianoc Good catch, I'll take care of it this evening!

PR #1780, opened 2018-01-24
POC/WIP: Attempt at prettier printing of UTF-8 strings

hcarty:
Basic, simple-minded use of `Uutf` for slightly safer printing of string literals.

Guaranteed to have bugs in its current state.

### DIFF
--- a/src/reason-parser/reason_pprint_ast.ml
+++ b/src/reason-parser/reason_pprint_ast.ml
@@ -2368,13 +2368,29 @@ let constant_string f s = pp f "%S" s
 
 let tyvar f str = pp f "'%s" str
 
+let pp_string_literal f s =
+  let buf = Buffer.create 4 in
+  Uutf.String.fold_utf_8 (
+    fun () _i uc ->
+      match uc with
+      | `Malformed s -> failwith (Format.asprintf "Invalid UTF-8 string literal: %s" s)
+      | `Uchar c ->
+        if Uchar.to_int c <= 255 then
+          pp f "%s" (Char.escaped (Uchar.to_char c))
+        else (
+          Uutf.Buffer.add_utf_8 buf c;
+          pp f "%s" (Buffer.contents buf);
+          Buffer.clear buf
+        )
+  ) () s
+
 (* In some places parens shouldn't be printed for readability:
  * e.g. Some((-1)) should be printed as Some(-1)
  * In `1 + (-1)` -1 should be wrapped in parens for readability
  *)
 let constant ?(parens=true) f = function
   | Pconst_char i -> pp f "%C"  i
-  | Pconst_string (i, None) -> pp f "%S" i
+  | Pconst_string (i, None) -> pp f "%a" pp_string_literal i
   | Pconst_string (i, Some delim) -> pp f "{%s|%s|%s}" delim i delim
   | Pconst_integer (i, None) ->
       paren (parens && i.[0] = '-') (fun f -> pp f "%s") f i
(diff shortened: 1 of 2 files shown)


hcarty:
As an example, the following formats properly:
```
let fancy = "🔥";
```

chenglou:
I'll leave a note here: the bspack workflow needs to be tested too. We need to ship refmt in bs and on web

rauanmayemir:
@hcarty `uutf` package needs to be added to opam dependencies, could you please add it to `esy.json` as well? I'd like to test this PR on my end.

src/reason-parser/reason_pprint_ast.ml
```diff
@@ -2368,13 +2368,29 @@ let constant_string f s = pp f "%S" s
 
 let tyvar f str = pp f "'%s" str
 
+let pp_string_literal f s =
+  let buf = Buffer.create 4 in
+  Uutf.String.fold_utf_8 (
+    fun () _i uc ->
+      match uc with
+      | `Malformed s -> failwith (Format.asprintf "Invalid UTF-8 string literal: %s" s)
```
hhugo: The string doesn't have to be valid utf. it seems wrong to fail here.
one can easily write a string litteral such as "\xc3\x28"

hcarty:
@rauanmayemir I'll try to set aside some time to work on this. I wanted to get a (very) basic version out in case it would be of interest here.

src/reason-parser/reason_pprint_ast.ml
hcarty: Agreed - `failwith` is not meant to be permanent.  The current PR really is meant to be a minimal start to how something like this could work.

hcarty:
Thank you each for taking a look at this.

IwanKaramazow:
What happens when someone's editor is in utf16 mode?

hcarty:
@IwanKaramazow This PR would fail.  A proper implementation could handle that a number of different ways, with the caveat that I am not properly familiar with the nuances of unicode handling.

The simplest conceptually is to do what @jordwalke and others have suggested, pulling string literal bytes directly from the input file and printing them verbatim.

If something more invasive along this lines of this PR is desirable then the source file's encoding could be detected with uutf, determining which `fold` to use over string literals.  Annotations could be used to specify the encoding of a specific string, allowing UTF-8 strings embedded in a UTF-16LE source, for example.

Are UTF-16 encoded files supported currently?

jordwalke:
Iwan, so we can make that UTF8 assumption at lex time? How hard would that change be? What are the downsides?

rauanmayemir:
One more thing to consider is characters number per line, hope reason parser will account for 'real length' when deciding to reformat switch cases and arrays.

```
let fancy = ["🔥🔥🔥🔥🔥", "🚀🚀🚀🚀🚀"];
```

jordwalke:
It would not account for that in the most naive implementation. Iwan is building a convenient Unicode library that would help with this.

Octachron:
Did you consider using the same partial solution as the ocaml toplevel in [4.06.0](https://github.com/ocaml/ocaml/pull/1231)) which escapes only C0-control character , DEL, `"`, and `//`?

bsansouci:
Heyyy what's the status for this? I'm really itching for a solution to this as I cannot refmt some of my code currently :(

jordwalke:
I talked about it with @let-def. I think we need to go with my suggested approach (basically just copying comments). Feel free to send a PR.

PR #1785, opened 2018-01-25
Outcome printer: remove function type single-arg parens

chenglou:
`(int) => int` now becomes `int => int`.
Still doubly wraps for tuple. Wraps for labeled args too, etc.


jordwalke:
🎉 Merge when ready.

chenglou:
cc @glennsl

PR #1787, opened 2018-01-25
Reason_pprint_ast: exclude bs.obj from extension_sugar

### LINKED ISSUE #1786: JS object at right hand side of pattern match arrow formatted incorrectly.
@chenglou Says it blames to https://github.com/facebook/reason/commit/af2d02e787b031da838581c305022f563e2ddc42

Here's a repro:

```sh
echo 'echo 'switch(x) { | _ => {"aria-hidden": 2} };' | refmt
```

Prints incorrectly as:

```
switch (x) {
| _ => %bs.obj
       {aria-hidden: 2}
};
```

Which cannot parse.

let-def:
This should fix #1786

### DIFF
--- a/src/reason-parser/reason_pprint_ast.ml
+++ b/src/reason-parser/reason_pprint_ast.ml
@@ -263,7 +263,8 @@ type infixChain =
 let expression_extension_sugar x =
   if x.pexp_attributes <> [] then None
   else match x.pexp_desc with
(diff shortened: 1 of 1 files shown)


cristianoc:
Thanks!
This fixes the original issue that was minimized for a repro in the issue.

jordwalke:
Thank you!

PR #1793, opened 2018-01-29
Relaxed semicolons WIP/RFC

jordwalke:
WIP demonstrating how to relax some semicolons. I'd like to see how @let-def suggests doing this with less boiler plate and how to extend it to structure items which is a little more difficult.

The reason why I had to break the rules into two disjoint sets (with_attributes, and without) is because there would otherwise be shift reduce conflicts on parses like this:

```
let x = 0
[@attr]
let y = 0
```

Because the parser can shift the attribute on, potentially expecting `let x = 0 [@attr] and y = 0`, but it can also reduce a single let binding `let x = 0`. So breaking the rules apart allowed me to say that you can omit the semicolon only when the next binding begins clearly with a token such as "let/module" and doesn't begin with an attribute.

### DIFF
--- a/esy.json
+++ b/esy.json
@@ -12,13 +12,13 @@
     "@opam/result": "=1.2.0",
     "@opam/ocaml-migrate-parsetree": "*",
     "@opam/ocamlbuild": "*",
+		"@esy-ocaml/merlin": "*",
     "@opam/jbuilder": "*"
   },
   "peerDependencies": {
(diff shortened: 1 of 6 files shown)


jordwalke:
Abandoning this in favor of @let-def's approach. It's good that this is here so that we can show how this approach requires obscuring the meaning of the grammar rules.

PR #1799, opened 2018-01-31
Cleanup pprint ast

let-def:
Extract comments and layout related code from Reason_pprint_ast to Reason_comment and Reason_layout modules.
Remove dead code from Reason_pprint_ast.

PR #1800, opened 2018-01-31
Align doc comments, support single line // comments

let-def:
Refactor the implementation of comments:
- doc comments are now aligned like other comments.
- `//` are accepted.

Comments are now indented inside `Easy_format`, this enables a uniform treatment of all comments  (normal comments, single line comments, doc comments, whether they are printed as atoms are attached to list wrappers).

Hence the modifications to Easy_format.
`pp_print_as` is used to trick `Format` to put newlines: `pp_force_newline` forces a newline in the current box, while using `pp_print_as` with a size larger than the current margin puts the newline in the next box that receives content.

To merge after https://github.com/facebook/reason/pull/1799 .


formatTest/typeCheckedTests/input/reasonLineComments.rei
```diff
@@ -0,0 +1,3 @@
+module JustString : {
+  include Map.S; // Comment eol include
```
jordwalke: We will need to get the syntax highlighters to be updated. Vim should be good. But imagine if you had a brace or mismatched paren in a comment.

src/reason-parser/reason_lexer.mll
```diff
@@ -636,6 +636,11 @@ rule token = parse
       }
 
 and enter_comment = parse
+  | "//" ([^'\010']* newline as line)
```
jordwalke: @let-def Are there other kinds of newlines we need to watch out for (something like windows)?

PR #1801, opened 2018-02-01
Uncurried syntax for outcome printer

chenglou:


src/reason-parser/reason_oprint.ml
```diff
@@ -294,33 +294,39 @@ and print_arg ppf (lab, typ) =
...
-        match args with
-        (* single argument should not be wrapped... *)
-        | [(_, Otyp_tuple _) as arg] ->
+        match (uncurried, args) with
+        (* single argument should not be wrapped *)
+        (* though uncurried type are always wrapped in parens. `. a => 1` isn't supported *)
+        | (false, [(_, Otyp_tuple _) as arg]) ->
           (* ...unless it's for tuple. ((int, int)) => string *)
           pp_print_string ppf "(";
+          print_bs_uncurried_dot ();
```
cristianoc: Remove? This line never prints.

Nit: the logic might be simpler by doing the wrapping in one place, and do the case analysis just to compute a boolan `should_wrap`.

chenglou:
cc @IwanKaramazow

PR #1803, opened 2018-02-01
Nested uncurrying + paren hugging

IwanKaramazow:
As @cristianoc pointed out, nested uncurrying should be possible:
```
add(. 2)(. 3);
```
now formats as
```
add(. 2, . 3);
```

Also adds test cases/support for uncurried syntax with "paren hugging":
```
f(. (1, 2));
f(. {
  a: "supersupersupersupersupersuperlong",
  b: "supersupersupersupersupersuperlong",
});
```

### DIFF
--- a/formatTest/unit_tests/expected_output/uncurried.re
+++ b/formatTest/unit_tests/expected_output/uncurried.re
@@ -113,3 +113,24 @@ Thing.map(
 type f = int => (. int) => unit;
 
 type f = int => (. int) => unit;
+
+add(. 2);
+
+add(. 2);
+
+add(. 2, . 3);
+
+add(. 2, . 3);
+
+type timerId;
+
+[@bs.val]
+external setTimeout :
+  ((. unit) => unit, int) => timerId =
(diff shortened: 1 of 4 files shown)


chenglou:
Hummm, not sure we should support `add(. 2, . 3)`? Should we?

IwanKaramazow:
Unfortunately, I think we should.
* The behaviour in master is a bug. I'd rather have the parser be "resilient" against this case.
* It's consistent with the formatting of `add(2)(3) -> add(2, 3)`
* It's consistent with types where we allow `type f = (int, . int) => unit` (but here it get's reformatted to `type f = int => (.int) => unit;`, unfortunately we don't have enough info at print time to do the same with expressions

jordwalke:
I don't see this as a super common use case, so I'm happy to see anything that makes the implementation cleaner and/or more consistent.

chenglou:
^ Sounds good then. Merging! Haven't corrected the outcome printer yet.

Issue #1807, opened 2018-02-04
Unable to build refmt.js


glennsl:
Able to build reason on 4.04.2, but unable to do so on 4.02.3. But first, just to get to the point of trying to build reason I had to:

* install merlin-extend and utop into the 4.02.3 switch (I also think I had to install utop on 4.04.2, which wasn't mentoned in the build instructions)
* change `js_of_ocaml.exe` to `js_of_ocaml` in `bspacks/reason_bspack.sh`

This is the log from running `reason_bspack.sh`:

```
**This script is switching you to ocaml 4.02.3 for the subsequent bspacking. Please switch back to your own version afterward. Thanks!**\n
# To setup the new switch in the current shell, you need to run:
eval `opam config env`
~/dev/remote/_/reason/bspacks ~/dev/remote/_/reason/bspacks
make: Entering directory '/home/glennsl/dev/remote/_/reason'
rm -rf ./formatTest/**/actual_output
rm -f ./formatTest/failed_tests
rm -f ./miscTests/reactjs_jsx_ppx_tests/*.cm*
jbuilder clean
make: Leaving directory '/home/glennsl/dev/remote/_/reason'
make: Entering directory '/home/glennsl/dev/remote/_/reason'
jbuilder build
    ocamldep src/reason-merlin/ocamlmerlin_reason.depends.ocamldep-output
    ocamldep src/ppx/ppx_react.depends.ocamldep-output
    ocamldep src/refmt/refmt_impl.depends.ocamldep-output
    ocamldep src/refmttype/reason_format_type.depends.ocamldep-output
    ocamldep src/reason-parser-tests/testOprint.depends.ocamldep-output
    ocamldep src/reason-parser/vendor/easy_format/reason.easy_format.dependsi.ocamldep-output
    ocamldep src/reason-parser/vendor/easy_format/reason.easy_format.depends.ocamldep-output
    ocamldep src/generate/select.depends.ocamldep-output
    ocamllex src/reason-parser/reason_lexer.ml
    ocamldep src/reason-parser/menhir_error_processor.depends.ocamldep-output
    ocamldep src/rtop/rtop.depends.ocamldep-output
    ocamldep src/ppx/ppx_react.dependsi.ocamldep-output
    ocamldep src/reason-parser/vendor/cmdliner/reason.cmdliner.dependsi.ocamldep-output
    ocamldep src/reason-parser/vendor/cmdliner/reason.cmdliner.depends.ocamldep-output
[...]

PR #1808, opened 2018-02-04
Reduce refmt.js code size (probably)

glennsl:
Since I can't build reason on 4.02.3 (see #1807) I haven't actually been able to test this, but from experimenting with the jsoo build of BuckleScript (see https://github.com/BuckleScript/bucklescript/issues/2490), these changes should reduce the size of refmt.js pretty significantly (bsc was reduced from 3.7MB wth pretty + cc to 3.0MB with disable share). Running closure compiler on it afterwards will just bloat it back up.

### DIFF
--- a/bspacks/reason_bspack.sh
+++ b/bspacks/reason_bspack.sh
@@ -96,9 +96,7 @@ ocamlfind ocamlc -bin-annot -g -w -30 -w -40 -package js_of_ocaml,ocaml-migrate-
 # link them together into the final bytecode
 ocamlfind ocamlc -linkpkg -package js_of_ocaml,ocaml-migrate-parsetree -g -o "$REFMT_API_FINAL.byte" "$REFMT_API.cmo" "$REFMT_API_ENTRY.cmo"
 # # use js_of_ocaml to take the compiled bytecode and turn it into a js file
-js_of_ocaml.exe --source-map --debug-info --pretty --linkall +weak.js +toplevel.js --opt 3 --disable strict -o "$REFMT_PRE_CLOSURE.js" "$REFMT_API_FINAL.byte"
-# # use closure compiler to minify the final file!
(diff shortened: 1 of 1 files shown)


jordwalke:
Okay, we'll merge this once we get it testable. We can port the compile-to-js of refmt to esy, which is pretty straightforward because esy keeps its build artifacts around, and that means we can use bspack to include the post-processed OMP. Then it should be one command to build the refmt.js and more people will ensure it never breaks.

Another way to reduce the refmt.js size: we can also make it so that we insert one little layer of indirection for the parse tree such that OMP can be removed for this particular workflow (compiling refmt to JS). I also noticed we are shipping about twice as many modules for OMP in its current state (in the bspack) than is necessary. (We only use half of the bspacked OMP modules). I would have suggested finding a way to strip out the OMP unused modules which should trim another 30%, but I think we could almost as easily completely remove the OMP dependency for building `refmt.js` which should remove more along the lines of 50% of the size.

jordwalke:
One question: Do we need the AST to be converted to 4.02.3 encoding for the sake of `refmt.js`? If it's just to parse and print, it could remain in 4.04, which is my basis for claiming we could remove OMP from `refmt.js`. If we need the parse tree to be converted to 4.02.3, we will always need OMP (but we could trim the size of it).

hhugo:
Why does one need `--linkall` here ? Without it, you can probably get rid of `+toplevel.js`.
Why `--disable strict` ?
Also, if you want to optimize for space,  you should probably remove --debug-info.


The dev version of js_of_ocaml include `weak.js` by default and include a workaround for the issue mentioned in  https://github.com/BuckleScript/bucklescript/issues/2490

After a new version of js_of_ocaml is released. I would expect the js_of_ocaml invocation to be just: 
`js_of_ocaml.exe --source-map --opt 3 --disable strict -o "$REFMT_CLOSURE.js" "$REFMT_API_FINAL.byte"`

`

hhugo:
Fwiw, js_of_ocaml 3.1.0 has been released.
I believe the call could just be:
js_of_ocaml.exe --source-map --opt 3 --disable strict -o "$REFMT_CLOSURE.js" "$REFMT_API_FINAL.byte"

PR #1812, opened 2018-02-09
Update to circle 2.0

dwwoelfel:
Uses CircleCI 2.0, bringing the build time from 25 minutes to under 5 minutes (!). Without a cache, the build takes about 4:30. After the dependencies have been cached it only takes about 2 minutes: https://circleci.com/gh/dwwoelfel/workflows/reason/tree/circle-2

Using Circle's workflows, it's easier to run tests across multiple versions of ocaml all at once. Circle spins up a separate instance for each version and runs them all in parallel.

Caching is also much better. You have a lot more control over when you clear the cache and you save time uploading duplicate caches. To be safe, I clear the cache whenever any of the .opam files change or when circle's config file changes.

If someone wanted to do the work, the build time could be further reduced by about a minute with a custom Docker container.


dwwoelfel:
@jordwalke is it safe to cache the build artifacts from esy build?

dwwoelfel:
Added the `esy` steps in a separate container (sorry for missing that!). Now builds are a little under 10 minutes without a cache and 6-8 minutes with a cache. All of the extra time is from running `esy build`. 

There is probably room for improvement on the `esy build` step. It spends a few minutes with any console output. Any idea what it's doing there?

jordwalke:
cc @andreypopp. Not sure what it’s doing for three minutes without any console output. It might have been fixed in a newer esy release? This is still a massive improvement so we should land and continue to improve it. It’s safe to cache the artifacts produced in ~/.esy that are placed as a result of esy build.

.circleci/config.yml
```diff
@@ -0,0 +1,140 @@
...
+    - run:
+        name: 'Clean'
+        command: make clean-for-ci
+    - run:
+        name: 'opam pin'
+        command: |
+          opam pin add -y reason .
+          opam pin add -y rtop .
+    - save_cache:
+        key: dependency-cache-{{ checksum "../ocaml-version" }}-{{ checksum ".circleci/config.yml" }}-{{ checksum "reason.opam" }}-{{ checksum "rebuild.opam" }}-{{ checksum "rtop.opam" }}
```
jordwalke: Curious if people ever intentionally let the `save_cache` and `restore_cache` diverge.

.circleci/config.yml
```diff
@@ -0,0 +1,140 @@
...
+          command: |
+            curl -sL https://deb.nodesource.com/setup_9.x | sudo -E bash -
+            sudo apt-get install -y nodejs
+            mkdir -p ~/.npm-global
+            npm config set prefix $NPM_CONFIG_PREFIX
+      - run:
+          name: "Note OCAML_VERSION for cache"
+          command: echo $OCAML_VERSION > ../ocaml-version
+      - restore_cache:
+          key: dependency-cache-{{ checksum "../ocaml-version" }}-{{ checksum ".circleci/config.yml" }}-{{ checksum "reason.opam" }}-{{ checksum "rebuild.opam" }}-{{ checksum "rtop.opam" }}
```
jordwalke: I think for the `esy_build` portion, all we should need is a checksum on `esy.json`, `esy.lock` and `.circleci/config.yml` file for the cache key. The esy build process has no dependency on opam.
I think you can also remove the `OCAML_VERSION` as well. That's all modeled in the `esy.json` which influences the `esy.lock`. You could probably even move the npm stuff from "common" steps to the esy_build steps since the opam workflow has no dependency on npm/esy.

We can probably continue to use the `ocaml/opam:debian-9_ocaml-4.06.0` image since it will likely be cached and won't harm anything to use it.

.circleci/config.yml
```diff
@@ -0,0 +1,140 @@
...
+      - run:
+          name: "Install esy"
+          command: npm install -g esy@next
+      - run:
+          name: 'Clean'
+          command: make clean-for-ci
+      - run:
+          name: 'opam pin'
```
jordwalke: So if I understand correctly, this wouldn't be required for the esy_build workflow as well since esy workflow has no dependency on opam, right?

.circleci/config.yml
```diff
@@ -0,0 +1,140 @@
...
+          command: |
+            opam pin add -y reason .
+            opam pin add -y rtop .
+      - run:
+          name: 'esy install'
+          command: ~/.npm-global/bin/esy install
+      - save_cache:
+          key: dependency-cache-{{ checksum "../ocaml-version" }}-{{ checksum ".circleci/config.yml" }}-{{ checksum "reason.opam" }}-{{ checksum "rebuild.opam" }}-{{ checksum "rtop.opam" }}
+          paths:
+            - ~/.opam
```
jordwalke: I don't think we need `~/.opam` here either.

andreypopp:
> cc @andreypopp. Not sure what it’s doing for three minutes without any console output. It might have been fixed in a newer esy release? This is still a massive improvement so we should land and continue to improve it. It’s safe to cache the artifacts produced in ~/.esy that are placed as a result of esy build.

I'm not sure, I've seen this locally... it might hang on crawling the contents of `node_modules` if there are a lot of packages bug 3 mins is way too much for that.

.circleci/config.yml
dwwoelfel: > You could probably even move the npm stuff from "common" steps to the esy_build steps since the opam workflow has no dependency on npm/esy.

I tried that, but looks like one of the tests requires node. The tests fail at `node ./formatTest/testOprint.js`: https://circleci.com/gh/dwwoelfel/reason/84

Make the recommended updates to the esy_build portion. Thank you!

.circleci/config.yml
dwwoelfel: Removed, thanks!

.circleci/config.yml
dwwoelfel: Removed!

.circleci/config.yml
dwwoelfel: I doubt it--made it use a yaml anchor to prevent people forgetting to update one and not the other.

dwwoelfel:
@jordwalke implemented your feedback--thanks! 

The build is failing right now--looks like http://download.camlcity.org is unreachable. Same thing happens for me locally, so I think it's probably not related to the build steps. Will retry the build later today and hope that camlcity is back up.

dwwoelfel:
Oh boy! Caching ~/.esy after `esy build` dramatically reduced the build times! We're back to 2 minute builds with a cache! https://circleci.com/workflow-run/65218346-2023-40ac-8d3b-45ed0da22f6b

I think this is ready to go now.

jordwalke:
Sounds great! Merging.

PR #1817, opened 2018-02-12
Support attributes/doc-comments on record type fields.

### LINKED ISSUE #1653: Attributes on record label declaration
Hi,

I don't think it's currently possible to attach attributes to label declarations.
The ocaml syntax to do this is:

```ocaml
type t = { x : int [@attr "payload"] }
```

And the corresponding place in the AST is `pld_attributes`.

I'm not sure what happens since `refmt --parse=ml` outputs just `type t = {x: int};`, but on the other hand there is the following rule in the parser:

```ocaml
label_declaration:
  | item_attributes mutable_flag as_loc(LIDENT) COLON poly_type
    { let loc = mklocation $symbolstartpos $endpos in
      (Type.field $3 $5 ~mut:$2 ~loc, $1)
    }
```

So I'd expect something like `type t = { [@attr "payload"] x : int }` to work.

(note that `type t = { x : [@attr "payload"] int }` exists but is different, it attaches the attribute to the type itself)

Does this syntax exist? Or are there any plans to support it? Thanks!

jordwalke:
Summary:Fixes #1653

Test Plan:Test cases

Reviewers:

CC:

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/attributes.re
+++ b/formatTest/typeCheckedTests/expected_output/attributes.re
@@ -256,8 +256,7 @@ let showLets = () =>
  */
 [@moduleItemAttribute]
 class boxA ('a) (init: 'a) =
-  [@onReturnClassExpr]
-  {
+  [@onReturnClassExpr] {
     /**Floating comment text should be removed*/;
     /**Floating comment text should be removed*/;
     pub pr = init + init + init;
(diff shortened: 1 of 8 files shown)


jordwalke:
@zploskey and @kyldvs both found this issue.

formatTest/typeCheckedTests/expected_output/attributes.re
```diff
@@ -267,20 +266,17 @@ class boxA ('a) (init: 'a) =
  * In non-curried sugar, the class_expr still sticks to "the simple thing".
  */
 class boxB ('a) (init: 'a) =
-  [@stillOnTheReturnBecauseItsSimple]
```
jordwalke: I found the formatting change in this diff to be a slight improvement, especially if we were to allow omitting the brackets:

```
@attr {
  bracedExpression
}
```

jordwalke:
Please let me know if any of the small formatting changes for attribute attachments didn't result in improvements and I'll be happy to revisit.

PR #1821, opened 2018-02-14
Trailing commas in reason objects

### LINKED ISSUE #1820: No trailing comma printed for Js obj sugar
```reason
type foo = {
  .
  "bar": baz,
  "bar2": baz2,
};
```

Printing this removes the trailing comma after `baz2`. cc @jordwalke

zploskey:
Enforce printing of trailing commas in reason objects that contain multiple rows.

Fixes #1820.


chenglou:
Thanks!

PR #1826, opened 2018-02-16
Preserve function body braces and distinguish open vs. X.()

jordwalke:
Summary:Created a new place to store some stylistic information in the
AST. `[@refmt "foo"]` attributes used at parse time, never printed.

Test Plan:

Reviewers:

CC:

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/attributes.re
+++ b/formatTest/typeCheckedTests/expected_output/attributes.re
@@ -120,7 +120,11 @@ let x = [@attrEverything] (true && false);
 /**
  * How attribute parsings respond to other syntactic constructs.
  */
-let add = a => [@onRet] a;
+let add = a =>
+  [@onRet]
+  {
+    a;
+  };
 
 let add = a => [@onRet] a;
 
@@ -382,8 +386,18 @@ type classAttributesOnKeys = {
   .
   [@bs.set] key1: string,
   /* The follow two are the same */
-  [@bs.get null] key2: [@onType2] Js.t(int),
-  [@bs.get null] key3: [@onType2] Js.t(int),
+  [@bs.get
+    {
+      null;
+    }
+  ]
+  key2: [@onType2] Js.t(int),
+  [@bs.get
+    {
+      null;
+    }
+  ]
+  key3: [@onType2] Js.t(int),
   key4: Js.t([@justOnInt] int),
 };
 
@@ -533,7 +547,13 @@ let res =
 
 let res =
   switch (x) {
-  | _ => [@attr] String.(Array.(concat))
+  | _ =>
+    [@attr]
+    {
+      open String;
+      open Array;
+      concat;
+    }
   };
 
 /* GADT */
@@ -601,10 +621,6 @@ type tttttt = {
   [@attr "testing with mutable field"]
   mutable x: int,
 };
-
-let tmp =
-  /** On if statement */
-  (if (true) {true} else {false});
 
 type foo =
   option(
--- a/formatTest/typeCheckedTests/expected_output/braces.re
+++ b/formatTest/typeCheckedTests/expected_output/braces.re
@@ -1,0 +1,181 @@
+let preserveThis = "preserveThis";
+
+let x = a => {
+  preserveThis;
+};
+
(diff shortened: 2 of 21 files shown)


jordwalke:
Some people that have requested this: @sgrove

formatTest/typeCheckedTests/expected_output/attributes.re
```diff
@@ -120,7 +120,11 @@ let x = [@attrEverything] (true && false);
 /**
  * How attribute parsings respond to other syntactic constructs.
  */
-let add = a => [@onRet] a;
+let add = a =>
+  [@onRet]
+  {
```
chenglou: What's happening here

jordwalke: The input file actually has braces. The old test output was stripping the braces and now this diff preserves them.
This is what the input has:
```
let add(a) { [@onRet] a };
let add = fun(a) => [@onRet] a;
```
It doesn't distinguish between `[@a] { x }` and `{ [@a] x }` as you can see.

formatTest/typeCheckedTests/expected_output/attributes.re
```diff
@@ -382,8 +386,18 @@ type classAttributesOnKeys = {
   .
   [@bs.set] key1: string,
   /* The follow two are the same */
-  [@bs.get null] key2: [@onType2] Js.t(int),
-  [@bs.get null] key3: [@onType2] Js.t(int),
+  [@bs.get
+    {
```
chenglou: What's happening here then?

jordwalke: The same thing. The attributes were specified with braces, and before this diff they were getting `refmt`ed away.

chenglou: Does that mean these will always get formatted like this now or something?

TheSpyder:
While I really like the local open change, I’m less sold on retaining optional braces. I usually like the way simplifying a function can lead to the braces becoming unnecessary and collapse the code; often into a single line.

Although I guess it’s easy to argue both sides. This does make me think that reducing how opinionated refmt is means we start needing some options to control behaviour 🤔

jordwalke:
The open brace issue is pretty important because they get removed before you have the opportunity to add a let binding before the return value. Today people already do force the braces to be preserved by adding a dummy let binding to unit. That’s pretty extreme. What this diff does is just makes it more natural to tell the formatter you want them preserved.
However I think that if we had the ability to only print braces when the function body breaks, then this feature would be less useful and at that time perhaps we can revisit.

formatTest/typeCheckedTests/expected_output/attributes.re
kyldvs: If you remove the braces I guess they will get formatted how they were previously.

jordwalke: Exactly. Previously, the braces were getting formatted away. Now, we preserve whether or not you included additional braces.

kyldvs:
I run into the bracket issue all the time, particularly because I'm used to a format-on-save workflow.

```
let foo = (foo) => {
  /* some placeholder */
};
```

Will format out the braces and I have to add them back constantly. Typically my workaround is to always add some empty unit statements to the function body like this:

```
let foo = (foo) => {
  ();
  /* some placeholder */
};
```

Really looking forward to this change

sgrove:
@kyldvs Literally my exact workflow! Looking forward to it as well.

phated:
@jordwalke are you still planning to land this (or something similar)? I've been encountering both behaviors you outlined originally and it makes editing in vscode with `formatOnSave: true` very jarring.  Often I'll habitually save, get one of those reformats when I planned on editing in the block and have to undo the format (which sometimes ends in a weird result).

jordwalke:
@jaredly I felt that the refactor included in this diff was really necessary to make it easy to retain these "preserved styles". I found the current approach in the printer/parser not really sustainable and kind of fragile. With my refactor in this stack, there's a specific place carved out just for stylistic preservation and it is clearly out of the way because it's its own category that is cleanly filtered out.

PR #1831, opened 2018-02-20
Outcome printer: fix unary variants and infix operators.

cristianoc:
Fixes https://github.com/BuckleScript/bucklescript/issues/2522#issuecomment-366934176.

Fix printing of unary variants, that were missing parentheses. Fix printing of infix operators which differ between ocaml and Reason (`==`, `===`, `!=`, `!==`).

### DIFF
--- a/src/reason-parser/reason_oprint.ml
+++ b/src/reason-parser/reason_oprint.ml
@@ -125,7 +125,7 @@ let parenthesized_ident name =
 
 let value_ident ppf name =
   if parenthesized_ident name then
-    fprintf ppf "( %s )" name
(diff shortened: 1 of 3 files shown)


src/reason-parser/reason_oprint.ml
```diff
@@ -125,7 +125,7 @@ let parenthesized_ident name =
 
 let value_ident ppf name =
   if parenthesized_ident name then
-    fprintf ppf "( %s )" name
+    fprintf ppf "( %s )" (Syntax_util.ml_to_reason_swap name)
```
jordwalke: Looks good. Merge when ready!

Issue #1823, opened 2018-02-15
`[@bs]` attribute not parsed correctly in a certain case


chenglou:
Sorry for the vague title. Here's a repro:

```reason
[@bs]foo([@bs] {val a = 1})
```

Previously, this becomes:

```ocaml
let _ = foo ((object (this) val a = 1 end)[@bs ])[@@bs ]
```

Now it is incorrectly:

```ocaml
let _ = foo (object (this) val a = 1 end)[@@bs ]
```

Not too important; super low pri as probably no one uses it (except one case in ReasonReact. I can circumvent this easily)

idkjs:
Same problem? Seems like if you copy paste the [Try It json example](https://reasonml.github.io/en/try.html?reason=DYUwLgBADiBODOB7AdhAvBA3gImQQwFsRsAuLbAMwEsExSJsAhRAI2wBoHg946zsAWgAsCVMEOwBfTtjwBzYmQDMAJkkBuAFCbQkAFZJUGGAhQQAPgD4IAKXgA6Oynu9YVZHKoUAngEFk3hbWdvYA8lBgVM4KYACiAB6oVrYOTsj2UHgIIAnIWjrgEPhE6JoQEAYoZUEpjob2ACYgAMaITaEsei1g1ckh4ZHOeMgNACpCIMgAFADaAAIs8AC6EFNTUACU6MEOACJUzWD2MesyxcQbl707YRFR6cNjE9PziytrlchbaDdpjS1tEAdLqHKafS4ba61Ab3Y7gXL5XQQai0AByhBApXKIX2hzhYCm5xkKN42Eh5T6DhhQxG40mswWy1WYMM31+9SarSaAGUwG4PCyUBCof07tF4YlEYVuLx0SU0NUcQcjiciVweHRyTVRYMHrTngy3szwdtan9OYDefy5IKvlcKTdqekYgjtCFgIgbZg9OYABIgYAezgAEhJYDlmODMvDGPMekkGy0QA)  into vs-code, merlin reformats it to something else on save and then will not compile it. 

Specifically, its turning this 
```
  |> Js.Option.andThen([@bs] ((p) => Js.Dict.get(p, "name")))
  |> Js.Option.andThen([@bs] ((json) => Js.Json.decodeObject(json)))
```
 into this 
```
  |> Js.Option.andThen( => [@bs] (p => Js.Dict.get(p, "name")))
[...]

PR #1832, opened 2018-02-21
Fix printing & parsing of nested uncurried syntax.

IwanKaramazow:
Fixes https://github.com/facebook/reason/issues/1823

Parses/prints `add(. 1, .2)` correct as `(add(. 1))(. 2)`


IwanKaramazow:
Still one case failing, looking into it now.

src/reason-parser/reason_parser.mly
```diff
@@ -608,30 +608,96 @@ let process_underscore_application args =
...
+     *   we get the two nested Pexp_apply associated with
+     *   (add(. 2, 3))(. 4, 5)
+     *)
+    let rec make_appl body = function
+      | args::xs ->
+          let (uncurried, args) = args in
+          let expr = if args = [] then body
+          else
+            let (args, wrap) = process_underscore_application args in
+            let expr = mkexp (Pexp_apply (body, args)) in
```
cristianoc: How about taking the location range from the arguments:
```
            let args_loc = match args, List.rev args with
              | ((_, s)::_), ((_, e)::_) -> mklocation s.pexp_loc.loc_start e.pexp_loc.loc_end
              | _ -> assert false in
            let expr = mkexp (Pexp_apply (body, args)) ~loc:args_loc in

```

IwanKaramazow: Thanks for the code, exactly what was needed.

chenglou:
M-M-MERGED

PR #1838, opened 2018-02-27
Print non-ascii characters (for utf8) & keep raw string literals

let-def:
This PR implements two changes to the parser:

1) Don't escape non-ascii characters.
This allow printing of utf-8 strings without plenty of escapings. Only control characters, as well as `"` and `\`, are escaped with this patch.
The printing behavior is now comparable with OCaml one.

2) Keep literal string representation.
This preserve the exact representation of strings, when possible. The representation is stored in an attribute `reason.raw_literal`.

Patch 2) improves behavior of refmt from reason to reason. Patch 1) improves behavior of toplevel, as well as refmt, for any source (ast, ml, ...) that doesn't contain the `raw_literal` attribute.

### DIFF
--- a/formatTest/unit_tests/expected_output/escapesInStrings.re
+++ b/formatTest/unit_tests/expected_output/escapesInStrings.re
@@ -2,8 +2,13 @@
 /*
     let str = "@[.... some formatting ....@\n\010@.";
  */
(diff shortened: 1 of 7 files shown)


jordwalke:
Thank you, @let-def !

PR #1839, opened 2018-02-27
ocamlmerlin-reason: implement for_completion command

let-def:
This PR refines merlin integration by implementing the `for-completion` stub: this inserts a fake identifier in the token stream when completion is triggered.

This enables slightly more cases to typecheck and adjusts the shape of an AST in the middle of an expression to take into account the identifier to be inserted.

See this example, where the most likely completion should be `list_to_string`:

Before, the type can't be inferred:
before

After, merlin can first compute the type and then sorts the completion results:
after

Note: the completion position has to be passed to the lexer, the patch introduces a reference in `Reason_toolchain` to pass this side information. This seems to be the easiest way to pass extra information. Any suggestion?

### DIFF
--- a/src/reason-merlin/ocamlmerlin_reason.ml
+++ b/src/reason-merlin/ocamlmerlin_reason.ml
@@ -25,8 +25,12 @@ module Reason_reader = struct
     else
       structure (Reason_toolchain.RE.implementation buf)
 
-  let for_completion t _ =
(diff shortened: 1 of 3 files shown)


chenglou:
cc @jordwalke for review

chenglou:
Guess this is good to merge?

PR #1841, opened 2018-03-01
[Docs] Adds links to all issues/prs in changelog

mikaelbr:
Hi!

Just a small suggestion to make changelog easier to deep dive into details around the changes. In some cases, Github makes these links automatically, but I think this is only in releases and wikis, not in regular markdown files (can't tag people either).


chenglou:
In my own editor. But after thinking about it a bit it's stupid. I should optimize this for others.

Thanks! Merged.

PR #1843, opened 2018-03-01
Pretty print lowercase jsx as attr of a jsx with braces.

IwanKaramazow:
Prints braces around jsx attributes containing lowercase jsx:
```
<form label=<div /> />;
/* post-refmt: */
<form label={<div />} />;
```
Makes lowercase jsx as jsx attribute consistent with https://github.com/facebook/reason/pull/1745

### DIFF
--- a/src/reason-parser/reason_pprint_ast.ml
+++ b/src/reason-parser/reason_pprint_ast.ml
@@ -1836,7 +1836,7 @@ let isJSXComponent expr =
     | (Nolabel, _) :: rest -> false
     | _ :: rest -> hasSingleNonLabelledUnitAndIsAtTheEnd rest
     in
-    if List.length jsxAttrs > 0
+    if jsxAttrs != []
        && hasLabelledChildrenLiteral
        && hasSingleNonLabelledUnitAndIsAtTheEnd args
     then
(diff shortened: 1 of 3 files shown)


src/reason-parser/reason_pprint_ast.ml
```diff
@@ -1822,25 +1822,28 @@ let is_direct_pattern x = x.ppat_attributes = [] && match x.ppat_desc with
...
+      | (Labelled "children", _) -> true
+      | _ -> false
+    ) args in
+    let rec hasSingleNonLabelledUnitAndIsAtTheEnd l = match l with
+    | [] -> false
+    | (Nolabel, {pexp_desc = Pexp_construct ({txt = Lident "()"}, _)}) :: [] -> true
+    | (Nolabel, _) :: rest -> false
+    | _ :: rest -> hasSingleNonLabelledUnitAndIsAtTheEnd rest
+    in
+    if List.length jsxAttrs > 0
```
chenglou: Doing `jsxAttrs = []` is O(1) instead of O(n). Or just match on a 3-tuple

IwanKaramazow: Thanks! Should have thought about this...

chenglou:
Thanks! So what caused the small regression before?

IwanKaramazow:
I wrongly assumed that 'createElement' was part of lowercase jsx

chenglou:
Oh so you thought it was `div.createElement`? kk

chenglou:
Thanks!

Issue #1835, opened 2018-02-24
Placement of closing `>` in JSX


alexfedoseev:
```reason
/* Now */
<Component
  prop1
  prop2>
  ("child" |> ReasonReact.stringToElement)
</Component>

/* VS */
<Component
  prop1
  prop2
>
  ("child" |> ReasonReact.stringToElement)
</Component>
```

I used to the latter, imo it's easier to spot this place this way. Also, it's consistent with:

```reason
<Component
  prop1
  prop2
/>
```

zploskey:
Another option would be to further indent the properties.
```reason
<Component
    prop1
    prop2>
  ("child" |> ReasonReact.stringToElement)
</Component>
```
This makes the distinction between props and children more clear without adding an extra line. The Google style guide suggests formatting tags this way.

alexfedoseev:
I'd be consistent w/ Prettier and `/>` 😛

zploskey:
I agree it would be best if `/>` and `>` formatted consistently.

alexfedoseev:
@IwanKaramazow @rickyvetter TBH current behavior introduces a lot of inconvenience:
- it's not possible to delete last prop w/o moving closing `>` to the next line
- it's not possible to comment out last prop w/o moving closing `>` to the next line
- it's not possible to add new prop after the last on by pressing `o` (vim) or `Cmd + Enter` w/o moving closing `>` to the next line.

Can it be reconsidered?

PR #1846, opened 2018-03-06
Better alignment of closing ">" when jsx breaks.

IwanKaramazow:
Fixes https://github.com/facebook/reason/issues/1835
Small change which makes jsx a pleasure to read in reason (and more consistent with Prettier...)

```
// Before
<Namespace.Foo
  intended=true
  anotherOptional=200>
  child
</Namespace.Foo>

// After
<Namespace.Foo
  intended=true
  anotherOptional=200
>
  child
</Namespace.Foo>
```

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/jsx.re
+++ b/formatTest/typeCheckedTests/expected_output/jsx.re
@@ -280,7 +280,8 @@ let lotsOfArguments =
     argument3=3
     argument4=4
     argument5=5
-    argument6="test">
+    argument6="test"
+  >
     <Namespace.Foo />
   </LotsOfArguments>;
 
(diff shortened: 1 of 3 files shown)


rickyvetter:
If JSX is just sugar for function calls then I think the old syntax makes more sense. `children` is just a final argument and be treated as such. In the Before case the opening/closing tags mark the entire function call and all of the args are listed on their own lines. The After case complicates that a bit and adds special significance to the final arg that doesn't really seem meaningful.

On a more subjective note, the Before case is used heavily in the FB codebase and I think it'd be really hard to introduce this. (see --jsx-bracket-same-line setting in prettier).

IwanKaramazow:
Will close for now. Unless we make refmt configurable, better to not touch the existing behaviour.

alexfedoseev:
> If JSX is just sugar for function calls then I think the old syntax makes more sense. children is just a final argument and be treated as such.

My 2 cents. IMO the main point of JSX is to make function calls look like markup. So to make them look like a function call shouldn't be a goal.

Issue #1836, opened 2018-02-24
Parens hugging for Functor?


chenglou:
```reason
include
  Impl(
    {
      type nonrec t = taaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa;
    }
  );
```

Ideally:

```reason
include Impl({
  type nonrec t = taaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa;
});
```

alexfedoseev:
I believe related: https://github.com/alexfedoseev/re-formality/blob/master/examples/SignupForm.re#L25-L30

PR #1847, opened 2018-03-06
implement paren hugging for functor application

IwanKaramazow:
Fixes https://github.com/facebook/reason/issues/1836


chenglou:
cc @glennsl haven't forgotten about this yet

Issue #1853, opened 2018-03-08
underscore as single arg name is a SYNTAX ERROR


bsansouci:
I get `Error: 1575: <UNKNOWN SYNTAX ERROR>` with:

```
let a = _ => 0;
```

You might think this is dumb but in ReasonReact we have:
```
<div onClick=(_ => self.send(Click)) />
```

all the time because the argument is the JS event which we don't always need.

chenglou:
cc @jordwalke another thing from the workshop

PR #1855, opened 2018-03-08
Allow underscore as single argument to a function.

cristianoc:
Fixes https://github.com/facebook/reason/issues/1853.
Allow single argument functions where the argument is an underscore without parentheses: ` _ => 34`.
It can be confusing when a single variable argument is formatted without parentheses: from `(x)=>34` to ` x => 34`. And typically renamed with an underscore when not used ` _x => 34`, but using just underscore is a syntax error.

@let-def just checking that the changes are safe.

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/basics.re
+++ b/formatTest/typeCheckedTests/expected_output/basics.re
@@ -34,13 +34,13 @@ let l2 =
   |> List.map(optParam(~v=?_, ()))
   |> List.length;
 
-let argIsUnderscore1 = (_) => 34;
+let argIsUnderscore1 = _ => 34;
 
-let argIsUnderscore2 = (_) => 34;
+let argIsUnderscore2 = _ => 34;
 
-let argIsUnderscore3 = (_) : int => 34;
+let argIsUnderscore3 = _ : int => 34;
 
-let argIsUnderscore4 = (_) : int => 34;
+let argIsUnderscore4 = _ : int => 34;
 
 let argIsUnderscore5 = (_: int) => 34;
 
--- a/formatTest/typeCheckedTests/expected_output/basics.re
+++ b/formatTest/typeCheckedTests/expected_output/basics.re
@@ -34,6 +34,18 @@ let l2 =
   |> List.map(optParam(~v=?_, ()))
   |> List.length;
 
+let argIsUnderscore1 = (_) => 34;
+
+let argIsUnderscore2 = (_) => 34;
+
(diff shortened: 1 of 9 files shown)


formatTest/typeCheckedTests/expected_output/basics.re
```diff
@@ -34,6 +34,18 @@ let l2 =
   |> List.map(optParam(~v=?_, ()))
   |> List.length;
 
+let argIsUnderscore1 = (_) => 34;
```
chenglou: Shouldn't we print to `_ => 34`?

cristianoc: Not sure what's best. For consistency, sure. For readability, particularly the first time, on the fence.

formatTest/typeCheckedTests/input/basics.re
```diff
@@ -22,6 +22,18 @@ let l1 =
 let l2 =
   [Some(1), None, Some(2)] |> List.map(optParam(~v =?_, ())) |> List.length;
 
+let argIsUnderscore1 = _ => 34;
+
+let argIsUnderscore2 = (_ => 34);
+  
+let argIsUnderscore3 = _ : int => 34;
```
jordwalke: Side note: I realize return value type annotations look a little strange without parens on the argument. Whether or not it's an underscore.

formatTest/typeCheckedTests/expected_output/basics.re
jordwalke: I'm on the fence too. On one hand, I think the parens make it much more readable when there is a return value type annotation. I think in Flow syntax, the parens are even required to annotate the return type because otherwise it looks like you're annotating the single arg:
```
x : int => 0
```

But if there's no return value type constraint, I think that `_ => ret` is more readable and also helps reduce some of our "paren fatigue".

```reason
let zeros = List.map(_ => 0, myList);
```

vs.

```reason
let zeros = List.map((_) => 0, myList);
```

cristianoc: An underscore argument with return type annotation seems unusual.
Considering the common cases, and for consistency, looks like printing without parens has an edge.

formatTest/typeCheckedTests/input/basics.re
Huxpro: @jordwalke I also realize this problem. What about just refmt it as `(_) : int => 34` https://github.com/facebook/reason/issues/1977
It's a little bit weird when the arg is `_` tho.

Issue #1856, opened 2018-03-09
_ in variant with multi-arg is a syntax error


bsansouci:
I accidentally wrote (notice the `_` at the end):

```reason
{ pmod_desc: Pmod_constraint({...inner_pmod_desc, pmod_desc: Pmod_structure(newList)}, _) }
```

This was a syntax error.

jordwalke:
What would you expect the meaning of this to be?
We do not support `_` for constructors because `_` technically means partial application in functions.

bsansouci:
I just meant that I get `SYNTAX ERROR`, it'd be nice to have a better error than that :)

jordwalke:
Got it!

PR #1859, opened 2018-03-09
Specialize syntax error for underscore.

cristianoc:
Fixes https://github.com/facebook/reason/issues/1856.

Try to give a more descriptive error message that “UNKNOWN SYNTAX ERROR” when the error is caused by the presence of an underscore.

This is an attempt to give some hints, though it's not so easy to do in full generality.

E.g.
```
let x = A(1,_);
```
gives
```
File "a.re", line 1, characters 12-13:
Error: 3205: underscore is not a valid identifier. Use _ only in pattern matching and partial function application
```

### DIFF
--- a/src/reason-parser/reason_parser_explain.ml
+++ b/src/reason-parser/reason_parser_explain.ml
@@ -49,6 +49,11 @@ let semicolon_might_be_missing state _token =
   else
     raise Not_found
 
+let token_specific_message = function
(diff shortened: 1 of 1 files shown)


jordwalke:
Cool! You can also try the custom messages setup, which was cleared out in the V3 syntax. It's not well documented but I went through it the other day to make sure I could still get the workflow working.

PR #1861, opened 2018-03-13
Remove stale docs for real

chenglou:
This has been redirecting to the new doc site for a new year. I think
we're good.


chenglou:
cc @jordwalke

PR #1865, opened 2018-03-16
Rename syntax_util to reason_syntax_util

chenglou:
This file might need to be dragged into BS repo because the outcome
printer relies on it now. No problem with that, but every other file in
the directory start with `reason`, so let's make this one start with
`reason` as well. Easier to copy paste around without name clashes too.


chenglou:
Just a file rename. Good to merge.

PR #1866, opened 2018-03-16
Remove hacks from reason_oprint; add commented cppo stuff for reason_syntax_util

chenglou:
This makes syncing reason_syntax_utils into BuckleScript much easier.

No change on Reason's side

### DIFF
--- a/src/reason-parser/reason_syntax_util.ml
+++ b/src/reason-parser/reason_syntax_util.ml
@@ -1,9 +1,155 @@
+(* Hello! Welcome to the Reason syntax util logic.
+
+  This file's shared between the Reason repo and the BuckleScript repo. In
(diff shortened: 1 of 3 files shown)


src/reason-parser/reason_syntax_util.ml
```diff
@@ -1,10 +1,156 @@
...
+(* #if defined BS_NO_COMPILER_PATCH then *)
 open Ast_404
+(* #end *)
 
 open Asttypes
 open Ast_mapper
 open Parsetree
 open Longident
 
+(** Check to see if the string `s` is made up of `keyword` and zero or more
```
chenglou: Just moving the relevant section for reason_oprint in front. No change

PR #1869, opened 2018-03-16
Print newline after doc comments before attributes

### LINKED ISSUE #1857: Doc comments get placed on same line as external
On the 3.1.0 version of `refmt`, a doc comment on the line preceding an external gets reformatted onto the same line as the external if it is short enough. I think we probably want to preserve the line break.

Before:
```reason
/** The url of the response. */
[@bs.send] external url : t => string = "";
```
After:
```reason
/** The url of the response. */ [@bs.send] external url : t => string = "";
```

zploskey:
Forces a line break following doc comments before the first non-comment attribute on a BuckleScript external.

**Before:**

```reason
/** doc comment */ [@bs.send] external url : t => string = "";

/**
 * Short multiline doc comment
 */ [@bs.send]
external url : t => string = "";

/** Longer doc comment before an attribute on an external. */ [@bs.send]
external url : t => string = "";

/* normal comment */
[@bs.send] external url : t => string = "";
```

**After:**

```reason
/** doc comment */
[@bs.send]
external url : t => string = "";

/**
 * Short multiline doc comment
 */
[@bs.send]
external url : t => string = "";

/** Longer doc comment before an attribute on an external. */
[@bs.send]
external url : t => string = "";

/* normal comment */
[@bs.send] external url : t => string = "";
```

Fixes #1857.

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/reasonComments.re
+++ b/formatTest/typeCheckedTests/expected_output/reasonComments.re
@@ -738,3 +738,12 @@ external url : t => string = "";
 
 /* normal comment */
 [@bs.send] external url : t => string = "";
+
+/** doc type */
+type q = {
+  a: int,
+  b: string,
+};
+
+/** doc let */
+let letter: q = {a: 42, b: "answer"};
--- a/formatTest/typeCheckedTests/expected_output/reasonComments.re
+++ b/formatTest/typeCheckedTests/expected_output/reasonComments.re
@@ -721,3 +721,20 @@ let r = {
   fieldOne: (identifier: string), /*eol1*/
   fieldTwo: (identifier: string) /* eol2 with trailing comma */
 };
+
+/** doc comment */
+[@bs.send]
+external url : t => string = "";
+
+/**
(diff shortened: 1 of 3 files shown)


zploskey:
Rebased and added an additional test case for short multi-line doc comments preceding externals.

chenglou:
cc @rickyvetter

jordwalke:
Are there sufficient test cases on non-extern items that have doc comments? (Like type definitions and regular let bindings?) If not, would you care to add them so I can merge?

zploskey:
I rebased this on master and reworked it to use `partitionAttributes` now that it can handle doc comments again. If there is a simpler way to do something here please let me know. I'm eager to get this merged so I don't have to keep padding out my doc comments just so the attributes don't wind up on the same line as them.

zploskey:
CC @IwanKaramazow

IwanKaramazow:
@zploskey This looks great! Thanks for the effort, I'll ping @chenglou to make sure it gets merged.

chenglou:
Good stuff. Thanks @zploskey!

PR #1871, opened 2018-03-16
More familiar assert

IwanKaramazow:
Although `assert` is a specific expression (Pexp_assert), we can make it more familiar for newcomers.
We could print it as a classic function application to reduce cognitive overhead.
I.e. you just apply `assert` as a function. I don't see any advantages too printing it as a special language construct.

```reason
/* Before */
assert (Node.namedChildCount(root) == 1);
assert false;

/* After */
assert(Node.namedChildCount(root) == 1);
assert(false);
```

### DIFF
--- a/formatTest/unit_tests/expected_output/assert.re
+++ b/formatTest/unit_tests/expected_output/assert.re
@@ -1,0 +1,23 @@
+switch (true) {
+| true => ()
+| false => assert(false)
+| _ => assert(false)
+};
+
+let root = {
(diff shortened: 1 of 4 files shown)


IwanKaramazow:
@andreypopp esy ci fails here, is there anything I can do here?

chenglou:
I think I'm fine with this. cc @jordwalke what do you think?

jordwalke:
Cool, non-breaking change too.

PR #1874, opened 2018-03-17
Add an interface file to reason_syntax_util

chenglou:
Since BuckleScript also needs this file (used by outcome printer), we're adding an interface file to keep the features a bit better documented/enforced


src/reason-parser/reason_syntax_util.ml
```diff
@@ -347,12 +345,6 @@ let syntax_error_extension_node loc message =
  in
  (str, payload)
 
-let swap_txt map txt =
```
chenglou: Unused

Issue #1160, opened 2017-03-22
Refmt preserving spaces in modules.


kennetpostigo:
I'm not sure if the style that the reason team landed on was to not preserve spaces within `module` and `module type`, if so then disregard this issue. 

An example is:
my code:
```reason
module type Example = {
  let example0: someArg0::string => string;

  let example1: someArg1::string => string;

  let example2: someArg2::string => string;
}
```

gets turned into:
```reason
module type Example = {
  let example0: someArg0::string => string;
  let example1: someArg1::string => string;
  let example2: someArg2::string => string;
};
```

Same thing happens with module:
my code:
```reason
module Example = {
  let example0 someArg0 => "0";

  let example1 someArg1 => "1";

  let example2 someArg2 => "2";
};

```

gets turned into:
```reason
module Example = {
  let example0 someArg0 => "0";
  let example1 someArg1 => "1";
  let example2 someArg2 => "2";
};
```

Now imagine if the functions weren't one liners and mixed with other values aside from functions. This gets very hard to read and sometimes you skip a line reading code or it's hard to tell where something may have ended. Not to take a shot at the regl codebase or anything but when reading it I confused often when a functions was indeed over. These spaces are helpful in showing what may be one unit or feature. Also makes me as a programmer psychologically more at ease when I see spaces. But maybe I'm just being stupid over these spaces.

* this may not be exclusive to modules but this is where i've noticed spaces not being preserved.

hcarty:
IIRC @IwanKaramazow was working on some automatic grouping/spacing changes for refmt which should address this.

IwanKaramazow:
Yes the core is ready, need to fix some issues with comments.

It'll group things of the same type & breaks if it's spreaded over multiple lines.
```reason
module Something = {
    type a = int;
    type b = string;

    let x = 1;
    let y = "test";

    let render {props, state} => {
        /* same type, e.g. a `let`, but has newline above for readability */
        let () = Lwt_main.run (something ());
        <div> props.x state.y </div>
    }
}
```

Whitespace will still be killed in the above example, because it's to short of a unit.
If `let example3` consists of multiple lines, it'll get a newline above
```reason
module Example = {
  let example0 someArg0 => "0";
  let example1 someArg1 => "1";
  let example2 someArg2 => "2";

 let example3 () => {
     let a = 1;
     let b = 2;
     a + b
 };
};
```

jordwalke:
[...]

PR #1876, opened 2018-03-17
Intelligent whitespace interleaving (WIP)

IwanKaramazow:
# Intelligent Whitespace interleaving

## What does this mean?
Whitespace inserted by the user will be tracked and preserved by refmt in certain places.
Amount of newlines interleaved (or lack thereof) is now deducted from whitespace
interleaved in the source instead of hard coded defaults in the printer.
More than one newline will be collapsed into one newline.

Example:
```reason
open Multicore;
open React; /* no newline above */

/* newline above */
let add = (a, b) => {
  let a = 1;

  let b = 1; /* the newline above this let binding is preserved */
  a + b;
};

/* another comment */
/* attach this comment to the above comment */

/* but here we want a newline above */
```

## Where is whitespace inserted by the user preserved?

* structure items (both top-level in a `.re` file & inside `module X = { ... }`)
* signature items (both top-level in a `.rei` file & inside `module type Y = { ... }`)
* inside let-bindings

## How does it work?

Preserving whitespace can be split into two parts:
1) Computation of inserted whitespace (whitespace isn't part of the AST)
2) Formatting of the computed whitespace into the layout tree (without disrupting comment insertion)

### Computation of whitespace
The Ocaml-AST doesn't contain whitespace. This isn't a problem, since whitespace
between two items can easily be computed based on the location diff between two items.

Imagine the following source code:
```reason
1| let a = 1;
2|
3| let b = 2;
```
We can easily determine if there's whitespace between `let a` and `let b`,
by substracting the linenumber of the start of `let b` by the linenumber of the end
of `let a`. If the diff is equal to or greater than 1, we know that there's whitespace.

```reason
let whitespace = loc2.loc_start.pos_lnum - loc1.loc_end.pos_lnum >= 1
```

In practise the computation is a little bit more complex because of comments.
```reason
1| let a = 1;
2| /* comment on line 2 */
3| let b = 1;
```
Here the location diff indicates whitespace, but the comment fills up the space between
the two items. The correct diff is:
```reason
let whitespace = loc2.loc_start.pos_lnum - loc1.loc_end.pos_lnum - height_of_comments_between >= 1
```

### Formatting computed whitespace (without disrupting comment insertion)
Once we know we need to interleave whitespace between two items, we can't just
print an `atom "\n"` node. This would clash with the comment interleaving
algorithm. Without inspecting every atom for possible `\n` text, we would get
the following scenario

```reason
before refmt        after refmt
------------       ------------
let a = 1;      |  let a = 1;
                |  /* comment */ --> attached above `let b = 2;`
/* comment */   |               --> atom "\n", was formatted before 
let b = 2;      |  let b = 2;
```

To circumvent this problem, whitespace needs to be interleaved after all
comments have been inserted into the layout.
In order to postpone the insertion of whitespace to the last moment,
the following layout made its way into existence:

```reasonm
  type layout =
      ...
    | Whitespace of WhitespaceRegion.t * t
```
`Whitespace` is a specific type of layout. It conveys an
"intent to interleave whitespace". It decorates a layout with information
for whitespace insertion above that node.
`WhitespaceRegion.t` contains specific info
about the region where whitespace should be interleaved.
The info is used for correct comment interleaving inside of the region
designated for whitespace.

## Perfomance implications
To compute if a range in the source code contains whitespace,
we need the comments for that range. Every time the whitespace computation
happens, we need to traverse all comments to find the comments that overlap.
The traversal of comments can be improved by storing the comments inside a
interval-(Red/Black or AVL)tree for cheaper lookup.

As a benchmark reference: there's a slowdown of around 50ms with this diff,
on a 10000 line file. Reason_pprint_ast.ml (formatted in Reason) was taken
as reference point on a MacBook Pro 2015 i5.

The comment interleaving algorithm in general is very slow, for every comment
we create a create/travers the/a whole new tree.
There is definitely room for improvement here. The performance will be adressed
in a separate PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ test-ci: install test
 
 test: build clean-tests
 	node ./formatTest/testOprint.js
-	./miscTests/rtopIntegrationTest.sh
+	# ./miscTests/rtopIntegrationTest.sh
 	./miscTests/jsxPpxTest.sh
 	cd formatTest; ./test.sh
 
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ test-ci: install test
 
 test: build clean-tests
 	node ./formatTest/testOprint.js
-	# ./miscTests/rtopIntegrationTest.sh
+	./miscTests/rtopIntegrationTest.sh
 	./miscTests/jsxPpxTest.sh
 	cd formatTest; ./test.sh
 
--- a/esy.json
+++ b/esy.json
@@ -20,7 +20,7 @@
   },
   "devDependencies": {
     "@opam/merlin": "*",
-    "ocaml": "~4.2.3"
+    "ocaml": "~4.6.0"
   },
   "notes-ctd": [
     "This is how you make an esy monorepo for development, but then release the monorepo as many individual packages:",
@@ -32,12 +32,10 @@
   ],
   "esy": {
     "build": [
-      [ "jbuilder", "build", "-p", "rebuild,reason,rtop"]
+      [ "jbuilder", "build", "-p", "reason"]
     ],
     "install": [
-      ["esy-installer", "rebuild.install"],
-      ["esy-installer", "reason.install"],
-      ["esy-installer", "rtop.install"]
+      ["esy-installer", "reason.install"]
     ],
     "buildsInSource": "_build"
   }
--- a/esy.json
+++ b/esy.json
@@ -20,7 +20,7 @@
   },
   "devDependencies": {
     "@opam/merlin": "*",
-    "ocaml": "~4.6.0"
+    "ocaml": "~4.2.3"
   },
   "notes-ctd": [
     "This is how you make an esy monorepo for development, but then release the monorepo as many individual packages:",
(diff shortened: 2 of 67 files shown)


Schmavery:
This is awesome! Are you considering doing the break-if-over-multiple-lines thing you mentioned [here](https://github.com/facebook/reason/issues/1160#issuecomment-288476520) in the future as well? I remember thinking that was a great idea.

IwanKaramazow:
@Schmavery Hey, sorry for the late response. Managed to implement the "intelligent" whitespace interleaving inside modules, i.e. more than one newline between "items" will collapse into one newline that will be preserved. Since whitespace is a tool for us programmers to organise our code, I don't think we should let an algorithm/heuristic interfere with our human newlines atm. That might change in the future, but I think it's the best way forward atm. I like to error on the side of not producing ugly formatted code at this stage 😄

IwanKaramazow:
@chenglou  @jordwalke  ready for review

chenglou:
Tada!

PR #1883, opened 2018-03-27
Clean up some code

chenglou:
flatten a few things. Might make it easier for newcomers.


ckknight:
If you check this diff with no whitespace changes, there's 8 lines removed and 4 added and they're really straightforward. 👍

PR #1891, opened 2018-04-03
Fix #1868: Merlin typed-hole formatted incorrectly

### LINKED ISSUE #1868: Typed holes `(??)` like Merlin with OCaml
It does work, but two issues:

image

1. `[Error] Unbound value ??` - maybe use a different message (e.g. `Warning/Error: Unimplemented hole`)
2. `refmt` incorrectly formats it like this

image

Typed-holes are ridiculously useful for statically-typed programming (Agda, Idris) - would be great to shine some light on this feature - instead of/as well as gradually buildng types, we can gradually build terms :-)

dc-mak:


dc-mak:
Travis seemed to fail on XCode with OCaml 4.06 because of `[ERROR] The sources of the following couldn't be obtained, aborting: ...`.

Issue #1893, opened 2018-04-04
Refmt fails for local opens with jsx lists


rauanmayemir:
Refmt correctly recognises the syntax when you write something like:
```
Module.[
  <Component>
    <Comp />
  </Component>
]
```

But if you use a one liner like `Module.[<Component />]`, refmt will fail with `<UNKNOWN SYNTAX ERROR>`. (at the same time it will remove parens if you write `Module.([<Component />])`)

Cc: @IwanKaramazow

PR #1894, opened 2018-04-04
Fix parsing/lexing of jsx inside a list under open sugar.

IwanKaramazow:
Fixes https://github.com/facebook/reason/issues/1893

Allow:
```
  Module.[<Component />];
```


chenglou:
Thanks!

PR #1897, opened 2018-04-08
Restore syntax error messages table

mrandri19:
Restores the `src/reason-parser/reason_parser.messages.checked-in` file which was empty. The file was rebuilt using `menhir --explain --strict --unused-tokens ./reason_parser.mly --list-errors > ./reason_parse
r.messages.checked-in`.
It took about 4 minutes and 8.5 GiB of RAM on a i7 7500U.


mrandri19:
@jordwalke ping

PR #1904, opened 2018-04-18
Wrap vendored easy_format in a new name

### LINKED ISSUE #1740: Unable to load easy_format in rtop
There seems to be a conflict with the vendored `easy_format`.

Tested with OCaml 4.06.0 and Reason pinned to master (5457093e) with opam.

This prevents yojson from working in rtop, for example.

hcarty:
This helps avoid conflicts with yojson in rtop, for example.

I've only tested with native as I don't have an npm/bucklescript on my system.

Fixes #1740

### DIFF
--- a/src/reason-parser/reason_layout.ml
+++ b/src/reason-parser/reason_layout.ml
@@ -1,4 +1,5 @@
 module Comment = Reason_comment
+module Easy_format = Reason_easy_format.Easy_format
 
 type break_criterion =
   | Never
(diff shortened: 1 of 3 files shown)


IwanKaramazow:
I've been having the `brew install opam` on travis too, doesn't seem related.

chenglou:
This doesn't work under `npm run prepublishOnly` (the bspacking workflow). Changing the name of easy_format should be fine

PR #1910, opened 2018-04-26
JSX string literal

cristianoc:
Special case string literals in JSX so ReasonReact.string is added automatically.

### DIFF
--- a/miscTests/reactjs_jsx_ppx_tests/expected_v2_3.re
+++ b/miscTests/reactjs_jsx_ppx_tests/expected_v2_3.re
@@ -28,7 +28,7 @@ module Bar = {
 module ReasonReact = {
   let element = (~key=?, ~ref=?, component) => 1;
 
-  let stringToElement = x => x;
+  let string = x => x;
 };
 
 let divRef = ReactDOMRe.createElement("div", [||]);
--- a/miscTests/reactjs_jsx_ppx_tests/expected_v2_3.re
+++ b/miscTests/reactjs_jsx_ppx_tests/expected_v2_3.re
@@ -27,6 +27,8 @@ module Bar = {
 
 module ReasonReact = {
   let element = (~key=?, ~ref=?, component) => 1;
+
+  let stringToElement = x => x;
 };
 
 let divRef = ReactDOMRe.createElement("div", [||]);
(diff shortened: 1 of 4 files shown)


IwanKaramazow:
This raises another natural question, what happens with floats, ints, lists and arrays?
Since their "type" can be determined at lex-time and compiled with a ppx, do we still want the conversion for those others?

wegry:
@IwanKaramazow wouldn't it make sense that if we aren't sure about conversions, we stick with vanilla React's behavior, i.e. only strings are valid and not floats, ints, list, and arrays?

rickyvetter:
There are some pretty serious problems that come with this because ppx doesn't have access to the type information. It doesn't allow strings to be lifted outside of the ppx in a typesafe manner which will almost certainly result in issues with refactoring and some really confusing bugs for newcomers.

As an example:

```js
<div> "foo" </div>
```
will typecheck and
```js
let foo = "foo";
<div> foo </foo>
```
will not.

gilbert:
I think the string literal lifting issue is a small price to pay and should not be considered a deal breaker.

rickyvetter:
@gilbert it's not just lifting that will cause issues:

```js
<div> {"Hello, " ++ name} </div>
```
won't typecheck and
```js
<div> {localize("Title of the Page")} </div>
```
won't typecheck unless localize only works as a React element. You'll need a different function to operate outside of React.

If everyone were very familiar with these nuances then sure, I think this would be fine for convenience. What I'm worried about is how to explain this differing behavior to people who aren't aware of the nuances.

gilbert:
I see, thanks for showing the other issues.

kujon:
> It doesn't allow strings to be lifted outside of the ppx in a typesafe manner which will almost certainly result in issues with refactoring and some really confusing bugs for newcomers.

TypeScript has got a very similar behaviour when using type guards. It's the most annoying thing in the world, because it makes you re-factor code only because the compiler doesn't understand.

Let's not go there with Reason please.

kennetpostigo:
@rickyvetter I wonder how that would work for people using bs-react-native? They could probably alias it?

rickyvetter:
Yeah if this was a path we wanted to explore we'd work with the `bs-react-native` folks to make sure it would work for them.

texastoland:
A text element makes sense. Alternatively what about a version of `{j|hello, $interpolation|j}` that coerces to `reactElement`? That seems less intrusive to me since a newcomer likely also wants interpolation.

PR #1912, opened 2018-04-27
Rename vendored OCaml libraries to avoid conflicts

### LINKED ISSUE #1740: Unable to load easy_format in rtop

hcarty:
To minimize code churn, each file using the vendored libraries has an
alias to the original module name.

This is a replacement for #1904 in a form that is hopefully more friendly to all of Reason's compilation targets.

Fixes #1740


chenglou:
Works. Thanks!

jordwalke:
Thanks, just found a case where this saved me!

hcarty:
I'm glad it helped!

PR #1921, opened 2018-05-04
Allow whitespace interleaving between cases in pattern matching.

IwanKaramazow:
Collapse/preserve n >= 1 whitespaces between cases in pattern matching into one newline between the cases. This allows better grouping of cases when desired. Pattern matching suffers from less claustrofobia if the user wants to preserve the whitespace. If there isn't whitespace inserted between the cases, there won't be any extra newlines. This behaviour is also the default in tools like Prettier and Golang.


chenglou:
Ready to merge? Let's cut a release after this

IwanKaramazow:
Ready to merge!

chenglou:
Yeeeeee

PR #1925, opened 2018-05-14
Super spread errors

IwanKaramazow:
Add helpful error message for users trying to spread at the front or middle of a list.

image


src/reason-parser/reason_parser.mly
```diff
@@ -3276,12 +3276,41 @@ expr_list:
...
+      let exprList = List.map (fun (dotdotdot, e) -> match dotdotdot with
+      | Some dotdotdotLoc ->
+        (* If the `...` appears in any other location then the last,
+         * show a friendly, clear message explaining the consequences. *)
+        let msg = "A list in Reason is a single-linked list providing very
+efficient read and insert operations at the head of a list.
+[z, ...abc] is very efficient.
+However, when you're not appending to the front of a list, we need to traverse
+the whole list to merge them both. Due to the less than ideal performance
+implications, we don't allow spreading at the front or in the middle of a list.
```
hcarty: Maybe `don't allow spreading syntax` to indicate that it's the sugar that's not allowed, rather than the language not allowing concatenation in general.

IwanKaramazow: @hcarty yes you're right, good to have you here!

chenglou:
Dat speed lol
cc @cristianoc
What should our general approach be? To use the error table or to change the parser?

cristianoc:
@chenglou what speed implication -- reversing the list?

Overall, I like this style of change as it does not increase the complexity of the grammar, in fact it reduces it.

IwanKaramazow:
Yes, I tried to keep the grammar as simple as possible here. The problem with the error table is that you need to know all the exact states/grammar. I’m more in favour of a 100% deterministic solution. It’s as @cristianoc said in Vienna, we need to simplify the grammar

chenglou:
Seems like this is just for value and not patterns? If so, here's a message:

```
Lists can only have one `...` spread, and at the end.
Explanation: lists are singly-linked list, where a node contains a value and points to the next node. `[a, ...bc]` efficiently creates a new item and links `bc` as its next nodes. `[...bc, a]` would be expensive, as it'd need to traverse `bc` and prepend each item to `a` one by one. We therefore disallow such syntax sugar.
Solution: directly use `concat`.
```

For long error messages:
- I state the core point in a concise way first, as that's what you wanna see after seeing the message a few times.
- Give an example. Most concrete thing to understand for newcomers. `a` and `bc` is nice because of it casually explains relationship between the items. Not using `a` and `b` because `b` is many items.
- Provide the solution.

IwanKaramazow:
I'll address them in this PR.

chenglou:
(Discussed elsewhere. Will merge now)

PR #1927, opened 2018-05-15
Add support for functors in JSX tags

### LINKED ISSUE #1924: JSX syntax doesn't support functors
Functors are a pretty powerful feature that could be used in conjunction with JSX/ReasonReact in order to make building UIs easier. As an example, let's say you have a ImageGallery component that can display a gallery of different image components. Using functors, you could define:

```reason
module ImageGallery(I: Image) = {
  let component = ReasonReact.statelessComponent("ImageGallery");
  let make = (~imgs=[||], _children) => {
    ...component,
    render: self =>
      <div>
        (ReasonReact.arrayToElement(Array.map(img => <I img/>, imgs)))
      </div>,
  };
};
```

Then in turn you would create the Image type and various implementations. For example, we could have `SquareImage : Image` and `CircularImage : Image` as two different implementations. Instead of having to create an element for each `Image` and then creating a wrapper `ImageGallery` around that, you could instead simply give it the list of images to load and then change the type of child used in the `ImageGallery` like so:

```reason
let images = [| loadImage("img/test.png"), loadImage("img/example.jpeg") |]

ReactDOMRe.renderToElementWithId(<ImageGallery(SquareImage) images />, "preview");
```

As it stands, the only way to do this with the current ReasonML JSX extension is by creating a temporary module:

```reason
let images = [| loadImage("img/test.png"), loadImage("img/example.jpeg") |]

module SquareImageGallery = ImageGallery(SquareImage);
ReactDOMRe.renderToElementWithId(<SquareImageGallery images />, "preview");
```

Though this is only a minor inconvenience, it seems like it should be an easy addition. I'm happy to implement this and then open a PR if this is something that others think would fit in well, but I was wondering whether there's a reason that this wasn't already a feature. Any thoughts?

OhadRau:
Fixes #1924

This PR enables the use of functors in JSX tags. It may also alter some edge cases in the syntax (though these would not compile properly in the past) -- previously, `<M./>` would parse correctly but fail to bind to a module; in certain cases, this will still parse but it should result in an error earlier in the pipeline (indicating a syntax error). This PR allows for the following new syntax:

```
<F(X)/>;
<F(X)> </F>;
// <F(X)> </F(X)>; is not currently supported, but this would be an easy addition
```

The syntax is designed to make it clear that the functor application only occurs once, but again it is a relatively easy change to add support for (or even require) a matching end tag.

Internally, the transformation is done by creating a `let module` expression to temporarily bind the result of the applied functor and then calling the `createElement` function on that result. Some more info, as well as an example use-case, is show in #1924.


cristianoc:
Could you check what happens with the use case
```reason
<Foo.Bar/>
```
That's where the `.` character is expected to be used.
Perhaps there isn't a test for that use at the moment, and it should be added.

OhadRau:
Yeah, that seems to still work. To clarify, it should only affect the case with trailing dots, since I'm basically allowing for an arbitrary Longident (rather than only the `Lident` and `Ldot` constructors) for the capitalized case. However, the `<M.x/>` case doesn't work with this change--is this something that's intended to work?

src/reason-parser/reason_lexer.mll
```diff
@@ -532,7 +532,7 @@ rule token = parse
   | "[|" { LBRACKETBAR }
   | "[<" { LBRACKETLESS }
   | "[>" { LBRACKETGREATER }
-  | "<" uppercase_or_lowercase (identchar | '.') * {
+  | "<" (uppercase identchar* '.')* lowercase identchar* {
```
cristianoc: To double check: the main change here is `uppercase_or_lowercase` to `lowercase`, plus some `.` cleanup correct?

The cleanup can be applied to the two cases below too, using I guess
```(uppercase identchar* '.')* uppercase_or_lowercase identchar*```

File: formatTest/unit_tests/expected_output/jsx_functor.re [hunk 2]
```diff
@@ -0,0 +1,17 @@
...
+  let createElement = (~children=[], ()) => {
+    displayName: "X",
+  };
+};
+
+module F = (X: (module type of X)) => {
+  let createElement = X.createElement;
+};
+
+let body = <div> <F(X) /> </div>;
```
cristianoc: Are several arguments supported?
```reason
<F(X,Y) />
```

src/reason-parser/reason_lexer.mll
OhadRau: Yes. The change to `lowercase` as the main pattern is to make it only parse lowercase tags (e.g. `<div>`) as `LESSIDENT`. The parser implements the uppercase (opening) tags as a different pattern without `LESSIDENT`. The `.` cleanup was specifically added to allow tags like `<X.x>`, but doesn't support `<X.Y>`. I could add this to the other cases, but                                           they work simply by resetting the lexer to the beginning of the pattern, consuming only up to the `<` token. I assume that the current patterns will run slightly faster, and should work correctly at the parsing level. That said, it may make more sense to switch to `(uppercase identchar* '.')* lowercase identchar*` and `(uppercase identchar* '.')* uppercase identchar*` for the other patterns to make them look more similar and make the patterns more accurate.

formatTest/unit_tests/expected_output/jsx_functor.re
cristianoc: I've tried to add a test for `<F(X,Y) />` but I'm getting a syntax error on the file generated in `actual_output`. Could you try adding a multi argument example to the tests?

OhadRau: Hm, I think the syntax error is from the idempotent tests. When I test it with `refmt` it doesn't seem to print right. Looking into this now

OhadRau:
Not sure if you got a notification from the last pushes, but I've fixed the pretty-printing for functors with multiple arguments. Let me know if there's anything else that might be an issue with this.

cristianoc:
I wasn't sure what the plan was about tests. The PR currently adds empty files for `jsx_functor.re`.

OhadRau:
Weird, I must've accidentally cleared those files before pushing. Should be fixed now.

cristianoc:
Thanks!
From the CI, it looks like idempotent tests are failing with a syntax error
```
Idempotent Test: ./jsx_functor.re
File "/home/opam/project/formatTest/unit_tests/actual_output/./jsx_functor.re", line 36, characters 26-28:
```

OhadRau:
I feel stupid cause I keep running `make test` and thinking it worked only to realize I pushed broken code lol. Should work now

src/reason-parser/reason_pprint_ast.ml
```diff
@@ -3273,6 +3274,38 @@ let printer = object(self:'self)
...
       else None
     )
+    | (Pexp_apply ({pexp_desc=Pexp_letmodule(_, ({pmod_desc=Pmod_apply(m1, m2)} as app), {pexp_desc=Pexp_ident loc})}, l), [], _jsx::_) -> (
+      (* TODO: Soon, we will allow the final argument to be an identifier which
+         represents the entire list. This would be written as
+         `<tag>...list</tag>`. If you imagine there being an implicit [] inside
+         the tag, then it would be consistent with array spread:
+         [...list] evaluates to the thing as list.
+      *)
+      let rec extract_apps args = function
```
cristianoc: The compiler gives a non-exhaustive pattern matching warning here.

src/reason-parser/reason_pprint_ast.ml
```diff
@@ -3273,6 +3274,38 @@ let printer = object(self:'self)
...
+      let rec hasSingleNonLabelledUnitAndIsAtTheEnd l = match l with
+      | [] -> false
+      | (Nolabel, {pexp_desc = Pexp_construct ({txt = Lident "()"}, _)}) :: [] -> true
+      | (Nolabel, _) :: rest -> false
+      | _ :: rest -> hasSingleNonLabelledUnitAndIsAtTheEnd rest
+      in
+      if hasLabelledChildrenLiteral && hasSingleNonLabelledUnitAndIsAtTheEnd l then
+        if List.length (Longident.flatten loc.txt) > 1 then
+          if Longident.last loc.txt = "createElement" then
+            let ftor::args = extract_apps [] app in
```
cristianoc: non-exhaustive pattern matching here too

perhaps the code can be refactored a bit

cristianoc:
Great!
This looks good to me. Is it ready to merge?

OhadRau:
Yep, seems like it's all done. Thank you!

OhadRau:
I think anyone who's used OCaml/Reason a lot would assume it based on syntax, because each instance of `X(Y)` isn't the module. Also, I think intuitively it looks like a function application, which would hopefully be enough to get the user to realize the side effects are occurring each time. Maybe this is something that could be addressed in the docs? I doubt anyone but long-time users of OCaml/Reason would try this out, since it's not something that exists in the "real" JSX. Because of that, I think >90% of Reason users will only be introduced to this syntax by the docs, meaning that we can specifically mention it with a big warning. The reference cell might be something to be concerned with, but as far as I know, there's no way of accessing anything other than `createElement` without a PPX and hopefully that is enough to push people towards reducing or eliminating side effects of module instantiation.

Issue #1929, opened 2018-05-16
Doc comments separated inconsistently from definitions in interfaces


hcarty:
When reformatting a doc comment attached to a value's type in an interface (`.rei`), `refmt` from the current `master` branch adds unwanted whitespace between the comment and the definition if the definition wraps.  Otherwise there is no whitespace added between the comment and definition.

The following is the output from `refmt`, even if the empty line between the first doc comment and `let some_definition` is removed:
```reason
/** Hi I'm a doc comment for some_definition */

let some_definition:
  (
    ~what_a_long_label: M.its_a_type,
    ~just_need_to_make_this: long_enough_to_wrap
  ) =>
  unit;

/** Hi I'm a doc comment for another_definition */
let another_definition: unit => unit;
```
I'd personally prefer no empty line between the doc comment and definition, but in any case it would be nice if the spacing did not depend on the length/wrapping of the value definition.

PR #1934, opened 2018-05-16
Interleave and attach doc comments correct on signatures

IwanKaramazow:
Fixes https://github.com/facebook/reason/issues/1929

Gives correct attachment of doc comments on signatures, interleaves whitespace where appropiate.


hcarty:
Amazing as always, @IwanKaramazow

chenglou:
Is it good to be merged? I'll let you self-review since you're the expert here now lol

Issue #1937, opened 2018-05-16
refmt errors on parentheses-less args of functions within ppx-attrs


andreypopp:
This code:
```
[@test.call (string) => string]
let processCommandItem = 12;
```
reformats into:
```
[@test.call string => string]
let processCommandItem = 12;
```
which makes `refmt` throw on a further invocation with:
```
File "./failure.refmt.re", line 1, characters 19-21:
Error: 2282: syntax error, consider adding a `;' before
```

Looks like `[@test.call string => string]` isn't being accepted due to missing `(` and `)` around the arg.

Version used: https://github.com/facebook/reason/commit/0401064b696d0e9384b7824fbd0c7d6638cb553c

This looks like a regression as it worked with the previous release (reason-cli@3.1.0).

andreypopp:
I'm wrong, the refmt from reason-cli@3.1.0 also throws in this case.

IwanKaramazow:
I'll fix the parser.
How would you want to see this formatted? With or without parens on the single arg?
```
/* with parens */
[@test.call (string) => string]
let processCommandItem = 12;

/* without parens */
[@test.call string => string]
let processCommandItem = 12;
```

andreypopp:
I'm not sure, maybe without parens to be consistent with regular functions:
```
% echo 'let x = (x) => x;' | esy refmt
let x = x => x;
```

Issue #1902, opened 2018-04-16
`[%raw (a) => 1]` turns into `fun%raw a => 1` then errors


chenglou:
Repro:

- Format `[%raw (a) => 1]`
- See `fun%raw a => 1`
- Format `fun%raw a => 1`
- See error

3 issues here:

- `fun%raw` is kinda weird since we don't use `fun` anywhere anymore (and since the only usage of `fun` is for function cases, which might be misleading)
- `[%raw a => 1]` without parens around `(a)` is a syntax error
- `fun%raw a => 1` fails to reformat

IwanKaramazow:
I can look at this

PR #1943, opened 2018-05-18
Fix parsing & printing of es6 function syntax inside attribrutes

IwanKaramazow:
Fixes https://github.com/facebook/reason/issues/1937
Fixes https://github.com/facebook/reason/issues/1902

### DIFF
--- a/src/reason-parser/reason_lexer.mll
+++ b/src/reason-parser/reason_lexer.mll
@@ -919,6 +919,7 @@ and skip_sharp_bang = parse
 
   let inject_es6_fun = function
     | tok :: acc ->
+        print_endline "INJECTING";
       tok :: fake_triple ES6_FUN tok :: acc
     | _ -> assert false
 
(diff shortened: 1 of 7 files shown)


src/reason-parser/reason_lexer.mll
```diff
@@ -919,6 +919,7 @@ and skip_sharp_bang = parse
 
   let inject_es6_fun = function
     | tok :: acc ->
+        print_endline "INJECTING";
```
hcarty: Is this intentional?

chenglou: lol?

IwanKaramazow: forgot to rebase lol

IwanKaramazow: Forgot to remove it 💃

jordwalke:
I think I ended up having a good use case for this one. Glad to see this fix!

PR #1945, opened 2018-05-20
Fix syntax error when parsing `let x=-.1;` and co.

anmonteiro:
This is a follow-up to #1642 employing the strategy highlighted in the following
comment: https://github.com/facebook/reason/pull/1642#issuecomment-344709969

It removes the special case around `LABEL_WITH_EQUAL`, making other valid inputs
parseable too.

cc @IwanKaramazow


IwanKaramazow:
Great patch, simplifies the grammar in a nice way. Good to go.

anmonteiro:
Yeah I think I might have broken those. 

I'll add tests and make sure they're working. Thanks for the feedback

anmonteiro:
This is actually trickier than I envisioned, and possibly even unsolvable.

Imagine the following:

```
```

anmonteiro:
This is actually trickier than I envisioned, and possibly even unsolvable.

Imagine the following:

```
let (=+) = (a, b) => a + b;
let (=+-) = (a, b) => b;

let foo = a =+-b;
```

now what is really `foo`? It could either be parsed as `a =+- b` or `a =+ (-b)`

anmonteiro:
I think the answer to my question is infix precedence.

I managed to fix this issue in the end and added some regression tests around it.

This was definitely a tricky one, and I think we can make it work!

src/reason-parser/reason_lexer.mll
```diff
@@ -353,18 +353,6 @@ let () =
           None
     )
 
-(* To "unlex" a few characters *)
-let set_lexeme_length buf n = (
```
anmonteiro: there's another one around line 190

chenglou:
Man I wish we can disable userland infix operator one day, and then provide a set of e.g. 5-10 good ones, whose precedence we can hard-code, and whose symbol won't be constrained by the first char.

Anyway, is this diff ready?

anmonteiro:
I really agree with that "first character" statement.

Yeah this is ready to go. It's unfortunate that it complicates the grammar a little bit but both Iwan and I can't see another immediate way to support this case.

Issue #1947, opened 2018-05-20
Simplify grammar package type constraint on let bindings 


IwanKaramazow:
```reason
/* before */
let thing: (module Thing) = (module MyModule);

/* after */
let thing: Thing = (module MyModule);
```
Idea: the `(module Thing)` package type constraint syntax looks very foreign. Since modules are usually associated with an uppercase-cased first letter identifier, the package type constraint syntax can probably be simplified to just using the uppercase-cased first letter identifier (with possible dot for submodules).

The grammar needs to be a little bit restructured to make this work. Currently module types can be lower cased too, `(module thing)` is legal Ocaml for some reason, resulting in a reduce/reduce conflict.
```reason
/* ambiguity */
let thing: foo = (module Thing);
/* is foo a module type or is it a core_type */
```
By special casing this case to accept uppercase identifiers in the module type case, the grammar could possibly allow parsing of `let thing: Thing = (module MyModule);` without too much trouble.

chenglou:
I'm kinda suspecting that newcomers would accidentally mistake that type for a variant or something. Happens often enough in other cases

IwanKaramazow:
I might have been prematurely excited by finding something that did not lead to a shift/reduce conflict in the current grammar. 😅Yes, it ~~might~~ will be confusing for newcomers. They already have a lot of stuff to learn. It's better to at least parse the "short form" and print the `module` keyword with refmt.

Issue #1715, opened 2017-12-19
Parser requires extra parens on first-class module destructuring


rickyvetter:
I think this should parse - there may be some conflicts that I'm not aware of and if that's the case I'd love to learn more!
```reason
let foo = (module Foo, baz) => Foo.bar(baz);
```

This version currently parses with no issues.
```reason
let foo = ((module Foo), baz) => Foo.bar(baz);
```

This is closely related to #1605 but the paren wrapping wasn't explicitly called out and I think they can be fixed independently. If it makes sense to merge them then feel free to close this out!

Issue #530, opened 2016-05-24
Better signature annotations of first class modules:


jordwalke:
In OCaml, you can have first class module types be expressed as:

```ocaml
module Modifier = (val (Db.Hashtbl.create ()): Db.Sig with type t = Mods.t)
```

In Reason, they need to be expressed as:
```rust
let module Modifier = (val (Db.Hashtbl.create (): (module Db.Sig with type t = Mods.t)));
```

Another example:
```ocaml
let join_iter (type ta) (type tb) (module A:Sig with type t=ta) (module B:Sig with type t=tb) fn = ...
```

```rust
let join_iter (type ta) (type tb) ((module A): (module Sig with type t=ta)) ((module B): (module Sig with type t=tb)) fn => ...
```

This should be pretty easy to fix. Tagging it "Good First Task".

PR #1946, opened 2018-05-20
Less verbose printing of first class modules in constraint patterns

IwanKaramazow:
Fixes https://github.com/facebook/reason/issues/1942

PR #1949, opened 2018-05-23
Simplify grammar & syntax (parens) of first class modules

IwanKaramazow:
Fixes https://github.com/facebook/reason/issues/530
Fixes https://github.com/facebook/reason/issues/1715
Fixes https://github.com/facebook/reason/issues/1947
Builds on https://github.com/facebook/reason/pull/1946

### Goals
1. Simplify menhir grammar by over-parsing
2. Allow first-class modules with less parentheses

### Result
* Simpler constraints on let-bindings, uppercase automatically implies a `Ptyp_package`
```reason
/* before */
let thing: (module Thing) = (module MyModule);

/* after */
let thing: Thing = (module MyModule);

/* lowercase stays the same, because it could be a core_type */
let thing: (module Foo.lowercaseModType) = (module MyMod);
```

* Less parentheses
```reason
/* before */
let f = ((module Add : S.Z), x) => Add.add(x);
module Modifier = (val (Db.Hashtbl.create (): (module Db.Sig with type t = Mods.t)));

/* after */
let f = (module Add : S.Z, x) => Add.add(x);
module Modifier = (val Db.Hashtbl.create (): module Db.Sig with type t = Mods.t);
```

* `let module` syntax will need to be deprecated, because of too much conflicts.
```reason
/* before */
let module Lowercase = (val y: x);

/* after */
module Lowercase = (val y: x);
```

Still WIP, checking some extra edge cases and finalizing the code.
Also looking for a way to not make this a breaking change and still support the `let module`-syntax.


chenglou:
let module is long deprecated and refmting to `module Foo = ...`. So I think it's an ok time to get rid of it

IwanKaramazow:
@chenglou this is ready.
TLDR
* simplify grammar where possible
* less parens for first class modules (parsing + printing)
* `let thing: Thing = (module MyModule);` parses and *prints* as `let thing: module Thing = (module MyModule);`
* no deprecations, `let module` stays as is.

chenglou:
For `let thing: Thing = (module MyModule)`, is it possible to still not accept it? Only `let thing: module Thing = (module MyModule)`?

I'm not too sure that we're at a stage where folks all use refmt yet

IwanKaramazow:
Sure, no problem, I'll restructure the grammar.

src/reason-parser/reason_parser.mly
```diff
@@ -1061,6 +1061,47 @@ let filter_raise_spread_syntax msg nodes =
         raise Reason_syntax_util.(Error(dotdotdotLoc, (Syntax_error msg)))
     | None -> node
     ) nodes
+
+let err loc s =
+  raise Reason_syntax_util.(
```
chenglou: Do we not have a similar utility already? Genuine question. We should have a single one that throws a single type of exception

src/reason-parser/reason_parser.mly
```diff
@@ -3710,7 +3762,24 @@ mark_position_pat
 pattern_optional_constraint:
 mark_position_pat
   ( pattern                 { $1 }
-  | pattern COLON core_type { mkpat(Ppat_constraint($1, $3)) }
+  | pattern COLON core_type
+    { mkpat(Ppat_constraint($1, $3)) }
+  (* If we kill the `let module …` syntax, this can be placed inside pattern.
```
chenglou: So apparently @jaredly still uses `let module`. Otherwise I think we can kill `let module`?

jaredly:
I super like `let module` because it's one less difference to memorize between statement-level syntax and expression-level syntax.
If we could make `module X = Y` work at the expression level, I'd be fine with removing `let module` everywhere

hcarty:
@jaredly Where does `module X = Y`  it currently work in expressions? I thought that worked anywhere `let module` worked.

jaredly:
aha I guess it does work :P ok, I'd be ok with it going away.

PR #1955, opened 2018-05-25
Punned arguments now ignore those with attributes

### LINKED ISSUE #1909: Formatting of attributes on named arguments (punned) causes parse error.
```
let run = (~dry as [@attr] dry: bool=false, ~modulesMap: string=?, logger) => { 
  
};
```

Formats as:

```
let run = (~[@attr] dry: bool=false, ~modulesMap: string=?, logger) => {
};
```

Which is a parse error.

I think that before detecting if the named argument is "punned" (uses same identifier as the argument name), we should make sure to not consider ones where the identifier has an attribute on it.

vishnugopal:
Punned arguments that have attributes are now ignored by refmt when formatting. This solves a parse error after refmt was run. This is intended to fix #1909

This is my first pull request, please be gentle!

### DIFF
--- a/formatTest/unit_tests/expected_output/uncurried.re
+++ b/formatTest/unit_tests/expected_output/uncurried.re
@@ -174,3 +174,10 @@ let a = foo(. foo(. 3));
 (add(1))(. 2, 3, 4);
 
 (add(1, 2, 3))(. 4);
+
+let run =
+    (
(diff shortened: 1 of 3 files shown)


vishnugopal:
I have signed the CLA now!

jordwalke:
This is really good, thank you.

Issue #1953, opened 2018-05-24
Remove rebuild


chenglou:
It's about time

anmonteiro:
I may be too late to the party, but one useful usage of `rebuild` that I've taken advantage of in the past was to build unikernels in Reason.

I didn't really use `reuild` itself, but pasted the contents of the `myocamlbuild.ml` file in the auto-generated unikernel file with the same name.

hcarty:
@anmonteiro Does dune support unikernel builds yet?

anmonteiro:
@hcarty It's been a year since I last tried; I haven't kept up with Dune, so can't answer that one.

hcarty:
I think the mirage folks have put a fair amount of testing and effort into it. Dune also has Reason support out of the box so it may provide a simpler build environment overall.

PR #1956, opened 2018-05-26
Remove rebuild

IwanKaramazow:
Fixes https://github.com/facebook/reason/issues/1953


chenglou:
@jordwalke

jordwalke:
Cool! Thank you. I'm not sure if anyone is still using `rebuild`, but if so, we can always revert this.

PR #1963, opened 2018-05-27
Don't swallow `<> </>` fragments when formatting

### LINKED ISSUE #870: Formatting makes `<></>` fragment dissapear
```
<div onClick=this##handleClick>
  <>foo</>
</div>
```

Turns into

```
<div onClick=this##handleClick>
  foo
</div>

```

anmonteiro:
fixes #870

### DIFF
--- a/src/reason-parser/reason_pprint_ast.ml
+++ b/src/reason-parser/reason_pprint_ast.ml
@@ -1989,7 +1989,7 @@ let is_punned_labelled_expression e lbl = match e.pexp_desc with
   | _ -> false
 
 let is_punned_labelled_pattern p lbl = match p.ppat_desc with
-  | Ppat_constraint ({ ppat_desc = Ppat_var { txt; _ }; ppat_attributes = _::_ }, _) 
+  | Ppat_constraint ({ ppat_desc = Ppat_var { txt; _ }; ppat_attributes = _::_ }, _)
     -> false
   | Ppat_constraint ({ ppat_desc = Ppat_var { txt; _ }; _ }, _)
   | Ppat_var { txt; _ }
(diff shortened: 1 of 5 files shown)


src/reason-parser/reason_pprint_ast.ml
```diff
@@ -5629,8 +5629,13 @@ let printer = object(self:'self)
     | {pexp_desc = Pexp_constant constant} as x :: remaining ->
       let raw_literal, _ = extract_raw_literal x.pexp_attributes in
       self#formatChildren remaining (self#constant ?raw_literal constant :: processedRev)
-    | {pexp_desc = Pexp_construct ({txt = Lident "::"}, Some {pexp_desc = Pexp_tuple children} )} :: remaining ->
-      self#formatChildren (remaining @ children) processedRev
+    | {pexp_desc = Pexp_construct ({txt = Lident "::"}, Some {pexp_desc = Pexp_tuple children} )} as x :: remaining ->
+      let {jsxAttrs} = partitionAttributes x.pexp_attributes in
+      if jsxAttrs != [] then
```
IwanKaramazow: Another option is the pattern match on `jsxAttrs`.  Pattern matching is probably cheaper than physical equality checks. (Correct me if I'm wrong).

IwanKaramazow:
Can you add some extra tests to verify other expressions?
For example function calls need to be wrapped in braces/parens.

src/reason-parser/reason_pprint_ast.ml
hcarty: Pattern matching should be safer to against future changes in compiler optimizations around equality.

anmonteiro: Hah, I actually had a pattern match at first but decided to go with this approach given it’s used everywhere else that does the same thing.

anmonteiro:
Addressed the comments in the new commits.

chenglou:
Thanks. Will be useful soon I think

PR #1966, opened 2018-05-28
Remove extra space before return type declaration in function definition

### LINKED ISSUE #1628: [refmt] Extra space before colon of return type on function definition
Input:

```
let f = (x: int): int => x;
```

Desired output:

```
let f = (x: int): int => x;
```

Actual output:

```
let f = (x: int) : int => x;
```

Prettier example: https://goo.gl/SuGxvv

anmonteiro:
fixes #1628

### DIFF
--- a/src/reason-parser/reason_pprint_ast.ml
+++ b/src/reason-parser/reason_pprint_ast.ml
@@ -4677,7 +4677,6 @@ let printer = object(self:'self)
         in
         ([makeList
            ~break:IfNeed
-           ~indent:(settings.space * settings.indentWrappedPatternArgs)
            ~inline:(true, true)
            (argsList@[formatJustTheTypeConstraint typeLayout])], e)
       | _ -> (argsList, return)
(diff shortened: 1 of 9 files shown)


src/reason-parser/reason_pprint_ast.ml
```diff
@@ -4675,7 +4675,11 @@ let printer = object(self:'self)
           source_map ~loc:ct.ptyp_loc
             (self#non_arrowed_non_simple_core_type ct)
         in
-        (argsList@[formatJustTheTypeConstraint typeLayout], e)
+        ([makeList
+           ~break:IfNeed
+           ~indent:(settings.space * settings.indentWrappedPatternArgs)
```
chenglou: oh man, settings.space? Have we been using this consistently? cc @IwanKaramazow

anmonteiro: I used the same setting that `wrapCurriedFunctionBinding` employs, given that's where it all trickled down before.

IwanKaramazow: `settings.space` exists and the space setting is used only in function bindings I think. It's pretty obscure, since it defaults to the default `2` here.

chenglou:
Looks good. Thanks!

PR #1969, opened 2018-05-28
Remove extra space before external type definition

anmonteiro:
This is a follow-up to #1966, this time removing the extra space before type
definitions for external declarations.


chenglou:
Thanks

PR #1972, opened 2018-05-29
Parse list and list spread in JSX

### LINKED ISSUE #1467: List and list spread as jsx children cause syntax error
We can ship this after the new release, since nobody uses list as children for now.

Should probably have some error message in the meantime though

anmonteiro:
fixes #1467


src/reason-parser/reason_pprint_ast.ml
```diff
@@ -5637,9 +5637,9 @@ let printer = object(self:'self)
         | ({txt="JSX"; loc}, PStr []) :: _ ->
           begin match self#simplest_expression x with
             | Some r -> self#formatChildren remaining (r :: processedRev)
-            | None -> self#formatChildren (remaining @ children) processedRev
+            | None -> self#formatChildren (children @ remaining) processedRev
```
anmonteiro: this has been broken forever 🙈

chenglou: What's this again?

anmonteiro: Printing of fragments. we wanna print the fragment's children before the fragment's siblings.

anmonteiro: i.e. depth first vs. breadth first

anmonteiro: this test:

```reason
<div> <> ...[ [1,2], [3,4] ] </> </div>
```

was getting formatted to

```reason
<div> <> 1 3 2 4 </> </div>
```

i.e. `1 3 2 4` instead of `1 2 3 4` :)

formatTest/unit_tests/expected_output/jsx.re
```diff
@@ -270,3 +270,5 @@ let x = foo /></ bar;
 <div onClick=this##handleClick>
   <> foo bar </>
 </div>;
+
+<Foo> 1 2 other </Foo>;
```
IwanKaramazow: Small question, am I wrong that I would expect `<Foo> [1, 2] other </Foo>` here?

anmonteiro: yeah I'm not sure about this one. it currently looks to me that lists are always spliced in

src/reason-parser/reason_pprint_ast.ml
IwanKaramazow: Same question as above, I think I expect `<div> <> [1,2] [3,4] </> </div>` here. Am I wrong?

anmonteiro: yeah, I'm not sure. need someone else to weigh in

formatTest/unit_tests/expected_output/jsx.re
chenglou: We should not eagerly cancel out list spread + list wrapping at a syntax level, since there might be ppxes that turns the list literal into something else (e.g. react jsx ppx turns children list into array. So `<Foo> ...[] </Foo>` should stay as such after formatting. Though I recall the implementation for keeping it like that becomes complicated?

anmonteiro: that isn't the case in master however. this is the current behavior:

```
$ echo '<div> ...[] </div>' | refmt
<div />;
```

anmonteiro: Looks like @jordwalke's comment (https://github.com/facebook/reason/pull/1429#issuecomment-334621175) is exactly the behavior that this PR maintains?

Drup:
I rebased and fixed quite a few things in https://github.com/facebook/reason/compare/master...Drup:jsxlist

In particular, inline lists and normal elements can now be interleaved: `<Foo> i [1,2] i [3] </Foo>`.
To allow that, I had to disallow array accesses as JSX children. One must writes `<> (a[2]) </>`. I think it's reasonable.
While I was down there, I factorized the JSX parser quite a lot. It''s shorter with the additions than it was before.

The printer is still completely incorrect, I haven't fixed it yet (and my motivation is wearing a bit :p).
Even in the current release, `refmt` is wrong with lists in JSX:
```
% echo "<Foo> ([1,2]) i </Foo>" | refmt
<Foo> 1 i 2 </Foo>;
```

Issue #1952, opened 2018-05-24
Fix the rest of the spread syntax errors


chenglou:
Thanks to #1925, we now have pretty good wrong list spread errors. These are still remaining:

- list spread pattern
- record spread `{a, ...r}`
- record spread pattern `{a, ...r}`
- array spread
- array spread pattern

As an aside, I'm convinced that #1925's way of fixing it is right (cc @cristianoc @IwanKaramazow). Because if we ever need to support those currently unsupported spread patterns, it'll be easy to unlock that. Simpler code for current maintenance too, so win-win.

IwanKaramazow:
I'll take care of it.

PR #1973, opened 2018-05-29
Super spread errors part 2

IwanKaramazow:
Fixes https://github.com/facebook/reason/issues/1952

-Arrays (expressions + patterns)
-List (expressions + patterns)
-Records (expressions + patterns)


src/reason-parser/reason_parser.mly
```diff
@@ -3330,54 +3352,23 @@ expr_list:
...
       let (es, ext) = match dotdotdot with
       | Some dotdotdotLoc -> (es, Some e)
       | None -> (hd::es, None)
       in
-      let exprList = List.map (fun (dotdotdot, e) -> match dotdotdot with
-      | Some dotdotdotLoc ->
-        (* If the `...` appears in any other location then the last,
-         * show a friendly, clear message explaining the consequences. *)
-        let msg = "Lists can only have one `...` spread, and at the end.
+      let msg = "Lists can only have one `...` spread, and at the end.
```
chenglou: Sorry could you change this one to remove `and`? The comma is probably enough to disambiguate meaning here.

src/reason-parser/reason_parser.mly
```diff
@@ -3330,54 +3352,23 @@ expr_list:
...
 Explanation: lists are singly-linked list, where a node contains a value and points to the next node. `[a, ...bc]` efficiently creates a new item and links `bc` as its next nodes. `[...bc, a]` would be expensive, as it'd need to traverse `bc` and prepend each item to `a` one by one. We therefore disallow such syntax sugar.
-Solution: directly use `concat`."
-        in
-        raise Reason_syntax_util.(Error(dotdotdotLoc, (Syntax_error msg)))
-      | None -> e
-      ) es in
+Solution: directly use `concat`." in
```
chenglou: Could you also change this one too:

```
Solution: directly use `concat` or other List helpers.
```

(I wrote it that way for List and Array because `concat` is shared by stdlib and Belt, but have different meanings. In reality you'd use `List.append` or `Belt.List.concat`, or `List.concat` or `Belt.List.concatMany`.)

src/reason-parser/reason_parser.mly
```diff
@@ -3701,22 +3702,42 @@ mark_position_pat
 ;
 
 %inline pattern_comma_list:
-  lseparated_nonempty_list(COMMA, pattern) { $1 };
+  lseparated_nonempty_list(COMMA, opt_spread(pattern))
+  { let msg = "Pattern matching on arrays can not have a `...` spread. It would allocate a whole new array every time." in
```
chenglou:
```
Array's `...` spread is not supported in pattern matches.
Explanation: such spread would create a subarray; out of performance concern, our pattern matching currently guarantees to never create new intermediate data.
Solution: if it's to validate the first few elements, use a `when` clause + Array size check + `get` checks on the current pattern. If it's to obtain a subarray, use `Array.sub` or `Belt.Array.slice`.
```

src/reason-parser/reason_parser.mly
```diff
@@ -3701,22 +3702,42 @@ mark_position_pat
...
-  | pattern_comma_list COMMA?                  { ($1, None)    }
+  lseparated_nonempty_list(COMMA, opt_spread(pattern)) COMMA?
+  { match List.rev $1 with
+    (* spread syntax is only allowed at the end *)
+    | ((dotdotdot, p) as hd)::ps ->
+      let (ps, spreadPat) = match dotdotdot with
+      | Some dotdotdotLoc -> (ps, Some p)
+      | None -> (hd::ps, None)
+      in
+      let msg = "Pattern matching on a list can only have one `...` spread, and at the end" in
```
chenglou:
```
List pattern matches only supports one `...` spread, at the end.
Explanation: a list spread at the tail is efficient, but a spread in the middle would create new list(s); out of performance concern, our pattern matching currently guarantees to never create new intermediate data.
```

PR #1984, opened 2018-06-04
Fix `module type of` formatting

anmonteiro:
fixes https://github.com/facebook/reason/issues/1614

~This is definitely a little hacky and special-cased, but it also looks like we _want_ to special case this particular pattern.~

EDIT: Not as hacky now, and it prints much nicer.

### DIFF
--- a/formatTest/unit_tests/expected_output/syntax.rei
+++ b/formatTest/unit_tests/expected_output/syntax.rei
@@ -28,3 +28,9 @@ let myRecordWithFunctions: adders;
  * Public result.
  */
 let result: int;
+
+/* https://github.com/facebook/reason/issues/1614 */
+module Event:
+  (module type of {
+     include ReactEventRe;
+   });
--- a/formatTest/unit_tests/expected_output/syntax.rei
+++ b/formatTest/unit_tests/expected_output/syntax.rei
@@ -30,7 +30,6 @@ let myRecordWithFunctions: adders;
 let result: int;
 
 /* https://github.com/facebook/reason/issues/1614 */
-module Event:
-  (module type of {
-     include ReactEventRe;
-   });
+module Event: (module type of {
+  include ReactEventRe;
+});
--- a/src/reason-parser/reason_pprint_ast.ml
+++ b/src/reason-parser/reason_pprint_ast.ml
@@ -4236,6 +4236,19 @@ let printer = object(self:'self)
     let letPattern = (label ~space:true (atom labelOpener) bindingPattern) in
     (formatTypeConstraint letPattern typeConstraint)
 
+  method formatModuleTypeOfSignatureBinding labelOpener bindingPattern typeConstraint =
+    let letPattern = makeList
+        ~break:IfNeed
+        ~inline:(true, true)
+        [(atom labelOpener);
+         makeList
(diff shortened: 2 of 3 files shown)


src/reason-parser/reason_pprint_ast.ml
```diff
@@ -4120,6 +4120,19 @@ let printer = object(self:'self)
...
+        ~inline:(true, true)
+        [(atom labelOpener);
+         makeList
+           ~pad:(true, false)
+           [bindingPattern; atom ":"]]
+    in
+    makeList
+      ~inline:(true, true)
+      ~break:Always
+      [letPattern; makeList ~wrap:("  ", "") [typeConstraint]]
```
IwanKaramazow:
```ocaml
makeList
  ~inline:(true, true)
  ~break:IfNeed
  ~postSpace:true
  [letPattern; typeConstraint]
```
I think you want `~postSpace:true` instead of `~wrap:("  ", "")`.
`postSpace` inserts a space after every separator between items in a list.
If the separator is empty (i.e. empty string), it still inserts a simple space.

formatTest/unit_tests/expected_output/syntax.rei
```diff
@@ -28,3 +28,9 @@ let myRecordWithFunctions: adders;
  * Public result.
  */
 let result: int;
+
+/* https://github.com/facebook/reason/issues/1614 */
+module Event:
```
IwanKaramazow: Ideally I would expect, which might be a bit more work...
```
module Event: (module type of {
  include ReactEventRe;
});
```

src/reason-parser/reason_pprint_ast.ml
anmonteiro: I tried that but `~inline:(true, true)` makes it be 0-based indented... `wrap` here is basically a hack

formatTest/unit_tests/expected_output/syntax.rei
anmonteiro: that'll definitely be a lot more work. 

I can work on that if that's the consensus on how we want this to look like.

src/reason-parser/reason_pprint_ast.ml
anmonteiro: I tried that too. This was the only thing I arrived at that worked 😢 

There's a case where we don't need the `wrap` hack but then if there's a preceding comment if formats like:

```reason
/* https://github.com/facebook/reason/issues/1614 */

    module Event:
    (module type of {
       include ReactEventRe
     })
```

formatTest/unit_tests/expected_output/syntax.rei
chenglou: @IwanKaramazow Is this the same problem as the other function formatting you were working on?

```reason
let a = (
  b,
  c,
) => ...
```

IwanKaramazow: It isn't, the function formatting is a separate thing, that needs lots of work.

IwanKaramazow:
Thanks for the additional improvements!

@chenglou merge approved

chenglou:
Thanks!

Issue #1958, opened 2018-05-26
Extra line between recursive bindings


chenglou:
```reason
let anotherRegularLetBinding = "hi"

and yetAnotherBinding = "bye";
```

refmt this removes the line. I guess this is desired right? We can still be a bit opinionated for whitespace when it makes sense. If so, I'll close the issue.

cc @IwanKaramazow @jordwalke

IwanKaramazow:
It's not yet implemented for recursive let bindings, probably should add it.

chenglou:
Should we? No opinion, except unless this makes the logic even harder

IwanKaramazow:
Not hard, just need to implement it.

PR #1990, opened 2018-06-06
interleave whitespace between recursive let and module bindings

IwanKaramazow:
Fixes https://github.com/facebook/reason/issues/1958

### DIFF
--- a/esy.json
+++ b/esy.json
@@ -32,6 +32,7 @@
     "install": [
       ["esy-installer", "reason.install"],
       ["esy-installer", "rtop.install"]
-    ]
+    ],
+    "buildsInSource": "_build"
   }
 }
--- a/esy.json
+++ b/esy.json
@@ -32,7 +32,6 @@
     "install": [
       ["esy-installer", "reason.install"],
       ["esy-installer", "rtop.install"]
-    ],
-    "buildsInSource": "_build"
+    ]
   }
 }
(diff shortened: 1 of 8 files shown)


chenglou:
Should we even allow whitespace between recursive let bindings? As much as possible I'd still like us to be opinionated on whitespace

IwanKaramazow:
I think we should for consistency, it would be a little bit weird if we didn't allow it here.
Imagine rewriting the Reason printer with recursive functions instead of using the object system:
```
let rec printExpression = (e) => beatifulPrinting(e)

and printPattern = (pat) => ...

and printModule = (m) => ...
```
Some space between the different bindings gives a little more room to breath

chenglou:
Would this actually happen though? Because the reality would be:

```reason
let rec printExpression = (e) => {
  beatifulPrinting(e)
  ...
}
and printPattern = (pat) => {
  ...
}
and printModule = (m) => {
   ...
}
```

Right?

IwanKaramazow:
Yes, but someone might want the right the following:
```reason
let rec main = (e) => {
  ...
}

and helper1 = (_) => {
  ...
}
and helper2 = (_) => {
   ...
}
and helper3 = (_) => {
   ...
}
```
`helper1`, `helper2`, `helper3` could be grouped together. Visually it becomes very clear that what the relation between the `helper{X}` and `main` is.

I might be completely wrong though, no opinion, and can close the PR.

jaredly:
I have a mild preference for keeping recursive bindings together

PR #1998, opened 2018-06-10
for_completion: handle end of token situation

let-def:
When completing, `ocamlmerlin-reason` inserts an identifier if there is not one yet under the cursor.

The rationale is that completion will introduce an identifier in the buffer. So when starting completion on whitespace, an empty identifier is introduced.
This let the parser produce an AST as if the completion was already done. Then merlin can use the type information around this identifier to guide completion.

Unfortunately, the current code doesn't handle the case where the cursor is right at the end of an existing identifier: it will introduce a second one, confusing the parser.

### DIFF
--- a/src/reason-parser/reason_lexer.mll
+++ b/src/reason-parser/reason_lexer.mll
@@ -1024,14 +1024,17 @@ and skip_sharp_bang = parse
   let completion_ident_pos = ref Lexing.dummy_pos
 
   let token lexbuf =
-    let before = lexbuf.Lexing.lex_curr_p.Lexing.pos_cnum in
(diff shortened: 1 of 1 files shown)


jordwalke:
Thanks for the fast fix, @let-def. I'll test it out.

jordwalke:
This is feeling pretty great so far! I'll merge after another day of testing, then cut a new release of all the tools.

chenglou:
No release yet please; see #2003

jordwalke:
Results of more testing: This is a huge improvement in local merlin editing. I think the latest release regressed the quality of merlin, even though the quality of editing increased. This seems to fix it. No new bugs in completion as far as I've been able to tell.

Issue #2011, opened 2018-06-19
Can the requirement to put a semicolon after an if/else block be relaxed?


bordoley:
I'm not familiar enough with the grammar  to know if this is possible or not, but if possible it would be really nice if the following code compiled:

```
let test = b => {
  if (b) {
    ignore();
  }
  ignore();
};
```
Notice, the missing semicolon after the if block. I realize this is somewhat of a dubious request, but its a frequent source of hard to find compile bugs (out of habit) and I imagine hits newcomers to reason from JS/Java/C#/etc. quite often.

I'd suggest only relaxing this constraint at compile time, but adding the semicolon back in when printing. If this is not possible due to grammar restrictions, feel free to close.

IwanKaramazow:
It’s on my todo list. This is a small case where the semi colon recovery fails, we can solve it at the grammar level

bordoley:
well that was fast. thanks!

chenglou:
Iwan's getting pretty good at this now lol

PR #2012, opened 2018-06-19
Relax semicolon requirement in more cases

IwanKaramazow:
Fixes https://github.com/facebook/reason/issues/2011

Extends the approach of https://github.com/facebook/reason/pull/1887, to allow semi-less reason in the context of:
- if
- while
- for
- switch
- for
- lazy
- assert

@let-def is this ok? Or am I opening Pandora's box?

### DIFF
--- a/formatTest/unit_tests/expected_output/modules_no_semi.re
+++ b/formatTest/unit_tests/expected_output/modules_no_semi.re
@@ -597,3 +597,33 @@ let foo =
 
 let f =
     (module M: M with type x = x and type y = y) => M.x;
+
+let test = b => {
+  if (b) {
+    ignore();
+  };
+
+  while (x) {
+    compute();
+  };
+
+  try (x()) {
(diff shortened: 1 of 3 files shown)


let-def:
It depends on the future features we might like to see in the language. But since, for now, it is just being more relaxed about input accepted while still generating strict code, I am not too worried.

This idea of allowing no semi-colon before `LIDENT`/`UIDENT` might have some unexpected consequences I cannot foresee though.

chenglou:
Alright then. Merging =)

jordwalke:
Yeah, @let-def the expectations we set about relaxed semicolons is that it's an IDE time assistant to fixing missing semicolons. I think that "nonbreaking changes" should be defined as changes that will not break your code after formatting in v1, then parsing in v2. "nonbreaking changes" make no promise about unformatted code in v1 being parsed in v2.

Issue #2017, opened 2018-06-21
refmt prints invalid Reason code for annotated polymorphic variant values when parsing OCaml syntax


hcarty:
```
$ echo 'let x : t = `Poly' | refmt --parse=ml
let x:  .t = (`Poly: t);
```
The type constraint is duplicated and gets a leading `.` on the lhs of the `=`, leading to invalid syntax.
Reason 3.2.0 installed via opam.

IwanKaramazow:
Hmm, this a serious bug. I'm fixing it, unacceptable.

PR #2019, opened 2018-06-22
Fix printing of Ptyp_poly with an empty string list (Ppat_constraint).

IwanKaramazow:
Fixes https://github.com/facebook/reason/issues/2017

There's a bug in Ocaml 4.06 resulting in
``let x: t = `Poly``
being parsed as
``let x: t = (`Poly: t)``

The constraint is duplicated.
See https://caml.inria.fr/mantis/view.php?id=7758 and https://caml.inria.fr/mantis/view.php?id=7344
for more background.

The bug in our printer ``let x : .t = (`Poly: t)`` (notice the erroneous dot) has been fixed.


chenglou:
Thanks

PR #2031, opened 2018-06-29
restore ocaml syntax printing in rtop

matthiaskern:
`#toggle_syntax` now also toggles printing to ocaml

cc @ryyppy

### DIFF
--- a/src/rtop/reason_utop.ml
+++ b/src/rtop/reason_utop.ml
@@ -8,6 +8,16 @@
  * LICENSE file in the root directory of this source tree.
  *)
 
+module ToploopBackup = struct
+  let print_out_value = !Toploop.print_out_value
+  let print_out_type = !Toploop.print_out_type
+  let print_out_class_type = !Toploop.print_out_class_type
+  let print_out_module_type = !Toploop.print_out_module_type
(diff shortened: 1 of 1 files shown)


src/rtop/reason_utop.ml
```diff
@@ -67,7 +87,8 @@ let init_ocaml () =
   UTop.parse_toplevel_phrase := UTop.parse_toplevel_phrase_default;
   UTop.parse_use_file := UTop.parse_use_file_default;
   UTop.history_file_name :=
-      Some (Filename.concat LTerm_resources.home ".utop-history")
+      Some (Filename.concat LTerm_resources.home ".utop-history");
+  restore_toploop_printing ()
```
ryyppy: Is there any reason why `restore_tooploop_printing` needs to be a function? might as well just drop the printer wiring logic in here (like in `init_reason`)

ryyppy:
Except the one thing I pointed out, this whole thing looks good to me... @jordwalke what do you think?
This would ensure proper result printing in ocaml syntax whenever the user is in OCaml mode!

cristianoc:
Thanks!

PR #2033, opened 2018-06-30
Only wrap `Pexp_function` in parens when necessary

### LINKED ISSUE #2032: Unnecessarily wraps an anonymous fun function in parens
```
    let predicate =
      predicate === Functions.alwaysTrue1 ?
        defaultPredicate :
        fun
          | None => false
          | Some(exn) => predicate(exn);
```
gets refmt'd to:

```
 let predicate =
      predicate === Functions.alwaysTrue1 ?
        defaultPredicate :
        (
          fun
          | None => false
          | Some(exn) => predicate(exn)
        );
```

Notice the additional unnecessary parens.

anmonteiro:
fixes #2032

before:

```reason
let predicate =
  predicate === Functions.alwaysTrue1 ?
    defaultPredicate :
    (
      fun
      | None => false
      | Some(exn) => predicate(exn)
    );
```

after:

```reason
let predicate =
  predicate === Functions.alwaysTrue1 ?
    defaultPredicate :
    fun
    | None => false
    | Some(exn) => predicate(exn);
```

### DIFF
--- a/src/reason-parser/reason_pprint_ast.ml
+++ b/src/reason-parser/reason_pprint_ast.ml
@@ -493,6 +493,8 @@ let namedArgSym = "~"
 let requireNoSpaceFor tok =
   tok = fastPipeToken || (tok.[0] = '#' && tok <> "#=")
 
+let funToken = "fun"
+
 let getPrintableUnaryIdent s =
   if List.mem s unary_minus_prefix_symbols ||
      List.mem s unary_plus_prefix_symbols
@@ -621,6 +623,9 @@ let rules = [
     (TokenPrecedence, (fun s -> (Left, s.[0] == '|' && not (s = "||"))));
     (TokenPrecedence, (fun s -> (Left, s.[0] == '&' && not (s = "&") && not (s = "&&"))));
     (TokenPrecedence, (fun s -> (Left, s.[0] == '$')));
+  ];
+  [
+    (CustomPrecedence, (fun s -> (Left, s = funToken)));
   ];
   [
     (TokenPrecedence, (fun s -> (Right, s = "&")));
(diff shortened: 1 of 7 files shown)


src/reason-parser/reason_pprint_ast.ml
```diff
@@ -5603,9 +5606,6 @@ let printer = object(self:'self)
                 match x'.pexp_desc with
                 | Pexp_let _ ->
                   Some (makeLetSequence (self#letList x))
-                | Pexp_function l when (pipe || semi) ->
```
anmonteiro: this match was useless. `Pexp_function` would be caught by `expression_immediate_extension_sugar`

jordwalke:
I've found it very difficult to prove that it is safe to omit parenthesis around functions when they are embedded in infix expressions, and when they return infix function applications. To merge this I think we want to have a lot more test cases where the function returns something that applies infix `x >>= y`, and also when the function itself is the right hand/left hand side of similar infix. (Also, your ternary test is good too - let's combine it with a bunch of the infix tests I mentioned as well).

anmonteiro:
@jordwalke thanks for the feedback.

In coming up with test cases, the following example does indeed print incorrectly with the current changes:

```reason
(fun
| None => x >>= y
| Some(x) => x >>= y)
>>= bar ;

/* becomes this, which doesn't pass idempotency */
fun
| None => x >>= y
| Some(x) => x >>= y
>>= bar;
```

I wonder if we can play with the precedence rues to make it work?

anmonteiro:
@jordwalke I think I added your suggested test cases, let me know if you'd like to see more (and which).

The latest changes introduce a custom `"fun"` precedence that's lower than e.g. bind operators (`>..`) but higher than the ternary operators, making `Pexp_function` print without paren wrapping in the ternary case (which is what the original issue is suggesting).

IwanKaramazow:
@chenglou I don't immediately see anything wrong with this PR, let's merge?

chenglou:
Thanks. Though I’ll hold off meeting for a day or two. New release now

chenglou:
Thanks!

PR #2034, opened 2018-06-30
add specific test file for OCaml 4.06.1

anmonteiro:
This is only useful for local development on the 4.06.1 switch, so I understand
if you don't want to get it in. It seems we can't move the CI to run on 4.06.1
for now because there isn't a 4.06.1 Docker image yet.


IwanKaramazow:
@chenglou can be merged.

Issue #1828, opened 2018-02-19
Comments at end of line for polymorphic variants are duplicated on refmt.


jordwalke:
Master branch:
```
  type widget_state = [
    | `DEFAULT /* here */
    | `HOVER
    | `ACTIVE
  ];
```

`refmt` leaves:

```
  type widget_state = [
    | `DEFAULT /* here */
    | `HOVER  /* here */
    | `ACTIVE  /* here */
  ];
```

And it causes the duplication to grow each refmt.

emnh:
I was about to report this bug also, but then I saw it was already there 👍

jordwalke:
@emnh did you only start seeing this in the master branch or 3.0.5? How are you testing that version? What kind of project?

emnh:
I saw it on https://gist.github.com/Octachron/4e833a22844fd90cd6d15b0af927dab0 . My project is https://github.com/emnh/rts-reasonml and I use:

emh@rainmaker ~ 0% refmt --version                                                                                                                                                                                           (git)-[master]
Reason 3.0.4 @ 731e605

jaredly:
Still an issue ([playground](https://reason.surge.sh/?syntax=Reason&code=ATAuE8AcFNgdwJYBMDm1QH0DOoCGpYBeYAbQCgQQAfYAAwBEBRAMQEEBVAGQBVgB6AFTAAFtABOsAXwqUatABIB5AGqMASjOp1WAYW4BJVTIC6AbiA&canvas=200))

IwanKaramazow:
For future reference:

The problem is that the location of the specific row fields isn't tracked in the omp 4.04 ast we're currently using. Ocaml 4.06 does contain the location of the row fields. Due to the fact there's no location the comments think that they need to be attached to `type widget_state = [ ... ]`, which results in duplication. The comment interleaving algorithm doesn't know that it needs to attach them to the rows, no location to indicate so. Upgrading the parser to use 4.06 will fix this. (Not a small task)

amiralies:
@anmonteiro is your branch going to be merged?

glennsl:
This is fixed in 3.6.0, thankfully, because the problem increased exponentially with the number of comments, as shown in the diff below:

<details><summary>Insane refmt diff</summary><p>

```diff
   type pathCommand = [
-    // MoveTo
-    | `M(float, float) // x, y
-    | `m(float, float) // dx, dy
-    // LineTo
-    | `L(float, flaot) // x, y
[...]

PR #2036, opened 2018-06-30
Upgrade Parser to 4.06 AST

anmonteiro:
This PR upgrades the parser to the 4.06 AST version.

This paves the way to solving a number of issues that are currently blocked because we use the 4.04 version of the AST. For example, this PR also fixes:

- #1828 
- #2302 

In addition, this PR introduces parsing / printing of some new syntactic constructs made possible by OCaml 4.06:

- [x] `open` statements in class _expressions_ (e.g. `class x = { open M; as self; };` (https://github.com/ocaml/ocaml/pull/1249)
- [x] `open` statements in class _types_ (e.g. `class type x = { open M; as 'a; };` (also https://github.com/ocaml/ocaml/pull/1249)
- [x] Inherited field in object type (https://github.com/ocaml/ocaml/pull/1118). This is the current syntax for inherited fields in object types:

```rust
type x = {. a: int};
type y = {
  .
  b: string,
  ...x, /* <-- NEW */
};
```


src/reason-parser/reason_parser.mly
```diff
@@ -1397,7 +1400,6 @@ conflicts.
 %on_error_reduce structure_item let_binding_body
                  as_loc(attribute)+
                  type_longident
-                 attribute+
```
anmonteiro: Menhir complains these are unreachable from a toplevel expression. Probably because there's `as_loc(attribute)+` above.

anmonteiro:
Tests are now failing on versions of OCaml < 4.06 because I implemented parsing of `Pcl_open` (e.g. `class x = { open M; as self; };`.

Not sure what's the proper way to parse these on older versions of OCaml; or is it even supported?

hcarty:
It was only added in 4.06.0. See https://github.com/ocaml/ocaml/blob/trunk/Changes#L801

anmonteiro:
@hcarty I'm aware. I'm trying to see if there's a way to run those tests only against 4.06.0.

anmonteiro:
I can't find a way to do it with the current setup. In any case, after thinking about it, I don't see the value of exercising the OCaml parser, since it's something we don't control. I'll test `Pcl_open` and the new AST types in `.re` files.

hcarty:
@anmonteiro I apologize - I misunderstood the question

anmonteiro:
I think this PR is now ready. Code reviews appreciated!

src/reason-parser/reason_oprint.ml
```diff
@@ -444,8 +444,8 @@ and print_simple_out_type ppf =
           Ovar_fields fields ->
             print_list print_row_field (fun ppf -> fprintf ppf "@;<1 -2>| ")
               ppf fields
-        | Ovar_name (id, tyl) ->
-            fprintf ppf "@[%a%a@]" print_typargs tyl print_ident id
+        | Ovar_typ o -> print_out_type ppf o
+            (* fprintf ppf "@[%a%a@]" print_typargs tyl print_ident id *)
```
IwanKaramazow: Commented out line can probably be omitted?

src/reason-parser/reason_parser.mly
```diff
@@ -2001,6 +2003,8 @@ mark_position_cl
   | let_bindings SEMI class_expr_lets_and_rest
     { class_of_let_bindings $1 $3 }
   | object_body { mkclass (Pcl_structure $1) }
+  | LET? OPEN override_flag as_loc(mod_longident) SEMI class_expr_lets_and_rest
```
IwanKaramazow: Is there a reason why we still need to support `let open ...` vs just `open ...` in newly added syntax? I tend to favour just `open` in favour of less grammar complexity. Thoughts?

src/reason-parser/reason_parser.mly
```diff
@@ -2289,17 +2293,24 @@ mark_position_cty
...
+      let ct = mkcty ~loc $3 in
       {ct with pcty_attributes = [uncurry_payload loc]}
     }
   | as_loc(LBRACE) class_sig_body as_loc(error)
     { unclosed_cty (with_txt $1 "{") (with_txt $3 "}") }
 ;
 
+class_sig_body_cty:
+  | class_sig_body { Pcty_signature $1 }
+  | LET? OPEN override_flag as_loc(mod_longident) SEMI as_loc(class_sig_body_cty)
```
IwanKaramazow: Same remark about `let open` as above.

src/reason-parser/reason_pprint_ast.ml
```diff
@@ -2968,7 +2975,7 @@ let printer = object(self:'self)
             match rf with
               | Rtag (label, attrs, opt_ampersand, ctl) ->
                 let pcd_name = {
-                  txt = label;
+                  txt = label.txt;
                   loc = pcd_loc;
```
IwanKaramazow: This should probably become `label.loc`? (And finally fixes the comment duplication?)

anmonteiro:
Went through the code review items and:

- added typechecked tests for the new AST constructs
- fixed #1828 now that we have locations of each poly variant in the AST (thanks @IwanKaramazow!)

jordwalke:
VERY cool.

anmonteiro:
superseded by #2487

PR #2040, opened 2018-07-01
Parse class exprs and types without trailing semicolon

### LINKED ISSUE #2037: Class expression doesn't parse without trailing semicolon
This should be consistent with block declarations.

This parses:

```reason
class x = { 
  as self; 
};
```

this doesn't:

```reason
class x = { 
  as self
};
```

anmonteiro:
fixes #2037


anmonteiro:
Addressed.

chenglou:
Thanks!

PR #2042, opened 2018-07-02
Punning in Js.t sugar (string literal keys)

### LINKED ISSUE #2039: Punning in Js.t sugar (string keys)
The following is a syntax error:

```rust
type x = {.
  "a": string,
  u
};
```

I would expect it to parse into (i.e. punning):

```rust
type x = {.
  "a":string,
  "u": u
};
```

anmonteiro:
fixes #2039


IwanKaramazow:
I'm not entirely sure if we should proceed with this feature. Punning is great, but mixing quoted and normal identifiers, might be a bit weird. 

A couple of questions I'm asking myself:
Why do we want to add this? Is it better than what we already have? Can it be taught to beginners? How do we explain it?

anmonteiro:
This PR is definitely meant to be discussed; I'm just putting it out there to see if people want this feature or not.

Input very much appreciated!

jaredly:
huh yeah I think this might be more confusing than useful. 😅

PR #2044, opened 2018-07-02
Print (foo^)##bar with the right precedence

### LINKED ISSUE #1554: Hard time parsing (curNode^)##childNodes and wrong refmt
It seems like `curNode^##childNodes` isn't valid, so I thought of using parens which works but causes refmt to barf:

```
(curNode^)##childNodes
```

turns into

```
curNode^##childNodes
```
which gives the error:
screen shot 2017-10-26 at 23 10 25

and then gets turned into
```
curNode ^## childNodes
```

anmonteiro:
fixes #1554

### DIFF
--- a/formatTest/unit_tests/expected_output/basicStructures.re
+++ b/formatTest/unit_tests/expected_output/basicStructures.re
@@ -830,3 +830,6 @@ it("should remove parens", a => {
   print_string("did it work?");
   print_string("did it work?");
 });
+
+/* https://github.com/facebook/reason/issues/1554 */
+(curNode^)##childNodes;
--- a/formatTest/unit_tests/input/basicStructures.re
+++ b/formatTest/unit_tests/input/basicStructures.re
@@ -681,3 +681,6 @@ it("should remove parens", (a) => {
   print_string("did it work?");
   print_string("did it work?");
 });
+
+/* https://github.com/facebook/reason/issues/1554 */
+(curNode^)##childNodes;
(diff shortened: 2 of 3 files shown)


IwanKaramazow:
`method simple_enough_to_be_lhs_dot_send ?(except_unary_postfix=false)`
The `except_unary_postfix` name arg feels very specific/specialized. Can we pass the relevant expression down to make it a bit more generic/extendable in the future? There might be other cases.
I might be totally wrong though, this is just an idea.

anmonteiro:
I agree it feels too specialized. I'm definitely open to other ideas, but the point here is in the presence of a `SHARPOP` we want to `formatPrecedence` of `UnaryPostfix`.

Maybe the optional argument is fine and just needs to be changed to `sharpop`?

IwanKaramazow:
random idea, what about passing `?infixOp` as some kind of context?

anmonteiro:
@IwanKaramazow check my latest commit. I replaced `except_unary_postfix` with an `infix_context` that we can pattern match against. 

I liked your idea, and I believe this might prove to be more extensible in the future if we wanna tweak other infix ops.

What do you think?

IwanKaramazow:
Feels like a nice step into the right direction :)
If you want to avoid the extra optional allocation, you can use the empty string as default value `?(except_unary_postfix="")`. The current approach is fine, won't make much of a difference here.

@chenglou ready for me, merge approved.

anmonteiro:
I thought about the empty string but it didn't feel very nice to me. I'll change it if you'd like though

IwanKaramazow:
No, not necessary, option makes sense here

chenglou:
Thanks

PR #2045, opened 2018-07-03
Fix Bigarray syntax

### LINKED ISSUE #2038: Bigarray syntax does not refmt cleanly
The following fails with a syntax error:
```
echo "a.{1, 2}" | refmt --parse=ml | refmt
```
with an intermediate output of `a.{1 2};`.

anmonteiro:
fixes #2038


hcarty:
Thanks!

IwanKaramazow:
1. (Parser) The grammar could be a bit cleaner with the following approach:
```ocaml
%inline bigarray_access: 
  DOT LBRACE lseparated_nonempty_list(COMMA, expr) COMMA? RBRACE { $3 }

  (* down the road you can write*) 
…
| E bigarray_access
  
…
 | simple_expr bigarray_access EQUAL expr
```
The %inline keyword causes all references to bigarray_access to be replaced with its definition. Inlining can also help recover a slight efficiency margin. I don’t think that we need an extra state here. Without the %inline keyword, the language recognized by the grammar would be the same, but the LR automaton would probably have one more state and would perform one more reduction at run time.

2. (Printer) What happens when the code line breaks for gets and sets? Can you add an extra test? `x.{superlongthing, anothersuperongthing, etc}`
3.  Can the tests be moved to `bigarray.re` (or something similar)? `syntax.re` is becoming a massive amount of random syntax.
4. Question: do we need type checked tests ?

anmonteiro:
@IwanKaramazow 

1. Great suggestion. I thought this was going to result in shift/reduce conflicts but apparently I was wrong. Changed!
2. Added a test for this. It _doesn't_ look pretty. I'll fix the printer for these cases later. How would you expect the examples to look (especially the assignment one)?
3. & 4. moved the tests to the typechecked folder which will also `refmt` them!

IwanKaramazow:
Ideally we want:
```
bigThing.{
 1,
 2,
 3,          --> notice the trailing comma
}

/* should be fairly easy with the classic */
let k = makeList [atom "bigThing"; atom "."] in
let v = makeList ~break:IfNeed ~postSpace:true ~sep:commaTrail ~wrap:("{", "}") [....] in
label k v
```

Not sure how easy it is for `set`, to get the same layout:
```
bigThing.{
  1,
  2,
  3,
} = 3.0
```

IwanKaramazow:
Nice work :)

@chenglou ready to merge.

chenglou:
Thanks

PR #2048, opened 2018-07-04
Recover custom error messages in Reason_parser_explain

### LINKED ISSUE #1643: Have record values' fields error on reserved keywords too
Repro:

```reason
let foo = {type: 1};
```

This gives the generic `<UNKNOWN SYNTAX ERROR>`, though it should give the ones we give to record types: `type is a reserved keyword, it cannot be used as an identifier. Try `type_' instead`

cc @let-def

anmonteiro:
fixes #1643


IwanKaramazow:
While we're add it, what about object keys?

anmonteiro:
Aren't those valid? I was gonna do it but this parses without errors:

```rust
let x = {"type": 3};
```

IwanKaramazow:
I mean objects from the ocaml oo system: `let x = {. type: 1}` and `let x = {.. type: 1}`

chenglou:
cc @let-def

anmonteiro:
@chenglou @IwanKaramazow I just reworked this PR to use the infrastructure that we have.

Turns out we were ignoring every custom error since we started committing the error message table again 🙈.

Moving the table lookup to the end of the function makes all our custom error messages work again.

Issue #774, opened 2016-09-19
(foo##bar).baz prints wrongly


chenglou:
Prints as `foo##bar.baz`

Error: `Js object ## expect syntax like obj##(paint (a,b))`

@bsansouci @Schmavery

jordwalke:
Hmmm the printer thinks that dot is lower precedence than ## but the parser knows that dot is highe precedence than ## it seems.

thangngoc89:
Here is a slightly different version with the same error: 

```
let foo = {|{"Uppercase": "foo"}|};

let parsed = foo |> Js.Json.parseExn;

Js.log(parsed##Uppercase);
```

anmonteiro:
I think this was fixed in #2050

PR #2050, opened 2018-07-05
Fix SHARPOP precedence

### LINKED ISSUE #1507: `location##streets[1]` precedence is wrong
`(location##streets)[1]` is the right precedence. But it formats to `location##streets[1]`. The first form is the desirable precedence, but the latter form is the desirable visual.

anmonteiro:
fixes #1507 #774

Not really sure how to write tests for this one.

I verified the following expressions manually:

```shell
$ echo 'foo##bar.baz' | ./_build/install/default/bin/refmt -p ml
;;(foo ## bar).baz

$ echo 'foo##bar[1]' | ./_build/install/default/bin/refmt -p ml
;;(foo ## bar).(1)

$ echo 'foo##bar[0]##baz[1]' | ./_build/install/default/bin/refmt -p ml
;;(((foo ## bar).(0)) ## baz).(1)
```


anmonteiro:
Just fixed the issue pointed out in my last comment with the latest commit.

src/reason-parser/reason_parser.mly
```diff
@@ -1209,6 +1209,7 @@ let package_type_of_module_type pmty =
 %token SEMISEMI
 %token SHARP
 %token <string> SHARPOP
+%token SHARPOP_WITH_EQUAL
```
IwanKaramazow: can you rename this to `SHARPEQUAL`? `SHARPOP_WITH_EQUAL` conveys `# with = `, if you look at the other token names.

anmonteiro: definitely

src/reason-parser/reason_lexer.mll
```diff
@@ -481,10 +481,15 @@ rule token = parse
   | "#=<" {
     (* Allow parsing of foo#=<bar /> *)
     set_lexeme_length lexbuf 2;
-    SHARPOP("#=")
+    SHARPOP_WITH_EQUAL
   }
   | "#" operator_chars+
-      { SHARPOP(lexeme_operator lexbuf) }
+      { let l = lexeme_operator lexbuf in
+        if l = "#=" then
```
IwanKaramazow: I wonder if we should add `| "#=" { SHARPEQUAL }` for clarity in the lexer. No hard opinions here.

chenglou:
M-m-merged!

PR #2051, opened 2018-07-05
Syntax sugar `["literal_string"]` for SHARPOPs `##` and `#=`

anmonteiro:
This is a proposal for `["string_literals"]` sugar for `##` and `#=`, stemming from a discussion with @chenglou.

The change is fully backwards compatible, meaning `foo##bar` will still parse just fine. `foo["bar"]` parsed before too but it's invalid array access, and without this PR it will result in a type error anyway.

Before:

```rust
foo##bar;

foo##bar #= baz;

/* slightly contrived: */
foo##bar##baz #= bar##baz[0];
```

After:

```rust
foo["bar"];

foo["bar"] = baz;

/* slightly contrived: */
foo["bar"]["baz"] = bar["baz"][0];
```

If everyone's on board, I suggest merging #2050 before this PR, so that we'll get some nice precedence fixes for free (see my comments inline below).

**TODOs**:

- [x] Currently, `#=` is only printed as `=` if the left expr is a chain of `##` applications. This can be relaxed to account for e.g. `foo##bar[0]##baz[2] = "1px black dashed";`, i.e. array access. 
  ~~- This is only useful once #2050 is in. I have a local branch with #2050 and this PR and it works great.~~ Done
- [x] Think about relaxing it as well for `.` record accesses

### DIFF
--- a/formatTest/unit_tests/expected_output/bucklescript.re
+++ b/formatTest/unit_tests/expected_output/bucklescript.re
@@ -20,20 +20,20 @@ x##y##z #= xxxx##yyyy##zzzz;
 let result =
   js_method_run1((!react)#createElement, foo);
 
-add(zz##yy, xx##ww);
+add(zz["yy"], xx["ww"]);
 
 /* These should print the same */
-let res = x##y + z##q; /* AST */
-let res = x##y + z##q; /* Minimum parens */
+let res = x["y"] + z["q"]; /* AST */
+let res = x["y"] + z["q"]; /* Minimum parens */
 
 /* These should print the same */
-let res = y + z##q##a; /* AST */
-let res = y + z##q##a; /* Min parens */
+let res = y + z["q"]["a"]; /* AST */
+let res = y + z["q"]["a"]; /* Min parens */
 
 /* Make sure it's actually parsed as left precedence
  * and that is maintained when printed */
-let res = z##(q##a); /* AST */
-let res = z##(q##a); /* Min parens */
+let res = z##(q["a"]); /* AST */
+let res = z##(q["a"]); /* Min parens */
 
 /* These should print the same */
 let res = !x##y; /* AST */
@@ -44,12 +44,12 @@ let res = !z##q##a; /* AST */
 let res = !z##q##a; /* Min parens */
 
 /* These should print the same */
-let res = ?!!x##y; /* AST */
-let res = ?!!x##y; /* Minimum parens */
+let res = ?!!x["y"]; /* AST */
+let res = ?!!x["y"]; /* Minimum parens */
 
 /* These should print the same */
-let res = ?!!z##(q##a); /* AST */
-let res = ?!!z##(q##a); /* Min parens */
+let res = ?!!z##(q["a"]); /* AST */
+let res = ?!!z##(q["a"]); /* Min parens */
 
 res #= ?!!z##q;
 res #= ?!!z##(q##a);
--- a/formatTest/unit_tests/expected_output/bucklescript.re
+++ b/formatTest/unit_tests/expected_output/bucklescript.re
@@ -4,18 +4,18 @@ bla #= Some(10);
 
 bla #= someFunc(Some(10));
 
-test##var #= Some(-10);
+test["var"] = Some(-10);
 
 obj##.prop;
 
 obj##.prod := exp;
 
-preview##style##border
-#= Js.string("1px black dashed");
+preview["style"]["border"] =
+  Js.string("1px black dashed");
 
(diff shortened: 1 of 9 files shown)


formatTest/unit_tests/expected_output/bucklescript.re
```diff
@@ -4,60 +4,60 @@ bla#=(Some(10));
...
 
 /* These should print the same */
-let res = ?!!z##(q##a); /* AST */
-let res = ?!!z##(q##a); /* Min parens */
+let res = ?!!z##(q["a"]); /* AST */
+let res = ?!!z##(q["a"]); /* Min parens */
 
-res#=?!!z##q;
-res#=?!!z##(q##a);
+res#=?!!z["q"];
```
anmonteiro: this is the only case that's failing idempotent tests, but this is a bad test that appears to pass. In fact, it gets changed from `res #= ?!!z##q;` to `res #=?!! z##q;`.

I'm worried, because #2050 fixes this.

formatTest/unit_tests/expected_output/bucklescript.re
```diff
@@ -4,60 +4,60 @@ bla#=(Some(10));
...
 
-test##var#=(Some(-10));
+test["var"] = Some(-10);
 
 obj##.prop;
 
 obj##.prod := exp;
 
-preview##style##border#=Js.string(
+(preview["style"]["border"] = Js.string)(
```
anmonteiro: this case is unfortunate, but it was present before. This change just makes it obvious.

with the currently released version of Reason:

```bash
echo 'preview##style##border#=Js.string("1px black dashed");' | refmt -p ml
;;(((preview ## style) ## border) #= Js.string) (("1px black dashed")
    [@reason.raw_literal "1px black dashed"])
```

Again, #2050 fixes the precedence for this case too, so I'm not worried.

jaredly:
ooooh this is wonderful!

chenglou:
Yeah, that's a given!

anmonteiro:
Just rebased this on current master. There was some breakage caused by a refactor in the printer that was included in #2055.

Some thoughts on error messages: I'm unsure how we'd accomplish good error messages in Reason itself, since `foo[bar]` is valid if `foo` is an array and `bar` is an int. Sounds like if we wanna support this it'd have to go in BuckleScript?

jaredly:
Maybe bettererrors? if `bar` is a string, show a nice message

chenglou:
Yep, in bettererrors (eh super-errors)

chenglou:
@anmonteiro since everyone seems to like this, let's ship it?

anmonteiro:
Sounds good to me. Just rebased on current master.

jordwalke:
@anmonteiro, I think @jaredly's concern is actually pretty valid. In fact there's technically a way this diff is not backwards compatible because you could have redefined an `Array` module in scope with a `get` function that accepted a `"string"` index. That's pretty unlikely, but the bigger issue I see is the one that Jared pointed out. People will think that this is a dynamic field access. I believe that there is a workaround or improvement which makes that confusion much less likely and provides a better opportunity to give warnings very early. That is to use single quotes for the keys here.  The reason is that string keys longer than length 1 aren't even possible to write. You could give a nice error there.

```
let myString = 'foo';        // ParserError here
let result = myobj[myString];
```
> **ParseError**: This is a static string only to be used when indexing a data structure such as obj['foo']. It cannot be dynamically computed outside of the immediate field access`.

It isn't possible/easy to provide such an up front error for double quote strings.
Single quote string keys are also commonly used to index dictionaries in other languages including JavaScript, so I don't think it's any worse. In fact, I think FB's style guide says to use single quote keys anyways.

Of course - the remaining wrinkle is that you still have some of the same problems as the double quote string key sugar, but for single characters like `'a'` because that is a valid expression. However, that is so much less frequently used than strings. I believe that by using single quote string keys the amount of confusion will be cut down by about 75% of what it would be with double quoted strings, and I don't see a lot of downside to them.

jordwalke:
It's unfortunate that both Reason and OCaml's syntax hard code`y[x]` access to the `Array` module. I do wish we could remove that and I'm open to possible ways we can do that.

If that were possible, wouldn't the only benefit to having `x['y']` be sugar for `x##y` be the forwards compatibility? If you had the ability to stop having `x[y]` be hard coded to the `Array` access, and instead be open to various data structures, would you rather have that? For example if you're doing UI work you have your build system make `[ ]` field access become `Js.get`/`Js.set`. If doing backend work, maybe you want to use it for all hash tables, regardless of their key datatype. That means you could use `x[0]` for integer keys, and `x["foo"]` for string keys.

PR #2055, opened 2018-07-06
Fix precedence of `->` in the presence of `##` or `[idx]`

anmonteiro:
TODOs:

- [x] roundtrip `|.` correctly by fixing the printer to print `foo |. bar[0]` as `foo->(bar[0])
- [x] fix the printer such that `foo->bar[0]` prints as such (and not `(foo->bar)[0]`)


IwanKaramazow:
Precedence and grammar looks ok. What do you mean by "roundtrip |. correctly"?

anmonteiro:
We want to format `foo |. bar[0]` to `foo->(bar[0])` because `|.` has different precedence than `->`

anmonteiro:
Solved the listed TODOs in the PR description.

We could still improve the printing of this though:

```bash
$ echo 'location##streets.foo[1]' | ./_build/install/default/bin/refmt
(location##streets).foo[1];
```

src/reason-parser/reason_parser.mly
```diff
@@ -1355,18 +1355,14 @@ conflicts.
...
-%nonassoc DOT POSTFIXOP
 
 (* We need SHARPEQUAL to have lower precedence than `[` to make e.g.
    this work: `foo #= bar[0]`. Otherwise it would turn into `(foo#=bar)[0]` *)
 %left     SHARPEQUAL
-
-%nonassoc LBRACKET
-(* SHARPOP has higher precedence than `[`, see e.g. issue #1507. *)
-%left     SHARPOP
+%left     SHARPOP MINUSGREATER LBRACKET DOT
```
IwanKaramazow: Is it possible to add a comment that explains the road that lead us to making all the `%left` assoc?

src/reason-parser/reason_pprint_ast.ml
```diff
@@ -4117,15 +4160,20 @@ let printer = object(self:'self)
                               atom "=";
                               (label (self#longident_loc lid) (atom "."))])
                    (self#formatNonSequencyExpression e)
+           | Pexp_apply (eFun, _) ->
+             let lhs = (makeList [atom lbl; atom "="]) in
+             let rhs = (match printedStringAndFixityExpr eFun with
```
IwanKaramazow: It looks a bit strange to have "implementation details" about token precedence leaking into the formatting logic of jsx.
Idea: what if we make `.`, `#`, `->` etc. `Simple` ?

Given:
- extreme high precedence
- `^.` already emits as `POSTFIXOP "^" DOT` in the lexer
- `->`, `#` now have the same "access"-like precedence as `.`

This makes them `Simple`, no? If the unparsing emits them as `Simple`, then we could avoid the leakage of precedence details in other parts of the printer.
Thoughts?

src/reason-parser/reason_pprint_ast.ml
```diff
@@ -7106,6 +7157,12 @@ let printer = object(self:'self)
           -> `LastArgIsCallback(callback, List.rev args)
       | _ -> `NormalFunAppl args
     in
+    let formattedFunExpr = match (Reason_ast.processFastPipe funExpr).pexp_desc with
```
IwanKaramazow: It's a bit strange to call `processFastPipe` again here. The rewriting/processing of the ast feels a bit intertwined with the pure printing logic.
What if we change the implementation of `processFastPipe` to do the transformation up front?
image

src/reason-parser/reason_pprint_ast.ml
```diff
@@ -4105,6 +4147,7 @@ let printer = object(self:'self)
         processArguments tail (nextAttr :: processedAttrs) children
 
       | (Labelled lbl, expression) :: tail ->
+         let expression = Reason_ast.processFastPipe expression in
```
IwanKaramazow: See comment below

src/reason-parser/reason_pprint_ast.ml
IwanKaramazow: Also can we add extra unit tests to verify the printing output?

```reason
<div bar=f->g->z />;
<div bar=foo##bar />;
```

IwanKaramazow:
One thing that feels strange is the precedence of
```reason
foo->f(1, 2);

/* gets parsed as */
(foo->f)(1, 2)
```

It's natural that `event->target##value` reads from left to right: `(event->target)##value`
But when using it as a pipe first chain, it feels weird:
```reason
compilation
->Plugin.buildAssets
->Js.Json.stringify
->Node.Fs.writeFileAsUtf8Sync(path);
```
Simplified we get the following ast:
```reason
jsonString->Node.Fs.writeFileAsUtf8Sync(path);
/* parses as */
(jsonString->Node.Fs.writeFileAsUtf8Sync)(path);
(jsonString |. Node.Fs.writeFileAsUtf8Sync)(path);
/* after bs ppx transform */
Node.Fs.writeFileAsUtf8Sync(jsonString)(path);
```
Notice how we get an extra curried function here in contrast with
`jsonString |. Node.Fs.writeFileAsUtf8Sync(path)`.
Bucklescript's ppx transforms this to `Node.Fs.writefileAsUtf8Sync(jsonString, path)`.
If someone would publish the ppx for native, would he need to be aware of this implementation detail?

Technically speaking `->` and `|.` are not 100% idempotent on the ast level. We could add an extra type checked test to cover/document this behaviour.

This is what the PR implements:
```reason
type fn = (int, int) => int;
let f: fn = (a, b) => a + b;
let (|.) = (a, f: fn) => f(a); /* notice how the return type is a function: "int => int" */

let i: int = 1->f(2);
```
This is the ast that I would expect we deliver to third party ppx:
```reason
type fn = int => int;
let (|.) = (a, b) => a + b; /* notice how the return type is just "int" */
let f: fn = (a) => a * 2;

let i: int = 1->f(2);
```
Isn't one of the goals of the bs fast pipe ppx transform to avoid an extra function allocation (due to the currying) ?

anmonteiro:
I agree with most your comments, but I need to verify that they’re feasible (e.g SpecificInfix -> Simple, I cant tell you that’ll work without trying it out first).

As for the last comment about `foo->bar->baz(qux)` I think we need @chenglou’s input here; when I asked him about this he was comfortable reading it from left to right.

jaredly:
@IwanKaramazow as I understand it, there's no AST-level difference between `Node.Fs.writeFileAsUtf8Sync(jsonString)(path);` and `Node.Fs.writeFileAsUtf8Sync(jsonString, path);`

IwanKaramazow:
@jaredly  I forgot to add parens: `(Node.Fs.writeFileAsUtf8Sync(jsonString))(path);`
image

I might be completely wrong, the compiler might be able to transform `(Node.Fs.writeFileAsUtf8Sync(jsonString))(path);` into `Node.Fs.writeFileAsUtf8Sync(jsonString, path);` at a later compiler stage. But purely ast-wise, there's a small difference.

chenglou:
The AST before transform for the first and last item might differ, but after the ppx transform, there's no difference. The ppx itself turns `(a |. b) c` into `(b a) c` and `a |. (b c)` into `b a c`. Which are equivalent in the AST post-transformation.

So unless we're talking about actually using `|.` without the BS ppx (which isn't currently possible on the BS side; custom `|.` is disabled through a clever trick), we can feel free to use either precedence here for `->` and application.

chenglou:
@anmonteiro any last change before we merge?

chenglou:
Seems legit now. Thanks!

PR #2058, opened 2018-07-08
support non-parenthesized label colon type equal optional in type dec…

### LINKED ISSUE #2056: Support `type asd = ~a:bool=? => int;`
Right now you need to wrap in parens: `type asd = (~a:bool=?) => int;`

Dunno what conflict this has

anmonteiro:
…larations

fixes #2056


IwanKaramazow:
Is it possible to also include the core_type without `EQUAL optional` `type asd = ~a:bool => int;` ?

chenglou:
Merging! Thanks

PR #2060, opened 2018-07-09
Recover jsx component near arrow of pattern match.

IwanKaramazow:
```reason
switch(foo) {
  | `Variant =><Component />
};
```

`=><` isn't an infix operator in this context and should be parsed as jsx.


anmonteiro:
Looks 👍 to me

chenglou:
So many hacks for JSX!

PR #2061, opened 2018-07-09
Unify module type formatting

### LINKED ISSUE #2057: Module type with attribute formats weirdly
```rust
module type Foo = {[@someattr]
                   let foo: 'a => 'a;};
```

anmonteiro:
Among other things, this fixes #2057 

There's still some cleanup left to do, but I'd love to get some initial feedback on this.

### DIFF
--- a/src/reason-parser/reason_pprint_ast.ml
+++ b/src/reason-parser/reason_pprint_ast.ml
@@ -5831,7 +5831,7 @@ let printer = object(self:'self)
         | Some id -> [atom ("%" ^ id.txt)]
       in
       makeList
-        ~postSpace:true ~indent:0 ~break:Layout.Always ~inline:(true, true)
+        ~postSpace:true ~indent:0 ~break:Layout.Always_rec ~inline:(true, true)
         (extension @ List.map self#item_attribute l @ [toThis])
 
   method exception_declaration ed =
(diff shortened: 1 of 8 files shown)


src/reason-parser/reason_pprint_ast.ml
```diff
@@ -6596,17 +6583,47 @@ let printer = object(self:'self)
...
-          ~break:IfNeed
-          ~inline:(true, false)
-          ~wrap:("{", "}")
-          ~postSpace:true
-          ~sep:(SepFinal (";", ";"))
-          items
-      | Pmty_extension (s, e) -> self#payload "%" s e
-      | _ -> makeList ~break:IfNeed ~wrap:("(", ")") [self#module_type x]
+        let someSigHasAttrs =
+          List.exists
```
IwanKaramazow: Instead of inspecting the children here, you can use `Always_rec` deeper in the tree.
This recursively breaks nodes from the child up to the root of the tree.

chenglou:
Alright so how's this one?

IwanKaramazow:
The logic for `module type` presented in the PR:
- break if there's a "row" with an attribute
- break if there's more than one "row"

This is brings `module type X = {...}` closer to `module X = {...}`. The printing output is a very nice improvement.

There's one small comment that could improve the code by using a feature provided by `easy_format`  instead of doing that logic in userland.

anmonteiro:
I’ll look into @IwanKaramazow’s suggestion this evening

chenglou:
Since we’re gonna release a 3.3.1 https://github.com/facebook/reason/pull/2071, we can get this one in too

anmonteiro:
Just addressed the code review item – that part of the code looks much simpler now.

IwanKaramazow:
Has my blessing :)

chenglou:
Wooo

PR #2062, opened 2018-07-09
Better error message for modules with lowercase identifiers

### LINKED ISSUE #1974: lowercase module name: Syntax error
When you start a module with lowercase letter you get a syntax error. It would be nice to show a clear warning instead. I.e.: Error: module must always start with a capital letter.
```
module Uppercase {
  /* this works */
}
module lowercase {
  /* Error: 983: <UNKNOWN SYNTAX ERROR> */
}
```

anmonteiro:
fixes #1974

### DIFF
--- a/src/reason-parser/reason_parser.mly
+++ b/src/reason-parser/reason_parser.mly
@@ -1047,8 +1047,8 @@ let record_pat_spread_msg =
 Explanation: you can't collect a subset of a record's field into its own record, since a record needs an explicit declaration and that subset wouldn't have one.
 Solution: you need to pull out each field you want explicitly."
 
-let lowercase_module_msg s =
-  Printf.sprintf "Module names must be uppercase identifiers. `%s` is not an uppercase identifier." s
+let lowercase_module_msg =
+  Printf.sprintf "Module names must start with an uppercase letter."
 
 (* Handles "over"-parsing of spread syntax with `opt_spread`.
  * The grammar allows a spread operator at every position, when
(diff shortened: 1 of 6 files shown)


src/reason-parser/reason_parser.mly
```diff
@@ -1047,6 +1047,9 @@ let record_pat_spread_msg =
 Explanation: you can't collect a subset of a record's field into its own record, since a record needs an explicit declaration and that subset wouldn't have one.
 Solution: you need to pull out each field you want explicitly."
 
+let lowercase_module_msg s =
+  Printf.sprintf "Module names must be uppercase identifiers. `%s` is not an uppercase identifier." s
```
hcarty: Do we have `uppercase identifier` defined somewhere?  It may be helpful to be explicit here - something along the lines of ``Module names must start with an uppercase letter.  `%s` does not start with an uppercase letter.``

anmonteiro: @hcarty that sounds good to me. I'm expecting @chenglou to bikeshed the error message anyway!

chenglou:
Yeah no need to say that "foo" isn't uppercased.

Thanks!

IwanKaramazow:
I'm not sure to what extend we should encode all those error cases into the grammar. There's the cost of tampering with the grammar that might result into shift/reduce conflicts in the future. For example, `LIDENT` is a common for core_types/mod_types. Some errors might be better positioned in the error message table to avoid needless grammar complexity.

chenglou:
Should we revert this one then?

Issue #2066, opened 2018-07-12
Doc comments and %raw interact badly


cristianoc:
This:
```re
/** header */

[%% raw
"console.log(42)"];
```

formats to
```re
%raw
/** header */
"console.log(42)";
```

See:
https://reasonml.github.io/en/try.html?rrjsx=true&reason=PQKhAIAsFMEMBNoCdwmAKHQbQKQ-ErAO7oBEAxgPYB2AzpQDbQB0DlA5gBQAsATAJSkAugG50QA

cristianoc:
I came across this while doing some code generation, where a header is generated first, and some raw expression is generated next.

PR #2069, opened 2018-07-12
Attach doc attrs before extension sugar

IwanKaramazow:
Fixes https://github.com/facebook/reason/issues/2066


cristianoc:
Thanks for the super-quick fix!

PR #2074, opened 2018-07-13
Parens hugging for Js.t obj types

### LINKED ISSUE #2073: Parens hugging for js.t obj types
```reason
type a = array(
  {
    .
    "name": string,
  },
);
```

Would be nice to refmt as:

```reason
type a = array({
  .
  "name": string,
});
```

anmonteiro:
fixes #2073

### DIFF
--- a/src/reason-parser/reason_pprint_ast.ml
+++ b/src/reason-parser/reason_pprint_ast.ml
@@ -3037,6 +3037,10 @@ let printer = object(self:'self)
             | [{ptyp_desc = Ptyp_object (l, o) }] when not (isJsDotTLongIdent li.txt) ->
                 label (self#longident_loc li)
                   (self#unparseObject ~wrap:("(",")") l o)
+            | [{ptyp_desc = Ptyp_constr(lii, [{ ptyp_desc = Ptyp_object (ll, o) }])}]
+              when isJsDotTLongIdent lii.txt && List.length ll > 0 ->
+              label (self#longident_loc li)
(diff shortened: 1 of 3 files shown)


IwanKaramazow:
This is the correct approach.
There's another issue where we should format the `Ptyp_constr` on the same line as `type foo =`
```reason
type foo = array({    -> notice how "array" sticks to "type foo"
...
});
```
This is a much bigger project on it's own, touching the internals of type declaration.
I propose that we solve this in another PR.

src/reason-parser/reason_pprint_ast.ml
```diff
@@ -3037,6 +3037,10 @@ let printer = object(self:'self)
             | [{ptyp_desc = Ptyp_object (l, o) }] when not (isJsDotTLongIdent li.txt) ->
                 label (self#longident_loc li)
                   (self#unparseObject ~wrap:("(",")") l o)
+            | [{ptyp_desc = Ptyp_constr(lii, [{ ptyp_desc = Ptyp_object (ll, o) }])}]
+              when isJsDotTLongIdent lii.txt && List.length ll > 0 ->
```
chenglou: Try `ll <> []`. `O(1)` check

chenglou: Eh ops, just do `ptyp_desc = Ptyp_object (_::_, o)`

chenglou: I'll correct that. Thanks!

PR #2075, opened 2018-07-14
[Printer] Turn most List.length into pattern matching checks

chenglou:
O(n) vs O(1)


chenglou:
Not passing yet. Hold on. Local setup borked

PR #2078, opened 2018-07-15
Change the precedence of fast pipe in the presence of function applic…

anmonteiro:
…ation

### DIFF
--- a/formatTest/unit_tests/expected_output/fastPipe.re
+++ b/formatTest/unit_tests/expected_output/fastPipe.re
@@ -5,7 +5,10 @@ bar->f->g->h;
 compilation
 ->Plugin.buildAssets
 ->Js.Json.stringify
-->Node.Fs.writeFileAsUtf8Sync(_, path);
+->Node.Fs.writeFileAsUtf8Sync(
+  _,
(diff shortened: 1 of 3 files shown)


chenglou:
The be clear: this is a precedence change. It doesn't affect BS side, because `x->y(z)` was turned into `x |. (y z)` and then through the ppx, `y x z`. Now it'll be turned into `(x |. y) z` which through the ppx becomes `(y x) z` aka also `y x z`. Same perf, same type checking (?), same semantics when `z` is a argument (aka, `x->y(~z)` still works.

This change is only to accommodate non-bsb-native users who don't get the `|.` ppx transform (into which `->` is desugared, and would still like to use it by redefining it themselves as an infix operator: `let (|.) = (a, b) => b(a)`\*. This won't have the same perf benefits (there might be intermediate functions), doesn't work with labeled arguments and won't type check the same way. But at least it'll work.

\* Before this diff, this wasn't possible, since we transform `x->y(z)` into `x |. (y z)`, whereas they needed `(x |. y) z`.

cc @bobzhang @IwanKaramazow to confirm!

PR #2080, opened 2018-07-19
Remove ppx_react

chenglou:
It's unused right now


chenglou:
cc @jordwalke. This is a dead ppx. I don't think you expose it in reason-cli, but yeah just a heads up.

PR #2082, opened 2018-07-21
Fix the precedence of UnaryPostFix in the presence of fast pipe

anmonteiro:
this fixes the case where parsing `(event->target^)##value;` would be
interpreted as `(event->(target^))##value;`


chenglou:
Thanks!

PR #2087, opened 2018-07-27
Implement correct printing for fast pipe sugar

IwanKaramazow:
```reason
/* before (master refmt) */
compilation->Plugin.buildAssets(foo, bar, baz)
->Js.Json.stringify
->Node.Fs.writeFileAsUtf8Sync(
  _,
  path,
);

/* after */
compilation
->Plugin.buildAssets(foo, bar, baz)
->Js.Json.stringify
->Node.Fs.writeFileAsUtf8Sync(_, path);
```

Also implements and tests the uncurried syntax under fast pipe sugar.
```
foo->f(.)->g(. a, b);
```

PR #2092, opened 2018-07-29
Fix precedence of infix ops in the presence of fast pipe

anmonteiro:
this fixes a regression introduced in https://github.com/facebook/reason/pull/2087


IwanKaramazow:
Discussed in private, approved.

chenglou:
Thanks!

Issue #1163, opened 2017-03-23
Inline records in variant type definitions aren't pretty-printed properly in rtop


hcarty:
Reason 1.11.0 + OCaml 4.04.0 from opam

Inline record arguments to variant constructors are not pretty-printed properly from rtop but they do show up properly in actual values:

For example:

```reason
Reason # type t = | A { x : int };
type t = A
Reason # A { x : 1 };
- : t = A {x : 1}
```

`refmt` does the right thing so it seems to be limited to `rtop`.

jordwalke:
Fixed here: https://github.com/facebook/reason/pull/2336

PR #2103, opened 2018-08-01
Print inline record constructors properly

reynir:
Fixes https://github.com/facebook/reason/issues/1163


reynir:
I added some Oprint tests for inline records, but I'm not sure if it's a good idea if support for OCaml <4.03.0 is desired.

jordwalke:
Extended here with extra test cases: https://github.com/facebook/reason/pull/2336

Thanks for getting the ball rolling.

PR #2111, opened 2018-08-05
Fix precedence printing of `a->(b##c)`

anmonteiro:
This change makes the formatting of fast pipe nodes reuse the same logic that we use elsewhere to format infix operators.

With this change I hope to mitigate future edge cases that would appear because of this custom fast pipe logic.

### DIFF
--- a/formatTest/unit_tests/expected_output/fastPipe.re
+++ b/formatTest/unit_tests/expected_output/fastPipe.re
@@ -109,3 +109,7 @@ foo->([@attr] f(.))->([@attr] g(.));
 - 1->foo;
 !foo->bar;
 (!foo)->bar;
+
+a->(b##c);
+
+a->b##c;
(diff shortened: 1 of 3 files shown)


IwanKaramazow:
@anmonteiro can you add an extra test for https://github.com/facebook/reason/issues/2118
```
 (switch (saveStatus) {
                | Pristine => ""
                | Saved => "Saved"
                | Saving => "Saving"
                | Unsaved => "Unsaved"
                })
              ->str

blocks->(blocks => {"blocks": blocks})

(state.title == "" ? "untitled" : state.title)->str
<title> ((state.title == "" ? "untitled" : state.title)->str) </title>
```

thangngoc89:
@IwanKaramazow I added a number 4 case on #2118. You might want to check it out

IwanKaramazow:
Discussed in private, ready!

PR #2123, opened 2018-08-10
Fix local open printing for `bs.obj`

### LINKED ISSUE #2121: [refmt] Js.t objects and local open
Since 3.3.2,

```reason
let styles =
  StyleSheet.create(Style.(
    {
      "container": style([flex(1.0), backgroundColor(String(Colors.white))]),
    }
  ));
```

gets reformatted to

```reason
let styles =
  StyleSheet.create(
    {
      open Style;
      {"container": style([flex(1.0), backgroundColor(String(Colors.white))])};
    },
  );
```

I would prefer if refmt kept the original formatting (as was the case for the previous refmt version).

Behavior is different for "normal" Reason records BTW.

```reason
let styles2 =
  StyleSheet.create(Style.(
    {
      container: style([flex(1.0), backgroundColor(String(Colors.white))]),
    }
```

becomes

```reason
let styles2 =
  StyleSheet.create(
    Style.{
      container: style([flex(1.0), backgroundColor(String(Colors.white))]),
    },
  );
```

anmonteiro:
fixes #2121 

This fixes a regression introduced in https://github.com/facebook/reason/pull/1979

### DIFF
--- a/formatTest/unit_tests/expected_output/bucklescript.re
+++ b/formatTest/unit_tests/expected_output/bucklescript.re
@@ -119,3 +119,6 @@ type y = {
   "height":
     (int, int, int, float, float, float) => unit,
 };
+
+/* https://github.com/facebook/reason/issues/2121 */
+Style.({"container": 3});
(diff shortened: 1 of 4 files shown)


IwanKaramazow:
Perfect, @chenglou let's merge!

Issue #2122, opened 2018-08-09
First class module printing bug.


jordwalke:
```
echo 'type t = ref( (module Console) )'  | refmt
```

Outputs:

```
type t = ref(Console);
```

Which is wrong and won't even parse.

Is this a new bug?

IwanKaramazow:
I’m fixing the issue

hcarty:
I've hit this in type definitions as well when pinned to `master`.
```reason
type t = {
  m: (module M);
};
```
refmts as
```reason
type t = {
  m: M;
};
```

PR #2124, opened 2018-08-10
Implement correct printing for Ptyp_package as core_type.

IwanKaramazow:
Fixes https://github.com/facebook/reason/issues/2122

Also adds support for less parens in the context of a Ptyp_constr/Ptyp_tuple:

```reason
/* before */
type t = ref((module Console));

/* after */
type t = ref(module Console);
```


hcarty:
Does this cover record fields like `type t = { m: (module M) };` too?

Thank you for working on this!   I want to use Reason's `master` branch for the improved merlin support but this bug breaks a lot of code for me vs 3.2.0.

IwanKaramazow:
@hcarty added a test case, I'll update the PR to allow parsing without parens in that case.

PR #2129, opened 2018-08-13
Bump version in package.ml

jordwalke:


jordwalke:
Hmm, looks like cache key is using an invalid component (opam file which doesn't exist), so CI builds are slower than normal. Should fix that in a follow up.

PR #2136, opened 2018-08-13
Fix `foo(~Foo.a)` becoming `foo(~Fooa=Foo.a);

### LINKED ISSUE #2128: `foo(~Foo.a)` becomes `foo(~Fooa=Foo.a);`
I don't think we should parse `Foo.a` at all.

```shell
$ echo 'foo(~Foo.a)' | ./_build/install/default/bin/refmt
foo(~Fooa=Foo.a);
```

As @IwanKaramazow pointed out, this is what the OCaml manual says about label names:

anmonteiro:
fixes #2128

### DIFF
--- a/formatTest/unit_tests/expected_output/syntax.re
+++ b/formatTest/unit_tests/expected_output/syntax.re
@@ -1385,3 +1385,7 @@ foo(~a: int);
 foo(~(a :> int));
 
 foo(~(a :> int));
+
+foo(~a=?Foo.a);
+
+foo(~a=Foo.a);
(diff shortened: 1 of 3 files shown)


jordwalke:
Thanks. I'm not really sure this punning shortcut is all that valuable, do you? Maybe now that I know it exists, I can think of use cases for it.

anmonteiro:
@jordwalke the goal for the PR was preventing things like `foo(~Fooa=Foo.a)`, although the punning is already there.

My agreement with @IwanKaramazow is we print without punning and don't really advertise this feature. So we're lenient in parsing but format to the right thing.

PR #2145, opened 2018-08-15
Remove upper constraint on Menhir

anmonteiro:


jordwalke:
This was placed here for a reason. Do you want to find out why and then explain why it's no longer needed?

anmonteiro:
@jordwalke I assumed the reason why Menhir was not bumped further was because an error that started happening with versions starting from the one after the upper constraint:

```
Error: symbol nonempty_list(attribute) is unreachable from any of the start symbol(s).
```

This diff also takes care of that, though, by removing the `attribute+` symbol from the on_error_reduce rule, given it's repeated (see `as_loc(attribute)+` above it).

jordwalke:
Great, thank you! This will help prevent some package manager conflicts.

rgrinberg:
Would be great to have a release of this on opam.

Issue #1980, opened 2018-06-02
[rtop] ^ deref becomes `not` when loading files with #use


ostera:
Hi! I was trying to get the [following example](https://reasonml.github.io/en/try.html?rrjsx=true&reason=DYUwLgBAhhC8ECcQDMAUAGAdOglAbgCgCYAueKAPQgGpNoK8g) loaded on `rtop`:

```reason
let a = ref(0.0);

a := a^ +. a^;
```

And if I type it in manually it works just fine:

```reason
Reason # let a = ref(0.0);
let a: ref(float) = {contents: 0.};
Reason # a := 1.0;
- : unit = ()
Reason # a;
- : ref(float) = {contents: 1.}
Reason # a^;
- : float = 1.
Reason # a^ +. a^;
- : float = 2.
Reason # a := a^ +. a^;
- : unit = ()
```

But when loading the code with `#use` I see this:

```reason
Reason # #use "./Test.re";
let a: ref(float) = {contents: 0.};
File "./Test.re", line 3, characters 5-6:
Error: This expression has type ref(float)
       but an expression was expected of type bool
```

Whereas I'd expect this to parse and run normally. I've recurred so far to using `.contents` instead of `^`.

#### Toolchain
```
λ rtop -version
The universal toplevel for OCaml, version 2.1.0, compiled for OCaml version 4.02.3
λ bsb -version
3.1.4
```

Let me know if I can provide any other information!

ostera:
As a follow up I understand that in the contrived example I provided I could have rebound `a`, but in my own code I need to mutate from different scopes and thus rebinding is not an option.

jaredly:
rtop hasn't gotten a ton of attention afaik :/
I think I know what's going on though! Somehow we've got a double conversion, because:
`a^` in reason becomes `!a` in ocaml,
and `!a` in reason becomes `not a` in ocaml.

thangngoc89:
I hit this bug today. Narrow this down to `ReasonToolchain.RE.use_file` https://github.com/facebook/reason/blob/69837eab531c18af1afb35c9fdf6e3931ce951d8/src/rtop/reason_toploop.ml#L17-L19

Issue #1970, opened 2018-05-28
[rtop] Fatal error on load with #use when file contains module not ended with semicolon


Shakadak:
Hello, when a module is defined in a file, but the ending semicolon was forgotten. If there is some code afterward, `rtop` crashes with `Fatal error: exception Syntax_util.Error(_, _)`.  
I suppose this is sort of expected, but it would be nice to have some kind of message to help regarding that semicolon.

```reason
module Test = {
    let id : 'a => 'a =
        x => x
}

let _ = print_endline("banana");
```

PS: On a side note, no error or warning is reported when watching with `bsb -make-world -w`, is this configurable ?

Issue #1965, opened 2018-05-27
[rtop] Type errors when loading file with #use on expression using (++) operator


Shakadak:
Hello,

I'm not quite sure this is the place for that kind of bug report but the rtop repo is empty.
```reason
type parser('a) = string => list(('a, string));

/* Parser combinators */
let bind : parser('a) => ('a => parser('b)) => parser('b) =
    (p, f) => inp =>
        p(inp)
        |> List.map(((v, inp')) => f(v, inp'))
        |> List.concat;

let word : parser(string) = {
        inp' => [(String.make(1, 'c') ++ "banana", inp')];
};
```

Maybe it's just me misunderstanding something.
Trying to load this code from a file with `#use` brings up the following error:
```
Reason # #use "./src/x.re";
type parser('a) = (string) => list(('a, string));
let bind: (parser('a), ('a) => parser('b)) => parser('b) = <fun>;
File "./src/x.re", line 12, characters 38-40:                      
Error: This function has type (bool) => bool
       It is applied to too many arguments;
       maybe you forgot a `;'.
```
When commenting either `parser`'s definition or `bind`'s definition, the error changes to:
```
Reason # #use "./src/x.re";
type parser('a) = (string) => list(('a, string));
File "./src/x.re", line 12, characters 18-37:
Error: This expression has type string                             
       but an expression was expected of type ref(('a) => 'b)
```
And when commenting them both, `word`'s definition gets accepted:
```
Reason # #use "./src/x.re";
let word: parser(string) = <fun>;
```
utop seems to have no problem loading the file reformated using `refmt -p ml src/x.re > x.ml`
```
utop # #use "x.ml";;
type 'a parser = string -> ('a * string) list
val bind : 'a parser -> ('a -> 'b parser) -> 'b parser = <fun>
val word : 'a -> (string * 'a) list = <fun>
```
Copy pasting the code in rtop seems to work fine.

utop is on version 2.0.1
ocaml on version 4.04.2
and reason on 3.0.4 I believe.

jaredly:
I think this is the same issue as #1980 -- `++` becomes `^` which then becomes "deref" (`!`)

PR #2146, opened 2018-08-15
Fix parsing of `use_file`

anmonteiro:
Because the `use_file` parser definition is recursive, we were applying
the operator swapping mappers more than once in the cases where the rule
recursed.

Instead, we should only apply the mapper to the final result of parsing
the `use_file` rule.

Fixes https://github.com/facebook/reason/issues/1965
Fixes https://github.com/facebook/reason/issues/1980
Fixes https://github.com/facebook/reason/issues/1970


jordwalke:
Thank you as always!

PR #2148, opened 2018-08-16
Fix fast pipe formatting for `a->(b->c)`

anmonteiro:
This diff also makes fast pipe `SpecificInfixPrecedence` by default,
instead of `Simple`, as recent bugs have shown that fast pipe formatting
is definitely not simple!

By having fast pipe be `SpecificInfixPrecedence` by default again, it
eliminates hopefully the last class of bugs that we hav seen recently.


IwanKaramazow:
@chenglou discussed with @anmonteiro on Discord, ready to merge.

PR #2149, opened 2018-08-17
Housecleaning: move to Dune and fix compilation warnings

anmonteiro:
TODO:

- [x] apparently `String.capitalize_ascii` doesn't exist in versions of OCaml prior to 4.04, so I'll have to change that back
- [x] bspack workflow still works
- [x] revert `; _` changes, disable warning 9
- [x] esy workflow


kennetpostigo:
@anmonteiro you could always just manually include it in

src/ppx/reactjs_jsx_ppx_v2.ml
```diff
@@ -73,12 +73,12 @@ let transformChildrenIfList ~loc ~mapper theList =
     (* not in the sense of converting a list to an array; convert the AST
        reprensentation of a list to the AST reprensentation of an array *)
     match theList with
-    | {pexp_desc = Pexp_construct ({txt = Lident "[]"}, None)} ->
+    | {pexp_desc = Pexp_construct ({txt = Lident "[]"; _}, None); _} ->
```
jordwalke: I think this warning is way too sensitive. You can customize the flags to put the `_` back.

jordwalke:
Thank you, please see my one comment.

anmonteiro:
Addressed your comments in the last 2 commits

jordwalke:
Thank you as always!

PR #2153, opened 2018-08-21
[internal] get the reason bspack build script working with esy

jaredly:
This basically means all you need is to clone the repo and do `cd bspacks; ./bspack.js` and it will All Work™

I think this is a much smoother workflow... lmk if there's anything I haven't thought of though.

(this is motivated in part by the desire to make reason releases easier to do)


src/refmt/package.ml
```diff
@@ -1,4 +1,4 @@
 
-let version = "3.3.3"
-let git_version = "fefe5e4db3a54a7946c2220ee037dd2f407011c9"
-let git_short_version = "fefe5e4d"
+let version = "4.4.5"
```
IwanKaramazow: This version seems interesting 😄

bspacks/esy.json
```diff
@@ -0,0 +1,27 @@
...
+    "ocaml": "4.2.3004"
+  },
+  "devDependencies": {
+    "ocaml": "~4.2.3004"
+  },
+  "esy": {
+    "build": [
+      ["echo", "No-op"]
+    ],
+    "_dependenciesForNewEsyInstaller": {
```
jordwalke: This shouldn't be needed. Was it? It was just a hack for some special cased packages to make sure some old projects will continue to install correctly. It no longer applies.

bspacks/esy.json
```diff
@@ -0,0 +1,27 @@
...
+  "name": "get-js-of-ocaml",
+  "notes": "This is to get js_of_ocaml",
+  "private": true,
+  "version": "3.3.3",
+  "dependencies": {
+    "@opam/js_of_ocaml": "*",
+    "@opam/menhir": " >= 20170418.0.0  <= 20171013.0.0",
+    "@opam/ocaml-migrate-parsetree": "*"
+  },
+  "peerDependencies": {
```
jordwalke: peerDependencies are now no longer needed. The constraint solver fixed that ;)

bspacks/reason_bspack.sh
```diff
@@ -7,17 +7,17 @@ set -e
 
 THIS_SCRIPT_DIR="$(cd "$( dirname "$0" )" && pwd)"
 
-echo "**This script is switching you to ocaml 4.02.3 for the subsequent bspacking. Please switch back to your own version afterward. Thanks!**\n"
-opam switch 4.02.3
-eval $(opam config env)
+cd $THIS_SCRIPT_DIR
+
+# Install & build menhir, omp
+esy i
```
jordwalke: <3

I had the same idea a long time ago (to automate the CI via esy).

bspacks/reason_bspack.sh
```diff
@@ -29,81 +29,152 @@ REFMT_PRE_CLOSURE="$buildDir/refmt_pre_closure"
...
-# for the native binary
-echo "let f = (a) => 1" | "$REFMT_BINARY.byte" --parse re --print ml
-# for the js bundle
-node ./testRefmtJs.js
+
+build_reason_402 () {
+  # rebuild the project in case it was stale
+  cd ..
+  # We need 4.02 for js_of_ocaml (it has a stack overflow otherwise :/)
+  sed -i '' 's/"ocaml": "~4.6.0"/"ocaml": "~4.2.3004"/' ./esy.json
```
jordwalke: I can see why you wanted the sandboxes feature. It's unfortunate you had to destroy the esy.json file.

bspacks/reason_bspack.sh
```diff
@@ -29,81 +29,152 @@ REFMT_PRE_CLOSURE="$buildDir/refmt_pre_closure"
...
+  make pre_release
```
jordwalke: I think you want to do something like `version=<VERSION> make pre_release`. Probably want to read it out of the `esy.json` file's version field.

bspacks/reason_bspack.sh
```diff
@@ -29,81 +29,152 @@ REFMT_PRE_CLOSURE="$buildDir/refmt_pre_closure"
...
+  esy build
+  git checkout esy.json esy.lock
+  cd -
+}
+
+# Get OMP source from esy
+get_omp () {
+  mkdir $ocamlMigrateParseTreeTargetDir
+
+  esy build-env node_modules/@opam/ocaml-migrate-parsetree > _get_root
```
jordwalke: In the latest esy you can make `esy.json` "scripts" entries which give you access to variables such as `#{ocaml-migrate-parsetree.build}`.

bspacks/reason_bspack.sh
ulrikstrid: Isn't
```
esy install
esy build
```
the same as just `esy` now? But I maybe we want it more explicit?

jordwalke: Yes. However, we might find some issues with the `esy=esy i && esy b` shortcut so it's good to be explicit in case it changes.

ulrikstrid:
When using `-i` on my mac I had to use `''` but that fails on Linux where it's not needed and I think that is what @chenglou is hitting.

chenglou:
@jaredly hold on, you make refmt work with bsc. Maybe the sed can be dropped altogether?

jaredly:
@chenglou I changed set to use gnu format

chenglou:
On newest Esy:

```sh
❯ version=3.4.5 ./reason_bspack.sh
info esy install 0.2.6
info esy build 0.2.6
info found esy manifests: esy.json
esy: error, exiting...
     invalid dependency @opam/ocaml: unable to resolve package
       processing package: get-js-of-ocaml@xxx
       processing package: @opam/base-ocamlbuild@base
```

jaredly:
@chenglou latest is 0.2.7 -- could you try that?

jordwalke:
@chenglou your error says there's a problem finding `@opam/ocaml`, but nothing should be depending on that (the package doesn't exist). Did you modify the `package.json` before testing? Maybe you forgot to revert your local tweak. (It should always be `ocaml` not `@opam/ocaml` because we host the compilers on npm).

andreypopp:
This looks like some opam 2 metadata got in the sandbox (opam 2 models
OCaml as a dependency and thus we see @opam/OCaml). This warrants trying
with esy@>0.2.6 which can handle opam 2 metadata.

jordwalke:
As long as these tests run on the latest I think this should be good to merge?

chenglou:
The tests don’t run on latest yet

andreypopp:
`esy@0.2.7` was released under `next`, not `latest`. I've just made `esy@0.2.8` to be `latest` so it should work with latest now.

jordwalke:
I think this should no longer be an error since no one will be installing 2.7 from now on. This good to merge? This makes it much easier to test the bspacked form of reason since we don't have to use opam.

chenglou:
Lemme year again when I get a laptop. I don’t know why my attempts keep failing

jaredly:
Ok, fixed it up. Now works with esy `0.2.11`

jordwalke:
I'd love to get this checked in - also to serve as an example of how to use bspack on arbitrary projects with arbitrary dependencies.
Did you intend on adding `bspack.ml` to this? Isn't it generated?
Let me know if it's ready to merge.

chenglou:
I'll test it later. What's the hurry? It's not like this part is every run except by myself

jordwalke:
I occasionally test this just to make sure I didn't break anything. It was always such a pain because it involved opam and a bunch of other things.
No rush though, fi you'd like to hold off.

jaredly:
@jordwalke I did intend to add bspack.ml, and it is generated -- I vendored it here so that you wouldn't have to have the bucklescript source locally

jaredly:
(from my side this is ready to merge)

chenglou:
Trying this again. I got:

```
File "reason_parser.ml", line 5, characters 2-42:
Error: Unbound value MenhirLib.StaticVersion.require_20180905
```

jaredly:
That means it's finding the wrong version of menhir :/ I'll take a look.

jchavarri:
I picked up from where @jaredly left, and Iupdated with latest master: https://github.com/jchavarri/reason/tree/joo. Feel free to incorporate it to this PR to avoid noise, if it's useful.

Most relevant changes:
- Move to `esy.lock` folder format of recents versions of esy
- Added `ppx_derivers` as explicit dev of bspacks, as it's been added as dep of omp recently
- Make sure bspacks uses the same version of menhir as the parser. I ran into the problem about `StaticVersion` [mentioned above](https://github.com/facebook/reason/pull/2153#issuecomment-427267936) but it was solved by cleaning build artifacts from parser and building everything again (so that both parser and bspacks use the same version)

I am blocked on one more error though, it seems the new build config for `reason_parser.mly` doesn't produce the `ml` file on build folder, so I get this:

```
File "src/reason-parser/reason_declarative_lexer.mll", line 70, characters 5-18:
Error: Unbound module Reason_parser
```

What'd be the change needed to produce a `reason_parser.ml` file from the menhir file?

jchavarri:
Managed to fix the menhir issue, there were more paths that had to be migrated from `_build` to  `_esy`.

So both refmt binary and refmt js api seem to be building ok now 🎉  Fixed some issues with omp modules being generated by bspack like `Migrate_parsetree__Ast_404` (picked the fix from https://github.com/facebook/reason/pull/2495/, thanks @ryyppy!)

Also, updated refmt js api to the latest breaking changes after `reerror`.

Same branch: https://github.com/jchavarri/reason/tree/joo

jchavarri:
@jaredly Added another small change so that the script gets bspack.ml from BS github repo: https://github.com/jaredly/reason/pull/3.

Some future tasks could be:

- migration to 4.06 for bs7
- publish refmt-js-api to npm (i think it's [`reason`](https://www.npmjs.com/package/reason)? very confusing...) directly from GH actions, whenever there's a new release or merge to `master`
- maybe publish `bspacked-refmt` to npm with same version using GH actions as well, so it can be consumed from BuckleScript's [esy.json](https://github.com/BuckleScript/bucklescript/blob/master/esy.json) that @ulrikstrid has been working on.

As far as I can tell though, this is ready to merge.

jaredly:
@jchavarri ok I added sandboxes, so now this script works with both 4.02 and 4.06! Very clean

ulrikstrid:
I opened [this PR against bucklescript](https://github.com/BuckleScript/bucklescript/pull/4007) which should allow us to use bspack from upstream instead of vendoring it.

jaredly:
Ok @jordwalke @ulrikstrid @jchavarri I think this is good to go!

ulrikstrid:
What is stopping this from going in? Seems like something that would be nice since the community wants to upstream reason changes into bucklescript more frequently

PR #2157, opened 2018-08-23
Add a test case for transitioning from |. to -> in JSX

bloodyowl:
After having spent some time today trying to upgrade to the new fast pipe syntax through refmt, this is simply to ensure the transition goes smoothly after the next release 😄


chenglou:
Sorry about that, and thanks.

Issue #2165, opened 2018-08-28
Bad type error location for jsx


jordwalke:
The location of type errors are completely missing from this Reason example (I haven't tried a Reason React example - this is on the raw Reason syntax without bs-ppx)

```reason
module AdgroupStore = {
  let get = adgroupID => "THIS IS AN ADGROUP";
};

module PageStore = {
  let get = pageID => "THIS IS A PAGE";
};


let containerize = (~pageID, ~adgroupID, component) => {
  component(~page=PageStore.get(pageID), ~adgroup=AdgroupStore.get(adgroupID), ~children=[], ());
};

module MyComponent = {
  let createElement = (~page, ~adgroup, ~children, ()) => {
    ();
    ();
  };
  let createElement = containerize(createElement);
};


let renderThis = <MyComponent adgroupID="34" pageID="99"/>;

```

There's no location whatsoever on the type error for some reason.

IwanKaramazow:
I'm looking into this together with https://github.com/facebook/reason/pull/2162.

jordwalke:
It's even in the compiler errors.

thangngoc89:
So this error is discovered first in Sketch. In parser reports the JSX position at `[ (0, -1), (0, -1) ]`

IwanKaramazow:
There's a deeper problem, still investigating.
Fixing the `MyComponent` Pexp_ident location alone doesn't fix it in all cases.

PR #2166, opened 2018-08-30
Parse correct location on jsx component name.

IwanKaramazow:
Fixes https://github.com/facebook/reason/issues/2165

Not putting the correct location on the Pexp_ident expression (the component name), results in type errors on line 0.

Example: `<div foo=bar />`
The component name `div` (`{pexp_desc = Pexp_ident}`), needs a correct location in order to have correct type errors. The location was missing before this PR.

With this patch the type errors are reported at the correct location:
image

### DIFF
--- a/src/reason-parser/reason_parser.mly
+++ b/src/reason-parser/reason_parser.mly
@@ -956,23 +956,24 @@ let rewriteFunctorApp module_name elt loc =
     | Lapply (m1, m2) -> mkmod ~loc (Pmod_apply (mkModExp m1, mkModExp m2)) in
   if applies module_name then
     let flat = flattenModName module_name in
-    mkexp(Pexp_letmodule({txt=flat; loc},
+    mkexp ~loc (Pexp_letmodule({txt=flat; loc},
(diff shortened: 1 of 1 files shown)


src/reason-parser/reason_parser.mly
```diff
@@ -2712,20 +2713,21 @@ jsx_arguments:
...
-    LESSIDENT jsx_arguments
-    { let name = Longident.parse $1 in
-      (jsx_component name $2, name)
+    as_loc(LESSIDENT) jsx_arguments
+    { let name = Longident.parse $1.txt in
+      (jsx_component {$1 with txt = name} $2, name)
     }
-  | LESS mod_ext_longident jsx_arguments
-    { jsx_component $2 $3, $2 }
+  | LESS as_loc(mod_ext_longident) jsx_arguments
```
jordwalke: Nice, we got the long ident covered too.

jordwalke:
Thank you! I'm surprised I didn't hit this more often.

PR #2172, opened 2018-09-06
Use updated Menhir in Circle CI

anmonteiro:
We can now use an updated Menhir for CI tests given that https://github.com/facebook/reason/pull/2145 is in.


chenglou:
Thanks

PR #2176, opened 2018-09-09
rtop Windows Support: Convert bash wrapper to ocaml

### LINKED ISSUE #2167: Make `rtop` work on windows
@bryphe confirmed that `rtop` works on windows as well! The only part that does not is the little shell script wrapper that forms `rtop` binary. Instead, we should make `rtop` compile a native `.exe` that does the same thing as the script wrapper. This would be a huge unblocker for Windows native Reason, and is relatively easy to implement. We just need the `rtop` directory to have a native executable that achieves the following logic:

This is also a great first task.

```sh
#!/usr/bin/env bash

# Copyright (c) 2015-present, Facebook, Inc.
#
# This source code is licensed under the MIT license found in the
# LICENSE file in the root directory of this source tree.


# 0. Touches .utoprc (otherwise utop will fail).
# 1. Avoids "#use"ing ~/.ocamlinit before reason is required.
#    Otherwise, the ~/.ocamlinit is written in `.ml` syntax that will
#    choke the reason parser.
# 2. Add the $HOME directory to the include paths so that in step 3
#     we can actually load the ~/.ocamlinit (but before we require
#     reason).
# 3. Run ml rtop_init.ml that "#use"es ".ocamlinit", and then finally
#    require's reason, so that reason is required after .ocamlinit is
#    parsed in standard syntax.

touch $HOME/.utoprc
touch $HOME/.utop-history
DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )

# intercept the -stdin flag of utop, and preprocess the code into reason code
# before forwarding it. Afaik currently there's no better way of intercepting
# the code after urtop receives code from stdin and before it processes it
if [[ $@ =~ "stdin" ]]; then
    refmt --parse re --print ml --interface false | utop-full $@
else
    utop-full -init $DIR/rtop_init.ml $@ -I $HOME -safe-string
fi

```

ulrikstrid:
Fixes #2167 

I think the file should be named something other than `rtop.ml` but I can't get dune to work with anything else.


src/rtop/rtop.ml
```diff
@@ -0,0 +1,31 @@
...
+(* Get the dir of the current executable *)
+let dir = Filename.dirname Sys.argv.(0) in
+
+(* If there is only 1 arg it's the exeuctable and we can remove that *)
+let argv = match Array.length Sys.argv with
+| 1 -> [||]
+| argLength -> Array.sub Sys.argv 1 argLength
+in
+
+let args = if Array.length argv == 1 && Sys.argv.(0) == "stdin" then
```
ulrikstrid: Change `Sys.argv` to `argv`

src/rtop/rtop.ml
```diff
@@ -0,0 +1,31 @@
...
+  let refmted = Pervasives.input_line (Unix.open_process_in "refmt --parse re --print ml --interface false") in
+
+  Array.concat [[|"utop";|]; argv; [|refmted|]]
+else
+  Array.concat [[|"utop"; "-init"; dir ^ "/rtop_init.ml";|]; argv; [|"-I"; home; "-safe-string"|]]
+in
+
+Unix.execvp "utop" args
```
ulrikstrid: Should be `utop-full`

src/rtop/rtop.ml
anmonteiro: You also need to change the `==` in the string comparison to `=`, otherwise it's always false

```ocaml
utop # "foo" == "foo";;
- : bool = false
```

src/rtop/rtop.ml
```diff
@@ -0,0 +1,36 @@
...
+| 1 -> [||]
+| argLength -> Array.sub Sys.argv 1 argLength
+in
+
+let args = if Array.length argv == 1 && argv.(0) = "stdin" then
+  let refmted = Pervasives.input_line (Unix.open_process_in "refmt --parse re --print ml --interface false") in
+
+  Array.concat [[|"utop-full";|]; argv; [|refmted|]]
+else
+  Array.concat [[|"utop-full"; "-init"; dir ^ "/rtop_init.ml";|]; argv; [|"-I"; home; "-safe-string"|]]
```
anmonteiro: I would probably use `Filename.concat` here to avoid path separator issues across platforms.

src/rtop/rtop.ml
```diff
@@ -0,0 +1,36 @@
...
+in
+
+(* Try to read the HOME environment
+ * if it is not set we assume it's windows and fallback to USERPROFILE *)
+let home = match getenv_opt "HOME" with
+| Some(p) -> p
+| None -> Sys.getenv "USERPROFILE"
+in
+
+let () = Unix.openfile (home ^ "/.utoprc") [] 0o640
```
anmonteiro: `Filename.concat` here too, and wherever you're joining paths.

src/rtop/rtop.ml
```diff
@@ -0,0 +1,36 @@
...
+|> Unix.close in
+
+let () = Unix.openfile (home ^ "/.utop-history") [] 0o640
+|> Unix.close in
+
+(* Get the dir of the current executable *)
+let dir = Filename.dirname Sys.argv.(0) in
```
hcarty: Copying a suggestion from discord - it's probably better to use `Sys.executable_name` here.  It has some extra logic to try to take advantage of OS-specific methods for determining the actual executable name.  It's not perfect but it is more accurate that `Sys.argv.(0)` in some cases.  https://caml.inria.fr/pub/docs/manual-ocaml/libref/Sys.html#VALexecutable_name

vbgl:
This fails when there is no `~/.utoprc` file:

> Fatal error: exception Unix.Unix_error(Unix.ENOENT, "open", "/…/.utoprc")

PR #2182, opened 2018-09-14
Update Getting Started link to point to the new page

tehlordvortex:
The location of the Getting Started guide changed, and the previous link simply loads a page that points users to the new page. Might as well just take them directly to the new location.


chenglou:
Thanks!

Issue #2189, opened 2018-09-20
[rtop] Odd printing of polymorphic variants


ostera:
**Rtop Session**
```reason
Reason # let f = fun | `Ok(x) => x == 1;
let f: [< `Nope | `Ok(int) ] => bool = <fun>;
Reason # let g = fun | `Ok(x) => x == "!!!" ;
let g: [< `Ok(string) ] => bool = <fun>;
Reason # let h = x => f(x) && g(x);
let h: [< `Okstring & int ] => bool = <fun>;
```

**Expected**
```reason
Reason # let f = fun | `Ok(x) => x == 1;
let f: [< `Nope | `Ok(int) ] => bool = <fun>;
Reason # let g = fun | `Ok(x) => x == "!!!" ;
let g: [< `Ok(string) ] => bool = <fun>;
Reason # let h = x => f(x) && g(x);
let h: [< `Ok(string & int) ] => bool = <fun>;
```

anmonteiro:
This is probably a bug in the outcome printer cc @IwanKaramazow

PR #2191, opened 2018-09-21
Fix outcome printing of polymorphic variants row fields

IwanKaramazow:
Fixes https://github.com/facebook/reason/issues/2189

### DIFF
--- a/src/reason-parser/reason_oprint.ml
+++ b/src/reason-parser/reason_oprint.ml
@@ -493,7 +493,9 @@ and print_row_field ppf (l, opt_amp, tyl) =
     else fprintf ppf "" in
   let parens = match tyl with
     | [ (Otyp_tuple _) ] -> false (* tuples already have parentheses *)
-    | [ _ ] -> true
+    (* [< `Ok(string & int) ] ----> string & int
+     * [< `Ok(string) ] -----> string *)
+    | _::_ -> true
     | _ -> false in
   fprintf ppf "@[<hv 2>`%s%t%s%a%s@]"
     l


src/reason-parser/reason_oprint.ml
```diff
@@ -493,7 +493,9 @@ and print_row_field ppf (l, opt_amp, tyl) =
     else fprintf ppf "" in
   let parens = match tyl with
     | [ (Otyp_tuple _) ] -> false (* tuples already have parentheses *)
-    | [ _ ] -> true
+    (* [< `Ok(string & int) ] ----> string & int
+     * [< `Ok(string) ] -----> string *)
+    | _::_ -> true
```
ostera: Will this work for only a pair of types or for more? Would it support this:

```reason
Reason # let f = fun | `Ok(x) => x == 1;
let f: [< `Ok(int) ] => bool = <fun>;
Reason # let g = fun | `Ok(x) => x == "!";
let g: [< `Ok(string) ] => bool = <fun>;
Reason # let h = fun | `Ok(x) => x == true;
let h: [< `Ok(bool) ] => bool = <fun>;
Reason # let i = fun | `Ok(x) => x == 1.0;
let i: [< `Ok(float) ] => bool = <fun>;
Reason # let l = x => f(x) && g(x) && h(x) && i(x);
let l: [< `Okfloat & bool & string & int ] => bool = <fun>;
```

IwanKaramazow: You mean:
image

It'll work for one or more than one. We're pattern matching on a single linked list.
`_::_` matches on:
* x::[] (single item, i.e. [x])
* x::[y] (two items, i.e. [x, y])
* x::[y, z] (3 items, i.e. [x, y, z])
etc.

PR #2195, opened 2018-09-21
Fix slow code in hot path

### LINKED ISSUE #2190: [refmt] hangs & pegs CPU for a certain ocaml file
Here's the file https://gist.github.com/jaredly/5a6eb7c66e5919ac932a053271785d8c
this is version 3.3.3.
No idea what's going on.

anmonteiro:
fixes #2190

The `consolidateSeparator` function was recursing on `Whitespace`
nodes, however `preOrderWalk` would already do that for us. Therefore,
we were calling this costly function in a hot path many more times than
necessary, effectively making it take exponential time (I think).


IwanKaramazow:
👍Good find, approved!

Issue #2134, opened 2018-08-13
"Attributes not allowed here" when using ppx_inline_test


andreypopp:
Getting
```
File "_none_", line 1:
Error: Attributes not allowed here
```
when trying to use ppx_inline_test:
```
let%test "some" = true;
```

Tried with 3.3.3 and master (as of 10dc9e99c955cfb4a4fdcb079dc1a891f3eb71a0).

Repro here https://github.com/andreypopp/reason-ppx-inline-test

andreypopp:
Looks like the issue is that ppx_inline_test (and also ppx_expect) is too strict with how it matches AST nodes.

I think on Reason side we could add location to `[@reason.literal_value "..."]` attribute so we don't end up with confusing:
```
File "_none_", line 1:
```
location.

PR #2197, opened 2018-09-21
keep `let not = ...` 

### LINKED ISSUE #2169: [refmt] not -> (!)
`let not = ...` is converted to `let (!) = ...` but I need exactly `not`:)

anmonteiro:
fixes #2169


anmonteiro:
This now also fixes #2134 because OCaml versions prior to 4.06 (maybe 4.05?) can't parse `let (foo [@attribute_here]) = ...`.

As such, we remove the (only the) literal attributes before printing to either OCaml binary or to OCaml.

anmonteiro:
I think this is ready for a final look now.

PR #2201, opened 2018-09-26
Fix precedence issue with unary ops and labelled args

### LINKED ISSUE #2200: Precedence issue with unary minus and named arg
The title says it all. Here's a min repro:
```reason
translate(
      ~x=-. x +. 10,
      ~y=-. y +. 10,
    );
```
Is turned into
```reason
translate(~x=-. (x +. 10), ~y=-. (y +. 10));
```

AST with named parameters:
```
[
  structure_item (test.re[1,0+0]..[4,53+5])
    Pstr_eval
    expression (test.re[1,0+0]..[4,53+5])
      Pexp_apply
      expression (test.re[1,0+0]..[1,0+9])
        Pexp_ident "translate" (test.re[1,0+0]..[1,0+9])
      [
        <label> "x"
          expression ([0,0+-1]..[0,0+-1])
            Pexp_apply
            expression (test.re[2,11+8]..[2,11+11])
              Pexp_ident "~-." (test.re[2,11+8]..[2,11+11])
            [
              <label> ""
                expression (test.re[2,11+12]..[2,11+19])
                  Pexp_apply
                  expression (test.re[2,11+14]..[2,11+16])
                    Pexp_ident "+." (test.re[2,11+14]..[2,11+16])
                  [
                    <label> ""
                      expression (test.re[2,11+12]..[2,11+13])
                        Pexp_ident "x" (test.re[2,11+12]..[2,11+13])
                    <label> ""
                      expression (test.re[2,11+17]..[2,11+19])
                        Pexp_constant Const_int 10
                  ]
            ]
        <label> "y"
          expression ([0,0+-1]..[0,0+-1])
            Pexp_apply
            expression (test.re[3,32+8]..[3,32+11])
              Pexp_ident "~-." (test.re[3,32+8]..[3,32+11])
            [
              <label> ""
                expression (test.re[3,32+12]..[3,32+19])
                  Pexp_apply
                  expression (test.re[3,32+14]..[3,32+16])
                    Pexp_ident "+." (test.re[3,32+14]..[3,32+16])
                  [
                    <label> ""
                      expression (test.re[3,32+12]..[3,32+13])
                        Pexp_ident "y" (test.re[3,32+12]..[3,32+13])
                    <label> ""
                      expression (test.re[3,32+17]..[3,32+19])
                        Pexp_constant Const_int 10
                  ]
            ]
      ]
]
```

```reason
translate(
      -. x +. 10,
      -. y +. 10,
    );
```

The AST without named parameters:
```
[
  structure_item (test.re[1,0+0]..[4,47+5])
    Pstr_eval
    expression (test.re[1,0+0]..[4,47+5])
      Pexp_apply
      expression (test.re[1,0+0]..[1,0+9])
        Pexp_ident "translate" (test.re[1,0+0]..[1,0+9])
      [
        <label> ""
          expression (test.re[2,11+6]..[2,11+16])
            Pexp_apply
            expression (test.re[2,11+11]..[2,11+13])
              Pexp_ident "+." (test.re[2,11+11]..[2,11+13])
            [
              <label> ""
                expression (test.re[2,11+6]..[2,11+10])
                  Pexp_apply
                  expression (test.re[2,11+6]..[2,11+8])
                    Pexp_ident "~-." (test.re[2,11+6]..[2,11+8])
                  [
                    <label> ""
                      expression (test.re[2,11+9]..[2,11+10])
                        Pexp_ident "x" (test.re[2,11+9]..[2,11+10])
                  ]
              <label> ""
                expression (test.re[2,11+14]..[2,11+16])
                  Pexp_constant Const_int 10
            ]
        <label> ""
          expression (test.re[3,29+6]..[3,29+16])
            Pexp_apply
            expression (test.re[3,29+11]..[3,29+13])
              Pexp_ident "+." (test.re[3,29+11]..[3,29+13])
            [
              <label> ""
                expression (test.re[3,29+6]..[3,29+10])
                  Pexp_apply
                  expression (test.re[3,29+6]..[3,29+8])
                    Pexp_ident "~-." (test.re[3,29+6]..[3,29+8])
                  [
                    <label> ""
                      expression (test.re[3,29+9]..[3,29+10])
                        Pexp_ident "y" (test.re[3,29+9]..[3,29+10])
                  ]
              <label> ""
                expression (test.re[3,29+14]..[3,29+16])
                  Pexp_constant Const_int 10
            ]
      ]
]
```

cc @anmonteiro @IwanKaramazow @chenglou

anmonteiro:
fixes #2200


bsansouci:
Awesome! Thanks for fixing. Any way we can avoid the parens around the number with unary minus since we know how to parse it correctly?

jordwalke:
After doing a little archaeology on this set of changes, I'm thinking this could just adding more complexity to something that was not too terrible to begin with.

This original diff [1642](https://github.com/facebook/reason/pull/1642) worked correctly right?

It had only one additional lexing rule, and didn't even need to use any `set_lexeme_length` tricks. Are there bugs with that version of the solution? On that diff @IwanKaramazow suggested that `set_lexeme_length` could be used to implement the same some things, and then that suggestion was tried in diff [1945](https://github.com/facebook/reason/pull/1945) attempted that. Then those issues required some massaging of the AST to make up for it in this diff.

Did diff 1642 have issues or bugs? It seems like the cleanest solution of them all, so long as it does't have any bugs of course.

jordwalke:
The reason why this is getting so messy, is because unary parsing precedence is very tricky and has special behavior. You can see how it is designated as special in all the precedence rules. To lex a token, which is then parsed as an infix operator `=-.`, and _then_ split it up into an equal sign but then try to treat it as a unary `-.`, is going to be full of tons of bugs because it's too late in the parser - the opportunity to receive the especial behavior of unary parsing precedence is lost. For these kinds of things, the `set_lexeme_length` is a really powerful hack that can keep the rest of the parser sane.

Diff [1642](https://github.com/facebook/reason/pull/1642/files) which special cased `~lbl=` was pretty clean. There were other unrelated parsing issues like `let x=-.1` so [1945](https://github.com/facebook/reason/pull/1945) was created. That diff 1945 tried to kill two birds with one stone, and maybe that just wasn't the best fit - maybe `let x=-.1` warrants a different solution - one that won't require us to put all these patches into the parser .mly phase?

jordwalke:
When we see `let x=-.1`, we really do wish we could split the `=` from the `-.1`, but not split it in the case of `let _ = x=-.1`. That seems like a really hard problem.

jordwalke:
Would it be cleaner to bring back the special token for `~label=` since that avoided problems with labeled args, then introduce another special token for `let ident =`? That's pretty out there, but I think it would cover all the cases.

anmonteiro:
I just tested this with the fix in #1642 and it suffers from the same problem.

anmonteiro:
Indeed #1642 doesn't suffer from this bug. I'll refactor the PR to include that approach.

anmonteiro:
After some more investigation, I think this is indeed a very hard problem.

With the approach employed by #1642, these wouldn't parse:

```reason
let x = (~a: int=- 1) => a;

let x: float=-.1;
```

The reason being the type constraint. And making yet another special case in the lexer / parser to support this would IMO not justify the cost.

bsansouci:
Whatever works, you guys are the specialists! What I would say is that I hope nobody _ever ever ever_ uses `=-.` as an infix. And if we wanted to ban that infix forever, I wouldn't mind ;)

jordwalke:
I'm glad you went with this general approach (thanks for the suggestion @IwanKaramazow). This helps avoid messing with the complicated precedence rules of shift/reduce and handles the squashing at an earlier stage. These earlier hacks to cover less important edge cases keep the overall grammar rules comprehensible. Similar to the approach with optional semicolons we could say that `~label=-.1` isn't _actually_ valid Reason code, but that we go and do the user a favor by correcting some common pitfalls. Thanks as always @anmonteiro

PR #2204, opened 2018-10-01
Don't install the oprintTest executable

### LINKED ISSUE #2112: oprintTest is installed
The executabe oprintTest used for the tests are installed when the package is installed. As far as I know the only thing required to avoid this is removing `(public_names (testOprint))` from the jbuild file. I have not tested this.

zploskey:
Fixes #2112.


jordwalke:
Thank you!

PR #2209, opened 2018-10-04
Windows: Fix building with 'esy' and add AppVeyor support

bryphe:
__Issue:__ When trying to build reason via `esy` on Windows:
- `git clone https://github.com/facebook/reason`
- `esy install`
- `esy build`

We get this error:
```
error: build failed
  build log:
    # esy-build-package: building: @esy-ocaml/esy-installer@0.0.0
    # esy-build-package: running: "make" "install"
    mkdir -p C:/esy-test-1/a4bc0cfe0d85f8939b916466d53e26881f06a767/3_/s/esy_ocaml__s__esy_installer-0.0.0-27355c1d/bin
    cp -r bin/esy-installer C:/esy-test-1/a4bc0cfe0d85f8939b916466d53e26881f06a767/3_/s/esy_ocaml__s__esy_installer-0.0.0-27355c1d/bin/esy-installer
    cp -r vendor-unknown/bin/* C:/esy-test-1/a4bc0cfe0d85f8939b916466d53e26881f06a767/3_/s/esy_ocaml__s__esy_installer-0.0.0-27355c1d/bin/opam-installer
    cp: cannot stat 'vendor-unknown/bin/*': No such file or directory
    make: *** [Makefile:19: install] Error 1
    error: command failed: "make" "install"
    esy-build-package: exiting with errors above...
```

The problem is that `esy-installer@0.0.0` doesn't support Windows, and the latest version is `0.0.1`, which has the fix. (Also, the behavior for `0.0.x` version numbers is not very intuitive: https://github.com/npm/node-semver/issues/79 - so we have to bump up the version in package.json, in addition to regenerating the lockfile).

I believe we could also remove the `esy-installer` dependency completely, since `esy` natively supports that command now - is that correct @andreypopp ?


bryphe:
Submitted CLA-

andreypopp:
> I believe we could also remove the esy-installer dependency completely, since esy natively supports that command now - is that correct @andreypopp ?

Yes, esy implements `esy-installer` command natively now.

bryphe:
I added an initial `appveyor.yml` to run the build - log here: https://ci.appveyor.com/project/bryphe/reason/builds/19379821 (green! 👍 )

Trying out the `esy b make test` step to see if the tests will work there too.

appveyor.yml
```diff
@@ -0,0 +1,16 @@
...
+    - C:\Users\appveyor\.esy\3_\i
+    - C:\Users\appveyor\.esy\source-tarballs
+
+install:
+    - npm install -g esy@0.3.0
+    - appveyor-retry esy install
+
+build_script:
+    - esy build
+    - esy b make test
```
anmonteiro: this is not truly using `esy` and its sandbox because `make test` is still building with dune. Additionally it's building the oprint test binary (should perhaps also be built with Esy too?)

bryphe: > this is not truly using esy and its sandbox because make test is still building with dune. 

Hmm, I'm not sure what you mean here - `esy b` means run a command in the esy build environment. It's often the case we'll run dune in the build environment via `esy`!😄 

But unfortunately this isn't sufficient to run the tests on Windows - here's the error log from AppVeyor: 
```
node ./formatTest/testOprint.js
bash: node: command not found
```
(`node` isn't present in the esy sandbox, so it fails)

In any case, I'll leave this out for now and track it separately - there's more work to get the tests running. 

Moving the oprint test binary outside of the Makefile as you suggested would be a good start; we'd also need to refactor the bash-driven tests to something that would work well in Windows too. Looks like even the node tests make some assumptions about being on a POSIX style platform (ie, they call `cat` which isn't around by default on Windows)

anmonteiro: @bryphe if you rebase on master it should work! I recently eliminated the need for Node.js in https://github.com/facebook/reason/pull/2213

anmonteiro: Ah, with respect to the other POSIX problems I can't help.

bryphe: Ah cool, thanks @anmonteiro ! I'll try it out - would be great to have some tests running on Windows.

bryphe:
I opened an issue tracking blockers to having the tests run here: https://github.com/facebook/reason/issues/2224

I think it makes sense to bring this in as-is, so we have the windows build running on CI via `esy`, and then we can fix the tests in a PR against #2224 as a next step. LMK if that sounds reasonable @IwanKaramazow @anmonteiro

bryphe:
Latest run for 354e310 passed on AppVeyor: https://ci.appveyor.com/project/bryphe/reason/builds/19381912

appveyor.yml
bryphe: More details here on test blockers: #2224 . Thanks for looking the change over @anmonteiro ! 👍

andreypopp:
@bryphe can we use `@opam/merlin` instead of `@esy-ocaml/merlin`, I think it's more recent.

jordwalke:
Do you think this is reliable enough to be merged in now?

bryphe:
@jordwalke - yes, this project gets built for `esy` too on Windows, and has been consistently green on Windows. 👍 I have it building on my local appveyor account here: https://ci.appveyor.com/project/bryphe/reason

Once this PR is in, we'll need to enable this in an AppVeyor account for facebook (we could also it turn it in the `esy` account, if that would be preferred).

bryphe:
@jordwalke - perhaps we should revisit this and move to an Azure CI strategy? Are you OK with that? And if so - is there already an Azure CI account set up for `reason`?

jordwalke:
Created an Azure org for Reason (called reasonml) and added you to it so feel free to add it there and/or close this issue.

bryphe:
Thanks @jordwalke ! I'll close this out and switch strategies to Azure CI.

Issue #2216, opened 2018-10-07
[refmt] Fragment + local open


cknitt:
refmt reformats

```re
ReasonReact.(<> {string("Test")} </>);
```

to

```re
ReasonReact.<> {string("Test")} </>;
```

which gives a syntax error.

PR #2217, opened 2018-10-07
Format fragments in context of local opens correct

IwanKaramazow:
Fixes https://github.com/facebook/reason/issues/2216

**Before**
```reason
ReasonReact.<> {ReasonReact.string("test") } </>
```

**After**
```reason
ReasonReact.(<> {ReasonReact.string("test") } </>)
```

### DIFF
--- a/src/reason-parser/reason_pprint_ast.ml
+++ b/src/reason-parser/reason_pprint_ast.ml
@@ -4164,17 +4164,18 @@ let printer = object(self:'self)
      *
      * (Also see https://github.com/facebook/Reason/issues/114)
      *)
-    match e.pexp_desc with
-    | Pexp_record _ (* syntax sugar for M.{x:1} *)
-    | Pexp_tuple _ (* syntax sugar for M.(a, b) *)
-    | Pexp_object {pcstr_fields = []} (* syntax sugar for M.{} *)
-    | Pexp_construct ( {txt= Lident"::"},Some _)
-    | Pexp_construct ( {txt= Lident"[]"},_)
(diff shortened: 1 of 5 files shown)


src/reason-parser/reason_pprint_ast.ml
```diff
@@ -4164,17 +4164,18 @@ let printer = object(self:'self)
...
      * (Also see https://github.com/facebook/Reason/issues/114)
      *)
-    match e.pexp_desc with
-    | Pexp_record _ (* syntax sugar for M.{x:1} *)
-    | Pexp_tuple _ (* syntax sugar for M.(a, b) *)
-    | Pexp_object {pcstr_fields = []} (* syntax sugar for M.{} *)
-    | Pexp_construct ( {txt= Lident"::"},Some _)
-    | Pexp_construct ( {txt= Lident"[]"},_)
-    | Pexp_extension ( {txt = "bs.obj"}, _ ) ->
+    match e.pexp_attributes, e.pexp_desc with
```
anmonteiro: I think you may want `stdAttrs`

IwanKaramazow: `ReasonReact.{<> {ReasonReact.string("test") } </>}` contains no `stdAttrs` but has separate `JSX` attributes.
```
    | [], Pexp_construct ( {txt= Lident"::"},Some _)
````
This branch matches when there's only a jsx attribute in the case of jsx. (no std attrs).

Should I clarify the code or rewrite the pattern match?

PR #2219, opened 2018-10-08
Improve printing of jsx props with Pexp_apply + spread with Pexp_apply/Pexp_fun

IwanKaramazow:
* inline Pexp_apply where possible with brace hugging
* improve formatting of spread children with Pexp_apply & Pexp_fun

image


alexfedoseev:
Thank you for this change! –2 stops is huge.

PR #2222, opened 2018-10-09
Stop vendoring libraries

anmonteiro:
we have other dependencies, and I'm quite certain this is a historical
artifact.

It's also less code to maintain.


chenglou:
We need those for bspacking

jordwalke:
@anmonteiro I thought we made some modifications to Easy_format but if the tests pass then that must not have been the case.

@chenglou do we actually need these two for bspacks? I thought these ones would be the most straightforward ones to bspack but I could be wrong.

@jaredly's diff for using the esy sources during bspacks should cover this. Let me know if I'm not considering something.

One nice thing about vendoring cmdliner though is that it ensures you never get a version conflict when someone else depends on a different version range of cmdliner (since it's such a ubiquitous package - that's one nice perk of vendoring).

anmonteiro:
I'll leave this open for now because I think it'd be a nice changeset – less maintenance burden. Feel free to close if you think it's not that useful.

jordwalke:
Yeah, I'm torn about this one @anmonteiro - I think it shouldn't be a burden for bs-packs with @jaredly's changes (we could confirm) but on the other hand vendoring very commonly used libraries can help avoid package manager conflicts.

PR #2223, opened 2018-10-09
Preserve function body braces

anmonteiro:
This is an alternative to #1826 implementing just the brace preservation
code, to make it easier to be reviewed.

I think this is now in a better place to get into the codebase given
that we have `Reason_attributes` outside of `Reason_pprint_ast` and that
we now remove our own attributes before printing to OCaml.

I think we can probably refactor this code further in future PRs, as
well as add other stylistic preservation changes.

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/oo.re
+++ b/formatTest/typeCheckedTests/expected_output/oo.re
@@ -14,12 +14,18 @@ class virtual stack ('a) (init) = {
       Some(hd);
     | [] => None
     };
-  pub push = hd => v = [hd, ...v];
-  initializer (
-    print_string("initializing object")
-  );
-  pub explicitOverrideTest = a => a + 1;
-  pri explicitOverrideTest2 = a => a + 1;
+  pub push = hd => {
+    v = [hd, ...v];
+  };
+  initializer {
+    print_string("initializing object");
+  };
+  pub explicitOverrideTest = a => {
+    a + 1;
+  };
+  pri explicitOverrideTest2 = a => {
+    a + 1;
+  };
 };
 
 let tmp = {
@@ -50,10 +56,12 @@ class virtual stackWithAttributes ('a) (init) = {
       Some(hd);
     | [] => None
     };
-  pub push = hd => v = [hd, ...v];
-  initializer (
-    print_string("initializing object")
-  );
+  pub push = hd => {
+    v = [hd, ...v];
+  };
+  initializer {
+    print_string("initializing object");
+  };
 };
 
 class extendedStack ('a) (init) = {
@@ -67,9 +75,15 @@ class extendedStackAcknowledgeOverride
       (init) = {
   inherit (class stack('a))(init);
   val dummy = ();
-  pub implementMe = i => i + 1;
-  pub! explicitOverrideTest = a => a + 2;
-  pri! explicitOverrideTest2 = a => a + 2;
+  pub implementMe = i => {
+    i + 1;
+  };
+  pub! explicitOverrideTest = a => {
+    a + 2;
+  };
+  pri! explicitOverrideTest2 = a => {
+    a + 2;
+  };
 };
 
 let inst = (new extendedStack)([1, 2]);
@@ -133,8 +147,12 @@ let anonClosedObject: {
   x: int,
   y: int,
 } = {
-  pub x = 0;
-  pub y = 0
+  pub x = {
+    0;
+  };
+  pub y = {
+    0;
+  }
 };
 
 let onlyHasX = {pub x = 0};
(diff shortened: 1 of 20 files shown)


src/reason-parser/reason_heuristics.ml
```diff
@@ -102,6 +102,13 @@ let isFastPipe e = match Ast_404.Parsetree.(e.pexp_desc) with
     {pexp_desc = Pexp_ident({txt = Longident.Lident("|.")})},
       _
     ) -> true
+  (* | Pexp_apply(
+    {pexp_desc = Pexp_apply(
+      {pexp_desc = Pexp_ident({txt = Longident.Lident("|.")})},
+        _
+      )},
+      _
+  ) -> Printf.eprintf "YEP\n%!";true *)
```
IwanKaramazow: Can be removed.

src/reason-parser/reason_pprint_ast.ml
```diff
@@ -1747,8 +1747,23 @@ let partitionFinalWrapping listTester wrapFinalItemSetting x =
...
+let should_preserve_requested_braces expr =
+  let {literalAttrs} = partitionAttributes expr.pexp_attributes in
+  match expr.pexp_desc with
+  | Pexp_ifthenelse _
+  | Pexp_try _ -> false
+  | _ -> Reason_attributes.has_preserve_braces_attrs literalAttrs
+
 let semiTerminated term = makeList [term; atom ";"]
 
+let makeBreakableBlock ?(inline=false) letItems =
```
IwanKaramazow: I'm not a huge fan of abstractions on top of `makeList`. The exact behaviour of `makeBreakableBlock` is very hard to predict without inspecting its implementation.
`makeBreakableBlock` seems to be used only once too. Is there a reason why you didn't use `makeList` with all named args in the letList printing? Just using `makeList` with all named args makes it very clear what is going to happen.

jaredly:
yayyyyyyy

src/reason-parser/reason_attributes.ml
```diff
@@ -38,6 +38,9 @@ let rec partitionAttributes ?(partDoc=false) ?(allowUncurry=true) attrs : attrib
   | (({txt="reason.raw_literal"}, _) as attr) :: atTl ->
     let partition = partitionAttributes ~partDoc ~allowUncurry atTl in
     {partition with literalAttrs=attr::partition.literalAttrs}
+  | (({txt="reason.preserve_braces"}, _) as attr) :: atTl ->
+    let partition = partitionAttributes ~partDoc ~allowUncurry atTl in
+    {partition with literalAttrs=attr::partition.literalAttrs}
```
jordwalke: `literalAttrs` was originally intended to represent string literals. It seems you're repurposing them here? The name should probably be changed accordingly.

jordwalke:
Is it currently the case that these (and other stylistic attributes such as string "literal" attributes) are removed when the `--print` is `binary`? I believe that should be the case and would fix the other issues we've had with ppx plugins complaining.

anmonteiro:
@jordwalke yes, we remove `literalAttrs` from printing.

I agree with your comment that the name should probably be changed.

formatTest/typeCheckedTests/expected_output/oo.re
```diff
@@ -14,12 +14,18 @@ class virtual stack ('a) (init) = {
...
-  pub push = hd => v = [hd, ...v];
-  initializer (
-    print_string("initializing object")
-  );
-  pub explicitOverrideTest = a => a + 1;
-  pri explicitOverrideTest2 = a => a + 1;
+  pub push = hd => {
+    v = [hd, ...v];
+  };
+  initializer {
```
IwanKaramazow: What's the rational behind printing braces here, while the original had parens?

formatTest/unit_tests/expected_output/modules.re
```diff
@@ -515,24 +516,42 @@ module M = {
 
 module N = {
   open M;
-  let z = M.(34);
+  let z = {
+    M.(34);
```
IwanKaramazow: Why does this local open sugar have braces?

formatTest/unit_tests/expected_output/modules.re
```diff
@@ -515,24 +516,42 @@ module M = {
...
+  };
   let z = {
     open M;
     34;
     35;
   };
+  let z = {
+    M.(34, 35);
```
IwanKaramazow: Same for every local open sugar, is this part of this PR? I'm not sure that we want braces if there's an open sugar transformation happening in printing

formatTest/unit_tests/expected_output/modules.re
anmonteiro: This is how the original input file looked.

formatTest/typeCheckedTests/expected_output/oo.re
jordwalke: @anmonteiro ?

jaredly: The original has braces: https://github.com/anmonteiro/reason/blob/373ed5f64d2f2a389e5a507270178dc5b1bf7efa/formatTest/typeCheckedTests/input/oo.re#L22

PR #2226, opened 2018-10-12
Set up error messages properly in non-recoverable mode

### LINKED ISSUE #2114: UNKNOWN SYNTAX ERROR on infix functions with missing arguments
Minimal example:

```
 "foo" ++
```

anmonteiro:
This PR takes advantage of the currently set up infrastructure for recoverable mode and avoids bailing out whenever we need to handle an error. Instead, it keeps reducing until we catch an exception with the right kind of error coming from the parser.

This enables many more error messages for free that were disabled until now.

Also fixes #2114 for example.

### DIFF
--- a/src/reason-parser/reason_toolchain.ml
+++ b/src/reason-parser/reason_toolchain.ml
@@ -686,19 +686,35 @@ module Reason_syntax = struct
       let process_invalid_docstring (text, loc) = Reason_lexer.add_invalid_docstring text loc in
       List.iter process_invalid_docstring (List.rev docstrings)
 
+  let custom_error supplier env =
+    let loc = last_token_loc supplier in
+    let token = match supplier.last_token with
+      | Some token -> token
+      | None -> assert false
+    in
+    let state = I.current_state_number env in
+    (* Check the error database to see what's the error message
+    associated with the current parser state  *)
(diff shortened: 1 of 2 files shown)


IwanKaramazow:
How big would you estimate the cost of:
image

src/reason-parser/reason_toolchain.ml
```diff
@@ -719,21 +735,11 @@ module Reason_syntax = struct
...
-      (* Check the error database to see what's the error message
-       * associated with the current parser state *)
-      let msg = Reason_parser_explain.message env token in
-      let msg_with_state = Printf.sprintf "%d: %s" state msg in
-      raise (Reason_syntax_util.Error (loc, (Reason_syntax_util.Syntax_error msg_with_state)))
-
+    | I.HandlingError _env as error_checkpoint ->
+      (* If not in a recoverable state, resume just enough to be able to
+       * catch a nice error message above. *)
+      let cp = I.resume error_checkpoint in
```
IwanKaramazow: Are there cases where we don't want to resume when we're handling an error?

anmonteiro: I think there are, but I'm not sure which heuristics to go for here.

PR #2246, opened 2018-10-21
Add synopsis and description to rtop.opam

anmonteiro:
Opam-repository requires it.


anmonteiro:
This should also have been included in #2244 but somehow I missed it.

PR #2254, opened 2018-11-01
Support single line comments of the form `// comment`

IwanKaramazow:
Implements parsing & printing of single line comments.

Example:
```reason
// single line comment
let x = 1;
```

The main logic is taken from https://github.com/facebook/reason/pull/1800.
Extra test cases and support for whitespace interleaving have been implemented.

### DIFF
--- a/formatTest/unit_tests/expected_output/basicStructures.re
+++ b/formatTest/unit_tests/expected_output/basicStructures.re
@@ -146,7 +146,7 @@ let x = arr^[0] = 1;
  *============================================================================
  */;
 
-let (/\/) = (+); /* // is not a comment */
+let (/++) = (+); /* // indicates the start of a comment, not an infix op */
 
 let something =
   if (self.ext.logSuccess) {
--- a/formatTest/unit_tests/input/basicStructures.re
+++ b/formatTest/unit_tests/input/basicStructures.re
@@ -134,7 +134,7 @@ let x = arr^[0] = 1;
  *============================================================================
  */;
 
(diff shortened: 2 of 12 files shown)


formatTest/unit_tests/expected_output/basicStructures.re
```diff
@@ -146,7 +146,7 @@ let x = arr^[0] = 1;
  *============================================================================
  */;
 
-let (/\/) = (+); /* // is not a comment */
+let (/++) = (+); /* // is not a comment */
```
kpsuperplane: Forgive me I'm pretty new to Reason but, I think the comment about `// is not a comment` should be updated?

formatTest/unit_tests/input/basicStructures.re
```diff
@@ -134,7 +134,7 @@ let x = arr^[0] = 1;
  *============================================================================
  */;
 
-let (//) = (+); /* // is not a comment */
+let (/++) = (+); /* // is not a comment */
```
kpsuperplane: Ditto on comment about not a comment

formatTest/unit_tests/expected_output/basicStructures.re
IwanKaramazow: Thanks! Updated.

formatTest/unit_tests/input/basicStructures.re
IwanKaramazow: Also updated.

PR #2263, opened 2018-11-15
Add `let!Foo` sugar mark 2

jaredly:
Ok, so this is much simpler. This doesn't include the ppx transform yet.

Why go with `!` instead of `.`?

1) having `let.` feels like an overload of attribute access or something, which I think is confusing
2) I also want it to be visually "louder"
2) F# uses `!`


jaredly:
closing in favor of https://github.com/facebook/reason/pull/2487

PR #2264, opened 2018-11-17
Switch from double to single quotes for sharpop access sugar

anmonteiro:


anmonteiro:
@jordwalke we're basically switching from a hack to another, because I didn't find a way to make this work at the lexer level. Here's what I tried and why it didn't work:

- lexing idents without quotes separate from idents with quotes
I though this would simply work but I found that the lexer is still greedy so `foo['bar']` would be lexed as `ident LBRACKET QUOTE ident_with_quote RBRACKET` so the separate lexing was basically useless..

- lexing idents entirely without quotes and recovering them at the parser level, e.g. `ident: LIDENT QUOTE+`
this caused some shift/reduce and reduce/reduce conflicts because of things like parsing poly_types, where we have `preceded(QUOTE,ident)` and we wouldn't be able to figure out when one ident ended and another quote started.

jordwalke:
I was pretty sure you could implement this by either extending [this rule here](https://github.com/facebook/reason/blob/master/src/reason-parser/reason_lexer.mll#L789
) to allow multiple characters between the single quotes, or to add another rule immediately after it that has `> 1` char so that you have two rules that are mutually exclusive. Either way, you could add a custom syntax error message any time someone uses a >1 character single quote string anywhere but directly in array access syntax.

anmonteiro:
Added a commit that cleans up the implementation and implements it in the lexer.

I'm working on adding an error message whenever `'foo'` is used outside of brackets. My current thinking is that I'll need to add it in the recovery section (`reason_toolchain.ml`) rather than the parser because including it in `expr` would result in grammar conflicts.

@IwanKaramazow wondering if you have a better idea?

IwanKaramazow:
How bad of a grammar conflict are we talking about?
Otherwise indeed, as soon as the parser enters error handling, pattern match on the token, if it's a `STATIC_IDENT`, show a nice error message.

PR #2268, opened 2018-11-19
Prove that no content can be printed after a single line comment.

IwanKaramazow:
Although the comment interleaving should never stick content after
a single line comment, we can't _prove_ it under the current algorithm.
Until proven otherwise, there's a chance content formats after a single
line comment resulting in a syntax error. Although it seems unlikely given the current code.

To completely erradicate the chance, the atom from easy_format is enhanced with
an `eol` parameter. When true, it formats the atom as if it was *end of
that line* and nothing will be formatted after it.


formatTest/unit_tests/expected_output/lineComments.re
```diff
@@ -150,7 +150,8 @@ type t3 = (int, int); // End of line on (int, int)
 
 type variant =
   | X(int, int) // End of line on X
-  | Y(int, int); // End of line on Y // Comment on entire type def for variant
+  | Y(int, int); // End of line on Y
```
jordwalke: I'd say this one was an improvement.

jordwalke: I also believe that this improvement lends itself better to idempotency. (No proof of that but just combining multiple things into one thing  as it did before seems to pose challenges).

src/reason-parser/reason_syntax_util.ml
```diff
@@ -153,6 +153,17 @@ let escape_string str =
...
 (* #if defined BS_NO_COMPILER_PATCH then *)
 
+(*
+    UTF-8 characters are encoded like this (most editors are UTF-8)
+    0xxxxxxx (length 1)
+    110xxxxx 10xxxxxx (length 2)
+    1110xxxx 10xxxxxx 10xxxxxx (length 3)
+    11110xxx 10xxxxxx 10xxxxxx 10xxxxxx (length 4)
+   Numbers over 127 cannot be encoded in UTF in a single byte, so they use two
+  bytes. That means we can use any characters between 128-255 to encode special
```
jordwalke: Great explanation, and thank you for the note. Every time I use/understand one of these printing tricks I had to dig up a Discord conversation.

jordwalke:
I wonder if this means we can remove the first approach that this was built on top of now. I wonder what, if anything would change.

PR #2269, opened 2018-11-19
Upgrade vendored easy-format before changing / forking it

anmonteiro:
should be applied before https://github.com/facebook/reason/pull/2268


anmonteiro:
@jordwalke I don't think we did, the diff here is pretty simple actually. Also noticed the same in #2222

PR #2275, opened 2018-11-29
Fix #1518: Handle CRLF with refmt

### LINKED ISSUE #1518: Refmt: CRLF is treated differently than LF
When typing:

```
let x = 1;
/* foo */
let y = 2;
```

With CRLF (windows) line endings you get:

```
let x = 1; /* foo */

let y = 2;
```

With LF (unix) line endings you get:

```
let x = 1;

/* foo */
let y = 2;
```

bryphe:
__Issue:__ The `Format` module only outputs LF (`\n`) newlines - when the channel is opened in binary mode on Windows, we'll get non-windows line endings.

__Fix:__ Conditionally open in binary/non-binary mode depending on the desired output. There are cases where we actually need to output binary data (via `output_value`, so for now, only set binary_mode = `false` when we are outputting `Reason` or `ML` print types).


jordwalke:
One question: On windows it's possible to configure your editor for either mode. I think one bug we saw was that we weren't preserving the line endings mode. What do you recommend? Should it be that if the file had even just one `CRLF` then we could turn all of the line endings into `CRLF`? I don't think we need to support "mixed" line ending files, but I think we need to preserve the file's current line endings assuming they are all one or the other.

dra27:
As a rule of thumb, if you are writing a new file, then use CRLF on Windows (i.e. for new files, obey the platform norms) but, yes, if you have read a file then you need to track whether the original was LF or CRLF. Even the new Windows LF-reading Notepad gets that right, just to provide a baseline 🙂

bryphe:
Good catch @jordwalke @dra27 ! Yes, we should be preserving newlines. I'll update the PR to account for this.

jordwalke:
For refmt, we could probably use a convention that the first (CRLF or LF) that we witness determines the kind of newline we output for the whole file. Then as long as files aren't mixed the output newlines are retained. If the file was mixed then the first "newline like" character we saw is as good of a heuristic as any.

(Longer term, what convention do streaming transform like programs use? How do you output a newline when you haven't seen any newlines in the input yet? You could default to CRLF, but then what if you later discover that the first newline like character is actually a LF?). Doesn't seem like there's any solution there. Luckily for refmt the situation is easier.

bryphe:
> For refmt, we could probably use a convention that the first (CRLF or LF) that we witness determines the kind of newline we output for the whole file.

Sounds good. I spent of time today looking at this strategy. It looks like our Lexer expects 'normalized' newlines here: https://github.com/facebook/reason/blob/a59c9fd6985f64e7b3073c36ba052ab9ada9f531/src/reason-parser/reason_toolchain.ml#L102

I think it would be risky to change this logic to `open_in_bin` (which we need to do to detect newlines) because it looks like some code depends on the newline normalization behavior here.

I was considering 'peeking' at the file prior to lexing - examine the first line and determine the EOL from that, then close it prior to lexing / parsing. Open to other suggestions, though!

> Longer term, what convention do streaming transform like programs use? How do you output a newline when you haven't seen any newlines in the input yet? You could default to CRLF, but then what if you later discover that the first newline like character is actually a LF?

Yeah, we actually hit this with `use_stdin` here - my plan was just to default to the platform-specific EOL in that case. I don't think there is a great solution, other than pushing the problem up a layer (making the EOL behavior configurable)

bryphe:
Alright, implemented the logic we discussed - if it's `use_stdin` + `ML`/`RE`, we'll use the default platform EOL. Otherwise, we'll check the first line in to detect the newline - and use that to decide if we should use text mode / binary mode.

As noted in the code - there is one current gap in the implementation. If you run `refmt` on a `CRLF` file on OSX/Linux, we'll output the file in text mode, but the line endings wouldn't be converted in that case (although, there'd likely be more problems - they also wouldn't be converted on input, which might bust parsing). @jordwalke and I discussed this and felt it was reasonable (this is the least likely case to run into).

bryphe:
Also - I had some basic test cases here: https://github.com/facebook/reason/pull/2275/commits/bd1dac71003498c6f642e4462254f068e71d3be0#diff-1ac7c12e03995e9441471c0e90889302L38 - would be a great place to leverage https://github.com/facebookexperimental/reason-native/tree/master/src/test-runner 😄

jordwalke:
Yeah, the Reason tests are probably a great use case for snapshot testing.

src/refmt/eol_detect.ml
```diff
@@ -0,0 +1,19 @@
+type eol =
+  | LF 
+  | CRLF 
+let show eol = match eol with | LF -> "lf" | CRLF -> "crlf"
+let default_eol = match Sys.win32 with | true -> CRLF | _ -> LF
+let crlf_regex = Str.regexp ".*\r\n$"
```
jordwalke: Oh, I wonder if we need to be able to do this without `Str` because `Str` cannot be compiled to JavaScript. For example sketch.sh includes the Reason parser as a library then compiles Reason itself into JavaScript. Pretty sure it wouldn't end up pulling in this module but just in case I think we can drop the `Str` dep.

I think after we see the first `\n` when scanning char by char, we can just then check if the char immediately before it was `\r`.

src/refmt/eol_detect.ml
```diff
@@ -0,0 +1,19 @@
...
+let get_eol_from_string s =
+  try ignore (Str.search_forward crlf_regex s 0); CRLF with | Not_found -> LF
```
jordwalke: Btw, in this version, instead of `CRLF` shouldn't it be the default char for the platform?

dra27:
> As noted in the code - there is one current gap in the implementation. If you run `refmt` on a `CRLF` file on OSX/Linux, we'll output the file in text mode, but the line endings wouldn't be converted in that case (although, there'd likely be more problems - they also wouldn't be converted on input, which might bust parsing). @jordwalke and I discussed this and felt it was reasonable (this is the least likely case to run into). We could always add a post-format conversion step if this is necessary.

This case - reading a CRLF file on Unix - is (as I noted in my OCaml workshop talk) the most compelling reason for _never_ using text mode for cross-platform applications, IMO.

bryphe:
> This case - reading a CRLF file on Unix - is (as I noted in my OCaml workshop talk) the most compelling reason for never using text mode for cross-platform applications, IMO.

@dra27 - cool! I'm 100% on board with that message 😄 

Was your talk recorded? I would be very interested.

hcarty:
@bryphe The video is up here, along with a few others from the session: https://youtu.be/1DAuSSljLFI

bryphe:
@hcarty - thank you! 

Excellent talk @dra27 - really interesting background on some of these issues, learned a lot. 

Great summary slide at the end too - worry about windows at the beginning, use abstractions that cover all OSes, and continuously test across platforms, Couldn't agree more!

src/refmt/eol_detect.ml
bryphe: Thanks @jordwalke - good catch. I'll implement this w/o using regexp 👍

src/refmt/eol_detect.ml
bryphe: The way I designed the solution here was that this `eol_detect` module reports the current state of the world (ie, whether the string/file has a `CRLF` or not) - no translation is happening at this point. Perhaps it could be named better to reflect th is?

IMO, it'd be confusing if `get_eol_from_string "test\r\n"` returned `LF` on Linux/OSX. 

We then decide how to act on that information elsewhere (like, whether we should use `open_out` or `open_out_bin`). Let me know if you have ideas to improve this!

src/refmt/eol_detect.ml
bryphe: Much simpler now (I think) w/ your sugestion and correctly handles the missed case in the other comment - if there is no newline in the file... we default to platform newline.

bryphe:
Hmm, seeing this error on CI:
```
#=== ERROR while compiling conf-pkg-config.1.1 ================================#
"pkg-config": command not found.
```

I don't believe this is related to my change (but I have seen it twice in a row...)

bryphe:
I think this is all set now - I added a `CRLF` test and we now handle `CRLF` on all platforms - so line endings should be preserved in any case. Please let me know if you have any additional feedback!

It'd be nice to hook up some additional tests to the internals via `rely` at some point, too... the `Eol_detect` and `Eol_convert` modules would be amenable to that kind of coverage.

kpsuperplane:
I'm not sure if it's entirely safe to assume if the first newline doesn't have `\r` then none of them do; I think inconsistent line endings is an edge case that happens often enough that we can't ignore it. 

With that said, searching the entire file for `\r` is understandably a pretty expensive operation so I'm a little conflicted on this. Would make your `detect_eol` a lot simpler with `String.contains s '\r'` though 😄

bryphe:
> I'm not sure if it's entirely safe to assume if the first newline doesn't have \r then none of them do; I think inconsistent line endings is an edge case that happens often enough that we can't ignore it.

Ah ya, good catch @kpsuperplane ! @jordwalke and I discussed this a bit -  I believe our conclusion was that it was an OK with this heuristic for now (although, as you mention, it's not perfect - we won't handle files with mixed-line endings). 

The mixed-line-ending case becomes a bit more complicated - do you try and preserve the newlines for particular lines? What happens when a line gets split or merged as a result of formatting? Ends up complicating things a bit. 

An alternative, and potentially simpler approach, is for mixed-case files, to just use the platform default and emit a warning. But detecting that case becomes more expensive.

jordwalke:
Great! We should publish some of these utilities for dealing with cross plat windows nuances so people don't have to think about them as much.

PR #2277, opened 2018-11-30
Drop parens around Pexp_fun in argument position

IwanKaramazow:
Before
```reason
  switch (action, state) {
  | Click =>
    ReasonReact.UpdateWithSideEffects(
      {...state, click: click + 1},
      (
        self => {
          let _ = 1;
          apply(bar);
        }
      )
    )
```
After
```reason
  switch (action, state) {
  | Click =>
    ReasonReact.UpdateWithSideEffects(
      {...state, click: click + 1},
      self => {
        let _ = 1;
        apply(bar);
      }
    )
```

### DIFF
--- a/formatTest/unit_tests/expected_output/syntax.re
+++ b/formatTest/unit_tests/expected_output/syntax.re
@@ -1284,10 +1284,10 @@ Thing.map(
   foo,
   bar,
   baz,
-  [@attr]
-  ((abc, z) => MyModuleBlah.toList(argument)),
-  [@attr]
-  ((abc, z) => MyModuleBlah.toList(argument)),
+  [@attr] (abc, z) =>
+    MyModuleBlah.toList(argument),
+  [@attr] (abc, z) =>
+    MyModuleBlah.toList(argument),
 );
 
 Js.Option.andThen((. w) => w#getThing());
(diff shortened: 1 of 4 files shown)


formatTest/unit_tests/expected_output/syntax.re
```diff
@@ -1284,10 +1284,10 @@ Thing.map(
   foo,
   bar,
   baz,
-  [@attr]
-  ((abc, z) => MyModuleBlah.toList(argument)),
-  [@attr]
-  ((abc, z) => MyModuleBlah.toList(argument)),
+  [@attr] (abc, z) =>
```
jordwalke: This formatting is a little weird. We might consider improving it next time we're in this part of the formatter.

src/reason-parser/reason_pprint_ast.ml
```diff
@@ -3600,13 +3600,13 @@ let printer = object(self:'self)
    * TODO: in the future we should probably use the type ruleCategory
    * to 'automatically' ensure the validity of a constraint expr with parens...
    *)
-  method unparseConstraintExpr ?(ensureExpr=false) e =
+  method unparseProtectedExpr ?(forceParens=false) e =
```
jordwalke: The name here "constraint expr" referred to type constraint IIRC. What does "protected" mean?

IwanKaramazow: I'll submit a PR with some comments.

formatTest/unit_tests/expected_output/syntax.re
IwanKaramazow: How would you like to see this?

PR #2283, opened 2018-12-05
Make CI green

bryphe:
Seeing errors on master of the form:
```
#=== ERROR while compiling conf-pkg-config.1.1 ================================#
"pkg-config": command not found.
```

(Also impacting PRs - specifically for the 4.02 builds).

This was addressed by installing the `pkg-config` and `ncurses` packages. Not sure why this is needed now - perhaps the docker images were updated?

I also took out a build step in azure-pipelines that isn't applicable at the moment - it was trying to `npm install` and `npm build` but that doesn't make sense for this project.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,6 +32,7 @@ common_steps: &common_steps
     - run:
         name: "Install deps"
         command: |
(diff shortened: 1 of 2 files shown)


jordwalke:
Thank you!

PR #2288, opened 2018-12-11
Print final closing brace when printing function application with paren hugging

IwanKaramazow:
Ensures idempotency of `<button className={Cn.make(["button", "is-fullwidth"])}>`


jordwalke:
Thank you! Looking good.

PR #2290, opened 2018-12-11
Fix stackoverflow in unparseExprRecurse

anmonteiro:
This is an alternative to #2289 that also renames `literalAttrs` to
`stylisticAttrs` as request in #2223's code review


bobzhang:
cc @chenglou cherry pick it for a release?

Issue #2160, opened 2018-08-26
[Printing] Ternary are not very readable


danielo515:
Hello,

Sadly refmt is not configurable, I can live with most of the already taken decisions, but I really think the default (and only!) configuration should favor readability.
There is a particular case with ternaries that makes them almost invisible on big portions of JSX.
For example, compare this where the ternary sections are clearly visible:
![captura de pantalla 2018-08-26 a las 11 02 48](https://user-images.githubusercontent.com/2270425/44626862-8334ea00-a924-11e8-96b8-cd94958666ce.png)

With this one, where there is almost impossible to spot where the ternary starts and where does it ends:

![captura de pantalla 2018-08-26 a las 11 02 14](https://user-images.githubusercontent.com/2270425/44626879-c000e100-a924-11e8-807a-3c862c98368a.png)

I asked about this on discord and @IwanKaramazow asked me to ping him on this issue, so here it is

davesnx:
This should be fixed after #2294

Issue #1987, opened 2018-06-05
Ternary formatting


bordoley:
Assuming #1700 isn't implemented (I'm not against that change btw) can the formatting of ternaries be improved. For the common if/else formatting, it would preferable to follow facebook standard which is:

```
booleanCheck
  ? trueResult
  : falseResult;
```
which is far more readable vs. the current formatting:
```
booleanCheck ?
  trueResult : falseResult;
```
  
For the nested ternaries, I'd love to see a JS style pattern match layout used, though I'm sure this is more contentious. For instance:
```
booleanCheck1 
  ? trueResult1 :
booleanCheck2
  ? trueResult2 :
booleanCheck3
  ? trueResult3 :
fallbackResult;
```

chenglou:
First one should be done. Not sure we should encourage the last one though. That should be pattern matching on a tuple

bordoley:
Regarding the second pattern, I don't find myself using it in my reason/ocaml code, as pattern matching is so much more powerful. That said my major annoyance with prettier in the javascript world is it's formatting of ternaries. Perhaps non-nested formatting of these would be a decent incremental step when introducing people to reason. Or even better, maybe convert nested ternaries into switch statements.

rickyvetter:
Pattern match on a tuple isn't lazy so it's not quite the same. I definitely think the simple case formatting is more important here.

jordwalke:
Flattening out the formatting seems nice. I don't have strong opinions on the formatting style of `?` placement (end of line or beginning). I suppose beginning of line would make it consistent with other operators.

PR #2294, opened 2018-12-13
Update ternary formatting

bandersongit:
Addresses https://github.com/facebook/reason/issues/2160 and https://github.com/facebook/reason/issues/1987

line breaks for ternaries are now formatted as
```
valueICareAbout > Threshold
  ? value
  : otherValuereallyreallyreallyreallyreallyreallyreallyreallyreallyreallyreallyreallyreallylong;
```
or (the slightly less desirable, but better than the current situation)
```
valueICareAbout > Threshold
  ? value : otherValueKindaLongButNotEnoughForBothfohsd;
```
@jordwalke and I spent some time trying to force the above situation in all overflow cases, but had considerable difficulty and we feel that this is still considerably better than the status quo.

nested ternaries are formatted as below (this is different from prettier, but I think considerably more clear): 
```
valueICareAbout > Threshold
  ? value
  : otherValue > thing
      ? thing1
      : thing2reallyreallyreallyreallyreallyreallyreallyreallyreallylong;
```

The snapshot changes are worth looking at to understand how some edge cases are handled (I think it behaves better than prettier).

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/imperative.re
+++ b/formatTest/typeCheckedTests/expected_output/imperative.re
@@ -61,8 +61,8 @@ let result =
       ();
     }
   )
-  == () ?
-    false : true;
+  == ()
+    ? false : true;
 
 switch (
   try (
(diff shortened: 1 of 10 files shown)


IwanKaramazow:
@bandersongit I'll review this weekend.

IwanKaramazow:
This is a very nice improvement. Code looks good at first sight.
@bandersongit Can you maybe add an extra test case(s) for the less desirable output with a comment for future reference?

bandersongit:
> This is a very nice improvement. Code looks good at first sight.
> @bandersongit Can you maybe add an extra test case(s) for the less desirable output with a comment for future reference?

@IwanKaramazow done!

Issue #2297, opened 2018-12-23
Invalid output with JSX: missing closing bracket


jchavarri:
Valid input: `let t = <view style={Style.make()}> ...items </view>`

Invalid output: `let t = <view style={Style.make()> ...items </view>;` 

Note the missing closing bracket after `Style.make()`.

Tested in `v 3.3.7,  @ 43efc14`.
Tested as well with `reason-language-server` v 1.4.1, it produces the same invalid output.

IwanKaramazow:
This has been fixed in https://github.com/facebook/reason/pull/2292/.

jchavarri:
Thanks @IwanKaramazow !! 🙂

PR #2298, opened 2018-12-23
Fix missing closing bracket in jsx with Pexp_apply unit

IwanKaramazow:
Addresses https://github.com/facebook/reason/issues/2297


IwanKaramazow:
This is already fixed in master.

PR #2308, opened 2018-12-31
rename fast pipe to pipe first

nikgraf:
Updating the reason core to match the documentation update: https://github.com/reasonml/reasonml.github.io/pull/485


jordwalke:
Thank you!

Issue #2301, opened 2018-12-25
Getting Stack overflow with 3.3.7 and master 9914cc69


andreypopp:
Getting Stack overflow with 3.3.7 and master 9914cc69:
```
% OCAMLRUNPARAM=b esy refmt --parse ml ./esy/Scope.ml
refmt: internal error, uncaught exception:
       Stack overflow
       Raised at file "parsing/location.ml", line 465, characters 14-25
       Re-raised at file "parsing/location.ml", line 465, characters 14-25
       Re-raised at file "parsing/location.ml", line 465, characters 14-25
       Re-raised at file "parsing/location.ml", line 465, characters 14-25
       Re-raised at file "parsing/location.ml", line 465, characters 14-25
       Re-raised at file "parsing/location.ml", line 465, characters 14-25
       Called from file "parsing/location.ml" (inlined), line 470, characters 31-61
       Called from file "src/refmt/refmt_impl.ml", line 96, characters 10-60
       Called from file "src/reason-parser/vendor/cmdliner/vendored_cmdliner.ml", line 1303, characters 27-34
       Called from file "src/reason-parser/vendor/cmdliner/vendored_cmdliner.ml", line 1350, characters 17-26
       Called from file "src/reason-parser/vendor/cmdliner/vendored_cmdliner.ml", line 1375, characters 8-50
```

Here's the [`Scope.ml`](https://gist.github.com/e40796ec62cced7494f477514d24f2de).

PR #2314, opened 2019-01-08
Parse and print `Ppat_open`

anmonteiro:
fixes https://github.com/facebook/reason/issues/2301


formatTest/typeCheckedTests/expected_output/mlSyntax.4.04.0.re
```diff
@@ -0,0 +1,5 @@
+module Foo = {
+  type t = {name: string};
+};
+
+let foo = (Foo.{name}) => ();
```
jordwalke: Want to add some more tests covering the various places patterns can occur? Destructuring tuples, switch statements etc.

anmonteiro:
@jordwalke I added some more parsing rules to support parsing e.g. `let Foo.[bar] = baz;` and other patterns.

There's only one caveat with the current state of this PR: it parses `let Foo.((bar, baz)) = qux;` but not `let Foo.(bar, baz) = qux;`. I'm considering it future work to relax the double parens requirement.

jordwalke:
Cool, those seem like nice additions too. Is this ready to merge then?

jordwalke:
@kpsuperplane: The Azure builds are getting an error very similar to one you were getting on reason-native:

```
Definition name facebook.reason didn't correspond to a valid definition
```

Do you recall what the culprit is? How do I fix it?

kpsuperplane:
@jordwalke  I solved it last time by renaming the project to something else (e.g. "facebook-reason"), running a successful build, and renaming it back. Never really figured out the root cause though

PR #2316, opened 2019-01-09
Don't pun record types if they contain attributes

### LINKED ISSUE #2315: refmt applies punning too agressively dropping extension points
Before refmt:
```
type t = {
  name : [@default x] name,
};
```
after:
```
type t = {name};
```

Tested with 3.3.7 and master.

anmonteiro:
Fixes #2315


jordwalke:
Thanks!

PR #2325, opened 2019-01-17
Fix HISTORY link formatting

jaredly:


jaredly:
wut

IwanKaramazow:
You probably pushed to the fb remote, happened to me too 😄

jaredly:
I did an in-browser edit. I guess that's not supported?

PR #2330, opened 2019-01-29
Preserve semicolon printing in seq-expr inside jsx-children

### LINKED ISSUE #2293: [refmt] Block formatting in JSX
Input:

```reason
<div>
  {
    let left = (limit - value->Js.String.length)->Int.toString;
    {j|$left characters left|j}->React.string;
  }
</div>
```

Output:

```reason
<div>
  {let left = (limit - value->Js.String.length)->Int.toString
   {j|$left characters left|j}->React.string}
</div>
```

IwanKaramazow:
Fixes #2293


IwanKaramazow:
@bloodyowl do you need this in bs too?

bloodyowl:
@IwanKaramazow yes

PR #2333, opened 2019-02-04
Sync outcome printer changes back from BS

chenglou:


jordwalke:
The errors "Definition name facebook.reason didn't correspond to a valid definition" are weird Azure ones that come up occasionally. I would ignore them for now.

chenglou:
kk

PR #2335, opened 2019-02-10
Fix operator swap for external identifiers

anmonteiro:


anmonteiro:
Rebased on master

jordwalke:
Refreshing my memory about the behavior before this diff:

- Parsing the identifier `pri` isn't possible in Reason (it's a token for objects)
- Parsing the identifier `pri_` in Reason becomes the identifier `pri` in the AST.
- Parsing the identifier `pri` in ML is possible, and of course remains `pri` in the AST. When printed back out in Reason it is rendered as `pri_`.

- Parsing the identifier `private` isn't possible in ML.
- Parsing the identifier `private_` is possible in ML, and remains `private_` in the AST
- Parsing the identifier `private` in Reason becomes the identifier `private_` in the AST tree. When printed to ML it is shown as `private_` of course. However, there this wasn't getting printed out correctly for `external` identifiers.

jordwalke:
I think I found an additional bug that this diff would fix. Can you add a test case for an OCaml syntax input of:

```
external pri : int -> int = ""
```

That should parse to an AST with `pri` and it should print to Reason as `pri_`, but it doesn't currently work.

PR #2336, opened 2019-02-12
Fix inline record printing in outcome printer

### LINKED ISSUE #1163: Inline records in variant type definitions aren't pretty-printed properly in rtop

anmonteiro:
This fixes #1163 and supersedes #2103

### DIFF
--- a/formatTest/oprintTests/expected_output/inlineRecord.re
+++ b/formatTest/oprintTests/expected_output/inlineRecord.re
@@ -1,0 +1,6 @@
+type t0 = T0 { t0: int, };
+type t1 = A { x: int, } | B | C { c1: string, c2: string, };
+type t2(_) =
+    D { x: int, }: t2(int)
+  | E { f: int => int, }: t2(int => int)
+  | F(unit): t2(unit);
--- a/formatTest/oprintTests/input/inlineRecord.re
+++ b/formatTest/oprintTests/input/inlineRecord.re
@@ -1,0 +1,11 @@
+type t0 = T0 { t0 : int };
+type t1 =
+  | A { x : int }
+  | B
+  | C { c1 : string, c2 : string };
+
+/* GADT */
+type t2(_) =
+  | D { x : int } : t2(int)
+  | E { f : int => int } : t2(int => int)
+  | F(unit) : t2(unit);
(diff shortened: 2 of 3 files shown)


formatTest/oprintTests/expected_output/inlineRecord.re
```diff
@@ -0,0 +1,6 @@
+type t0 = T0 { t0: int, };
+type t1 = A { x: int, } | B | C { c1: string, c2: string, };
```
IwanKaramazow: Doesn't Reason have parens around constructor args?

anmonteiro: I don't think so. Here's an rtop session:

```rust
Reason # type a = | A {x:int}: a;
type a = A(): a;
Reason # type a = | A {x:int};
type a = A();
Reason # type a = | A({x:int});;
Error: Syntax error
Reason # type a = | A {x:int}: a;
type a = A(): a;
Reason # type a = | A ({x:int}): a;
Error: Syntax error
```

anmonteiro: Maybe it should, though?

hcarty: The lack of support for wrapping parens for inline records, when they're required in other cases, is very confusing.  It's an inconsistency which would be good to fix.

anmonteiro:
@jordwalke rebased on master.

formatTest/oprintTests/expected_output/inlineRecord.re
```diff
@@ -0,0 +1,6 @@
...
+type t2(_) =
+    D { x: int, }: t2(int)
```
jordwalke: Great! You have a GADT test case.

formatTest/oprintTests/expected_output/inlineRecord.re
jordwalke: Yeah, I just saw someone get pretty confused about this too.
One somewhat related question I have is if there might ever be a change in OCaml that requires syntactically distinguishing values of inline variant records vs. values of variants that happen to contain records as their payload. I doubt it. That tells me that we should probably move to consistent parens everywhere.
If we hypothetically one day figured out a way to be able to drop parens for single argument things, then we would apply it consistently everywhere, not just on variants. Good discussion, thanks.

PR #2342, opened 2019-03-08
Fix operator swap for type declarations

### LINKED ISSUE #2341: Reason reserved words are not translated in type names coming from OCaml
If you have an OCaml module which uses a Reason reserved word as a type name there is no way to access that type name from within Reason code.  For example, the `pub` type here couldn't be referenced from Reason:
```ocaml
type pub = { x : int }
```
For values, this is handled by adding a trailing `_` to the value's name.  We should have the same transformation for types.  And probably record fields, method names and anything else which could conflict if they're not translated yet.

anmonteiro:
Fixes #2341

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/mlSyntax.re
+++ b/formatTest/typeCheckedTests/expected_output/mlSyntax.re
@@ -110,3 +110,6 @@ let pri__ = "private";
 
 external private: unit => unit = "";
 external pri_: unit => unit = "";
+
+type pub_ = int;
+type pub__ = int;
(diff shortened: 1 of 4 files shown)


jordwalke:
Is this a breaking change in any way that you can think of? (For example if someone relied on accessing a `pub_` identifier in from an ocaml module).

jordwalke:
I think this needs a little rebase. Thanks!

anmonteiro:
I _think_ it's not a breaking change. If somebody used `pub_` in a Reason file I think it wouldn't work in the OCaml AST? I'm unsure.

Rebased!

jordwalke:
Thank you!

jordwalke:
I think this came up with values being named `pub`, and people using them from an ML authored library. When we fixed it, something did break in one library but we worked with that author to fix it by having them export the same value with `let pub = ` and `let pub_ = `. If it comes up with types, we can take a similar approach because this diff fixes a bug that people could have been relying on.

PR #2345, opened 2019-03-10
Preemptive changes for OMP that supports OCaml 4.08

anmonteiro:
At the Mirage retreat this week I was talking to @armael regarding OCaml Migrate Parsetree support for OCaml 4.08 (related issue/comment: https://github.com/ocaml-ppx/ocaml-migrate-parsetree/issues/50#issuecomment-467164936).

Apparently, the error type has changed in 4.08 and calling into `Location.register_error_of_exn` is not going to work with the next version of OMP that snapshots the `Location` module.

Therefore we worked together on preemptively fixing this such that Reason can support OCaml 4.08 once OMP does.

Test plan: Current tests include error tests that are still passing.

### DIFF
--- a/src/refmt/refmt_impl.ml
+++ b/src/refmt/refmt_impl.ml
@@ -92,9 +92,12 @@ let refmt
     | _ -> `Ok (List.iter (fun file -> refmt_single (Some file)) input_files)
   with
   | Printer_maker.Invalid_config msg -> `Error (true, msg)
-  | exn ->
-          Location.report_exception Format.err_formatter exn;
-          exit 1
+  | Reason_syntax_util.Error (loc, error) ->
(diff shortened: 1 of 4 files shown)


src/refmt/refmt_impl.ml
```diff
@@ -92,9 +92,12 @@ let refmt
     | _ -> `Ok (List.iter (fun file -> refmt_single (Some file)) input_files)
   with
   | Printer_maker.Invalid_config msg -> `Error (true, msg)
```
anmonteiro: @jordwalke re: your question on Discord: I don't think we throw any other exceptions here, and therefore it is OK to only catch these.

Feel free to prove me wrong and I'll update the code accordingly.

anmonteiro:
@jordwalke this is rebased on master

PR #2348, opened 2019-03-14
Pass certain files through cppo

chenglou:
This file is shared between Reason and BuckleScript. We conditionally compile to the right stuff depending on the cppo settings. Doesn't change anything for
Reason itself


chenglou:
CI failed because of 2 things that are failing in other PRs too

PR #2351, opened 2019-03-19
Fix Azure Pipelines

ulrikstrid:


ulrikstrid:
cc @jordwalke is there anything else you want me to add here?

Issue #2349, opened 2019-03-18
refmt removes semicolons inside ternary blocks in jsx


baldurh:
This bug appears in `bs-platform@5.0.0-dev.5`

```reason
{true
	? {let foo = "foo"; // refmt removes the semicolon here
		<span> {ReasonReact.string(foo)} </span>}
	: <span> {ReasonReact.string("bar")} </span>}
```

Reproduced in https://github.com/baldurh/repro-refmt-issue-bs-platform-5

cc @bobzhang

IwanKaramazow:
I'm fixing this.

PR #2352, opened 2019-03-19
Don't remove semis in blocks inside ternary expressions as jsx children

IwanKaramazow:
Fix https://github.com/facebook/reason/issues/2349


chenglou:
Cc @bobzhang

bobzhang:
@chenglou Looks great, can we have this fix upstreamed ? thanks

jordwalke:
Thank you @IwanKaramazow !

Issue #2350, opened 2019-03-18
[refmt] Lexer fails on line comment, if it's the last line without a newline.


cristianoc:
e.g. if this line is the entire file
```reason
// 34
```

```
refmt: internal error, uncaught exception:
       File "src/reason-parser/reason_lexer.mll", line 725, characters 6-12: Assertion failed
```

IwanKaramazow:
What version/commit are you running?
I can't seem to reproduce this on master…

IwanKaramazow:
Seems that the code doesn't explicitly handle EOF, submitted a PR that does exactly that in https://github.com/facebook/reason/pull/2353/files

PR #2353, opened 2019-03-19
Handle single line comments ending with eof

IwanKaramazow:
Fixes https://github.com/facebook/reason/issues/2350


chenglou:
@bobzhang

PR #2363, opened 2019-03-30
Parse and print braces around inline records

anmonteiro:
This was brought up in https://github.com/facebook/reason/pull/2336


anmonteiro:
We should also fix this in the outcome printer when #2336 is in. I don't wanna have to deal with merge conflicts :)

jordwalke:
This is great! It's a "minor version bump" change, which means it's a new feature that people might depend on in their code.

For example, for this we would bump the version from `3.4.1` to `3.5.0`. No one's existing code will break (that would be a `4.x.x` bump) but some people can write code that depends on the new feature so it's an `y` bump in `x.y.z` (that's called a "minor version" bump).

If John uses this feature in his code, he will need to specify in their package versions:

```
"@esy-ocaml/reason": ">= 3.5.0 < 4.0.0"
```

If Mary has a package that depends on "end of line comments" feature, she would have specified:

```
"@esy-ocaml/reason": ">= 3.4.0 < 4.0.0"
```

Because `3.4.0 added end of line comments.

And if those two people mixed their packages into an app, the package solver would know to use reason version `3.5.0` with your change. I just thought I'd write out the the steps of how I think through potential scenarios for breakage so other people understand what's behind versioning in Reason.

PR #2365, opened 2019-03-31
Touching readme to trigger Azure - this time from forked repo

jordwalke:
Let's see if the Azure CI still works when PRs come from a forked repo.


jordwalke:
Interesting. Linux esy builds on pull requests, but was failing on the master branch. It might be because they are using different caches.

PR #2367, opened 2019-03-31
Fix Azure for pull requests. Use the REST API to download artifacts.

jordwalke:
It turns out we can't reliably download artifacts using the old method for pull requests. Something about permissions. With the REST API all artifacts can be downloaded publicly. Let's use it.


jordwalke:
Squashing and merging! Now PRs will be green and fast!

PR #2370, opened 2019-04-01
Format inline records with parens in outcome printer

anmonteiro:
This is a followup to the syntax change added in #2363, but for the
outcome printer


jordwalke:
Thank you!

PR #2375, opened 2019-04-11
update ocaml version constraint to allow 4.08, bump package version to 3.4.2, add test script for esy, update README.md and HISTORY.md

bandersongit:
Bumped ocaml version constraints now that https://github.com/facebook/reason/pull/2345 is released. It looks like there were already notes written up for 3.4.2 so I bumped to that version as this seems fine for a patch version.


anmonteiro:
There have been similar changes pushed to master. Closing this one.

PR #2377, opened 2019-04-13
Catch more errors

anmonteiro:
This fixes a regression I introduced in #2345.

Also added a regression test.


IwanKaramazow:
LGTM

PR #2378, opened 2019-04-13
Prettify try to hug braces

anmonteiro:


jordwalke:
Is there a case that tests the primary value of hugging parens for try?

```
try({
  let foo = "asldkfj";
  ...
}) {
| 
};
```

anmonteiro:
I can’t remember if there was one.

That’s definitely a case that I tested manually though.

I’ll check if there’s one in a little bit.

anmonteiro:
@jordwalke Added a test for the let list

PR #2380, opened 2019-04-14
Translate many more OCaml identifiers to Reason

### LINKED ISSUE #2341: Reason reserved words are not translated in type names coming from OCaml

anmonteiro:
This changeset now fixes #2341.

Also fixes https://github.com/facebook/reason/issues/2494

Thanks @hcarty for the detailed examples!


jordwalke:
Looks good! I don't think we should/need-to normalize the polymorphic variant names though.

jordwalke:
Thanks @anmonteiro !

PR #2390, opened 2019-04-20
Proposal: Pad record braces

anmonteiro:
This is a proposal to add one space before and after record / object open and close braces, respectively.

I feel (subjective) that being a bit more sparse in these cases makes the syntax look nicer. Opinions welcome.


jaredly:
I don't love it, but could probably get used to it. have we talked about configuration recently? Could be a trial case for that.

PR #2399, opened 2019-04-30
Use utop as a library for rtop

ulrikstrid:
Use `UTop_main.main ()` to setup the utop toplevel to reduce complexity of implementation


src/rtop/rtop.ml
```diff
@@ -1,32 +1,25 @@
...
-let home = match getenv_opt "HOME" with
-  | Some(p) -> p
-  | None -> Sys.getenv "USERPROFILE"
+let () = try Topdirs.dir_directory (Sys.getenv "OCAML_TOPLEVEL_PATH") with | Not_found -> ();;
 
-let () = Unix.openfile (Filename.concat home ".utoprc") [] 0o640
-         |> Unix.close
+let () = UTop.require ["reason.easy_format"; "reason";]
 
-let () = Unix.openfile (Filename.concat home ".utop-history") [] 0o640
```
jordwalke: I wonder if it now wouldn't preserve history or load settings.

ulrikstrid:
The diff is random, it probably makes more sense to compare `rtop.ml` with `rtop_init.ml` as that is what I reimplemented

src/rtop/rtop.ml
ulrikstrid: It kept history when I tested while building it using `esy x rtop`, my guess is that it just loads `.utop-history` or something like that anyway.

hcarty: If it loads the same history as utop won't that cause issues on systems where both utop and rtop are used?

ulrikstrid: I didn't get any history in rtop after using utop but I might have to test this some more

ulrikstrid: Just tried again, can confirm that they do not share history. But it might be because esy sandboxes my rtop.

jordwalke: We should make sure it doesn't fail on a clean install or at least figure out where it's getting the history from - it's getting it from *somewhere*. This was an original bug we had (the history file didn't exist) and rtop failed on first launch.

ulrikstrid: https://github.com/facebook/reason/blob/master/src/rtop/reason_utop.ml#L54

This is where it comes from

hcarty: That's promising - it should continue to use an rtop-specific history file then.

Issue #2395, opened 2019-04-27
Braces not preserved on callback arguments 


anmonteiro:
```reason
let () = foo(bar, (x) => { baz });
```

PR #2403, opened 2019-05-05
Preserve function body braces when Pexp_fun is an argument to a function

anmonteiro:
fixes https://github.com/facebook/reason/issues/2395


jordwalke:
This will be very helpful when writing Rely tests!

formatTest/unit_tests/expected_output/syntax.re
```diff
@@ -1124,17 +1124,17 @@ test(~desc=?[@attr] "my test", ~f=?[@attr] () => {
   x + y;
 });
 
-describe("App", () =>
-  test("math", () =>
+describe("App", () => {
+  test("math", () => {
```
jordwalke: Looks so good!

Issue #2384, opened 2019-04-16
JSX removes semicolon in edge case


bloodyowl:
Reason 3.4.2 @ 3c6a9ca9

```reason
let v =
  <A>
    <B>
      ...{_ => {
        let renderX = x => {
          let y = x ++ x;
          <div key=y />;
        };
        renderX("foo");
      }}
    </B>
  </A>;
```

gets refmted to

```reason
let v =
  <A>
    <B>
      ...{_ => {
        let renderX = x => {let y = x ++ x
                            <div key=y />};
        renderX("foo");
      }}
    </B>
  </A>;
```

which is invalid. it seems that nesting has something to do with it because the following gets formatted as expected

```reason
let v =
  <B>
    ...{_ => {
      let renderX = x => {
        let y = x ++ x;
        <div key=y />;
      };
      renderX("foo");
    }}
  </B>;
```

MoOx:
Thanks @bloodyowl for debugging my problem 🙏

bloodyowl:
cc @IwanKaramazow as you worked on the original fix

MoOx:
I am trying to upgrade some stuff to reason-react 0.7 with bs-platform 5 and I am facing this again on some simpler example and I cannot do anything as a workaround... (or at least I don't find any lol)

```reason
[@react.component]
let make = (~currentLocation, ()) =>
  ReactCompat.useRecordApi({
    ...component,
    render: _self =>
      <SafeAreaView style=styles##wrapper>
        <View style=styles##container>
          Consts.menuLinks->Belt.Array.map
          (
            item =>
              {let isActive =
                 item.isActive(currentLocation##pathname, item.link);
               <ViewLink
                 key={item.link} href={item.link} style=styles##itemWrapper>
                 <View style=styles##item>
                   {item.icon(
                      ~width=24.,
                      ~height=24.,
                      ~fill=isActive ? colorActive : colorInActive,
                      (),
                    )}
                   <Text
                     style=Style.(
                       concat([
                         styles##itemText,
                         isActive ? styles##itemTextActive : style([]),
                       ])
                     )>
                     item.text->ReasonReact.string
                   </Text>
                 </View>
               </ViewLink>}
          )
          ->ReasonReact.array
        </View>
      </SafeAreaView>,
  });
```

get turned into 

```reason
[@react.component]
let make = (~currentLocation, ()) =>
  ReactCompat.useRecordApi({
    ...component,
    render: _self =>
[...]

PR #2406, opened 2019-05-07
Fix JSX removing semicolons

anmonteiro:
Turns out we can have our cake and eat it too. This fixes e.g. #2384 and a few others, resulting from a bug I introduced in #2223

### DIFF
--- a/formatTest/unit_tests/expected_output/jsx.re
+++ b/formatTest/unit_tests/expected_output/jsx.re
@@ -623,21 +623,17 @@ ReasonReact.(<> {string("Test")} </>);
 </div>;
 
 <div>
-  {
-    let left = limit->Int.toString;
-    {j|$left characters left|j}->React.string;
-  }
+  {let left = limit->Int.toString;
+   {j|$left characters left|j}->React.string}
 </div>;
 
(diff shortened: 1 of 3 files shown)


formatTest/unit_tests/expected_output/jsx.re
```diff
@@ -623,21 +623,17 @@ ReasonReact.(<> {string("Test")} </>);
...
 <View style=styles##backgroundImageWrapper>
-  {
-    let uri = "/images/header-background.png";
-    <Image
-      resizeMode=`contain
-      style=styles##backgroundImage
-      uri
-    />;
-  }
+  {let uri = "/images/header-background.png";
```
jordwalke: We don't have this form of wrapping anywhere else (not intentionally at least). Is there a reason why this isn't breaking into multiple lines?
```
{
  let uri = 
  ...
}
```

anmonteiro: This has been here for a while, e.g. https://github.com/facebook/reason/pull/2208.

We did it initially to save space on JSX patterns like:

```js
<div>
  {
    React.string("foo")
  }
```

jordwalke:
Thank you for the fix!

PR #2410, opened 2019-05-11
Fix the build for newer versions of ocaml-migrate-parsetree

anmonteiro:
This fixes the build on the current master with `ocaml-migrate-parsetree` v1.3.0, which produces some warnings for code relying on e.g. `Ast_404` at the toplevel.


jordwalke:
Thank you very much as always, @anmonteiro!

PR #2426, opened 2019-06-15
OCaml 4.08 support

anmonteiro:
This PR makes Reason build on OCaml 4.08. 

It depends on https://github.com/let-def/merlin-extend/pull/9

The 4.08 CI steps are failing because of the above dependency. Perhaps we should vendor merlin-extend too as it's such a small dep?


src/reason-parser/reason_parser.mly
```diff
@@ -59,6 +59,12 @@ open Ast_helper
 open Ast_mapper
 open Reason_string
 
+module Clflags = struct
+  include Clflags
+
+  let fast = unsafe
```
anmonteiro: This is pretty annoying. It won't build in <4.07 so we might have to introduce yet another `.cppo` compat file

src/reason-parser/dune
```diff
@@ -1,42 +1,58 @@
...
 (ocamllex
  (modules reason_lexer))
 
 ; Use select.exe from janestreet/ppx_ast to choose lexer_warning.ml based on ocaml version
+
 (rule
  (targets lexer_warning.ml)
- (deps ../generate/select.exe lexer_warning.ml-4.07 lexer_warning.ml-4.06 lexer_warning.ml-default)
- (action (with-stdout-to %{targets} (run ../generate/select.exe lexer_warning.ml-4.07 lexer_warning.ml-4.06 lexer_warning.ml-default))))
+ (deps ../generate/select.exe lexer_warning.ml-4.08 lexer_warning.ml-4.07
```
jordwalke: Just for your information, `select.ml` is where we currently make similar compiler version specific decisions. That is an alternative to cppo. (It might not be as easy to do so for these particular special cases in this diff).

jordwalke:
Woohoo! Thanks as always.

PR #2433, opened 2019-07-21
Remove reason_toploop

rgrinberg:
It's redundant since reason_utop is already present and doing the same kind of initialization.


anmonteiro:
this seems necessary. removing toploop makes the rtop integration tests break.

PR #2439, opened 2019-08-09
Reerror: error-resilient parsing (same as Merlin for OCaml)

let-def:
This PR implements a new "parsing loop" that is more robust to errors and should improve the experience in an IDE.

The parser itself is not affected (we still have a declarative grammar), but we now have a slightly different interpretation in interactive use-cases.


let-def:
Esy is missing the "fix" dependency. I don't know the proper way to update it.

ulrikstrid:
I can add `@opam/fix` via a resolution in a package.json. I wonder if this is a bug with parsing the .opam file in `esy`.

cc @andreypopp

jordwalke:
Been testing this out and it's so good.
@ulrikstrid - it wouldn't be a resolution, it would just be a regular dependency.

@let-def you just add `"@opam/fix": "*"` to the "dependencies" section of `src/refmt/esy.json` and `./esy.json` (the later is just the development-time package configuration, the later is what will be released).

If you just add those fields, the CI should pass and we can release this puppy.

src/reason-parser/TODO
```diff
@@ -0,0 +1,18 @@
+Recovery (Urgent):
```
jordwalke: Did you want to delete this file and the Plan document?

src/menhir-recover/Makefile
```diff
@@ -0,0 +1,15 @@
+.PHONY: all clean
```
jordwalke: Was this (menhir-recover) copied form another project somewhere?

src/reason-parser/TODO
let-def: Yes

src/menhir-recover/Makefile
let-def: Yes, it is a slightly improved version of https://github.com/ocaml/merlin/tree/master/src/ocaml/preprocess/recover (the same algorithm I wrote for Merlin/OCaml parser).

jordwalke: Would you upstream your changes into merlin?

let-def: I still have some work to do to generalize the algorithm a bit, but I plan to upstream it either into a separate repository or into Menhir directly.

jordwalke:
Woohoo! Closing now that it's merged (after @anmonteiro rebased it). Very excited to announce this one.

PR #2450, opened 2019-09-21
Support OCaml 4.09

anmonteiro:


anmonteiro:
The Windows failure just looks like some NPM idiosyncrasy regarding permissions on Windows.

src/reason-parser-tests/testOprint.cppo.ml
```diff
@@ -31,7 +31,11 @@ let main () =
   let lexbuf = Reason_toolchain.setup_lexbuf true filename in
   let impl = Reason_toolchain.RE.implementation in
 
+#if OCAML_VERSION >= (4,9,0)
+  Compmisc.init_path ();
+#else
   Compmisc.init_path false;
```
jordwalke: I wish in changes like this, the compiler would use labeled arguments with defaults to add new flags to avoid breaks.

jordwalke:
All green on windows!

PR #2454, opened 2019-09-28
WIP - Use rely instead of shell scrips for tests

sync:
Few advantages:

- Rely is super fast
- Rely will generate snapshots automatically (something we had to manually do i guess) and diff snapshots for us with meaningful error messages.
- Much easier and less error prone to add new tests

https://reason-native.com/docs/rely/

I used git to move previously written snapshot but git for some doesn't want to treat them as a move.

### TODO formatTests
- [x] customMLFiles
- [x] errorTests
- [x] idempotentTests
- [x] oprintTests
- [x] typeCheckedTests
- [x] unit_tests

Won't convert the miscTests, it's using ocamlc and there aren't that many files in there I think...

```
Running 1 test suite
 PASS  formatTest

Test Suites: 0 failed, 1 passed, 1 total
Tests:       0 failed, 76 passed, 76 total
Time:        1.343s
```


sync:
@jordwalke before i do any more work does this look ok ?

esy.json
```diff
@@ -4,18 +4,20 @@
...
+    "@opam/result": "*",
+    "@opam/utop": " >= 1.17.0 < 2.3.0",
+    "@opam/shell": "*",
+    "@reason-native/rely": "^3.1.0",
+    "ocaml": " >= 4.2.0 < 4.9.0"
   },
   "devDependencies": {
     "@opam/merlin": "*",
-    "ocaml": "~4.6.0"
+    "ocaml": "~4.7.0"
```
sync: i wanted to use shell for our testing (and require ocaml > 4.7.0)

esy.json
```diff
@@ -4,18 +4,20 @@
   "license": "MIT",
   "version": "3.5.0",
   "dependencies": {
```
sync: i used esy add and it shuffled things around (i have version of esy running off master)

esy.json
```diff
@@ -4,18 +4,20 @@
...
     "@opam/menhir": " >= 20170418.0.0",
-    "@opam/utop": " >= 1.17.0 < 2.3.0",
     "@opam/merlin-extend": " >= 0.4",
-    "@opam/result": "*",
     "@opam/ocaml-migrate-parsetree": "*",
-    "@opam/dune": "*"
+    "@opam/ocamlfind": "*",
+    "@opam/result": "*",
+    "@opam/utop": " >= 1.17.0 < 2.3.0",
+    "@opam/shell": "*",
```
sync: https://ocaml.janestreet.com/ocaml-core/latest/doc/shell/Shell/index.html

with shell you can do this:
```
let patch = run_full(~expect=[0, 1], "diff", ["-u", "--", file1, file2]);
```

sync: just as an example of what is possible with it, we actually don't do this manually but use rely

esy.json
```diff
@@ -26,10 +28,55 @@
...
+        "reason/library"
+      ],
+      "bin": {
+        "ReasonApp.exe": "ReasonApp.re"
+      }
+    }
+  },
+  "scripts": {
+    "test": "esy x RunReasonTests.exe"
   }
```
sync: followed previous pesy test structure

sync:
OK looking at the test failing it seems that shell with a >4.7.0 requirement is a bad idea. Will look at doing something else.

I know there is something i should be able to reuse here.
https://github.com/facebookexperimental/reason-native/blob/master/tests/__tests__/refmterr/refmterr_test.re

I'll also go look at what fnm does for their integration testing

esy.json
sync: no longer using shell

esy.json
jordwalke: Yeah, that's fine.

esy.json
jordwalke: Great! The convention we suggest is to do exactly this - make an `esy x` test script, so basically every project can be cloned, then you run `esy` then `esy test`

test/_snapshots/formatTest_unittests.fff7ed63.0.snapshot
```diff
@@ -174,37 +175,37 @@ let addOne = x => x + 1;
...
+  | false => \"bye\"
   };
 
 /*
  * Many levels of if elseif should be formatted very nicely.
  */
 let result =
   if (something) {
-    Console.log("First Branch");
+    Console.log(\"First Branch\");
```
jordwalke: It's weird that quotes are escaped, but I guess that's just how snapshots work.

jordwalke: @bandersongit Is this intentional? It makes it harder to read snapshots to see what has actually changed.

testExe/dune
```diff
@@ -0,0 +1,5 @@
+(executable
+ (name RunReasonTests)
+ (public_name RunReasonTests.exe)
+ (package reason)
```
jordwalke: Along with my other comment, you would make it be part of the `reason-test` "package".

esy.json
```diff
@@ -4,14 +4,15 @@
...
     "@opam/menhir": " >= 20170418.0.0",
-    "@opam/utop": " >= 1.17.0 < 2.3.0",
     "@opam/merlin-extend": " >= 0.4",
-    "@opam/result": "*",
     "@opam/ocaml-migrate-parsetree": "*",
-    "@opam/dune": "*"
+    "@opam/ocamlfind": "*",
+    "@opam/result": "*",
+    "@opam/utop": " >= 1.17.0 < 2.3.0",
+    "@reason-native/rely": "^3.1.0",
```
jordwalke: Rely should be a dev dependency, and there should be another empty file at the root called `reason-test`, and then you can add a field to "esy" field called `buildDev` which does `dune build -p reason,rtop,reason-dev`. This way, not everyone that depends on reason has to depend on rely.

testExe/dune
sync: when i make it part of the reason-test package and run esy x test i get the following error: 

```
error: unable to resolve command: RunReasonTests.exe

esy: exiting due to errors above
```

sync: @jordwalke do you know what to do, i'll also have a look on other projects

test/_snapshots/formatTest_unittests.fff7ed63.0.snapshot
sync: when no escape rely complains it cannot parse the file, same for emojis

testExe/dune
sync: looking at this at the moment https://github.com/facebookexperimental/reason-native/blob/de93a483355d9133be42ff25ace7a3afdcf03974/tests/dune

jordwalke: Yeah, I think I know why it does this. `esy x cmd` simulates your package environment from the perspective of someone that installed it / depended on it. Therefore it doesn't run the `buildDev` command - but the regular build command.

I think I just realized some good news though: You don't need to make a separate `buildDev` command - you can just use the regular `build` command. The reason why, is that we won't actually ship this `esy.json` file - no we have smaller separate `.json` files for each package `reason.json`/`rtop.json` etc. So no worries about the separate buildDev command - that should clear up your challenge.

test/_snapshots/formatTest_unittests.fff7ed63.0.snapshot
bandersongit: It looks like snapshot escaping is part of the original snapshot work @kyldvs did for the old internal TestRunner and has survived intact. I agree that not using String.escaped/emoji handling would be good to have. Curious to hear what @kyldvs's thoughts are. I can file something if we agree this is something that is worth addressing in a future release  (probably via a configuration option to not break everyone's snapshots?)

kyldvs: I'm pretty sure you need to escape it for non-printable characters to have a good diff output.

There are also some crazy edge cases we ran into with snapshots and escaping with Jest, but I think they had JSON output so it might not apply here. I would just be careful when messing around with things here.

davesnx:
Hey @sync, It seems like there's only conflicts the reason why this hasn't landed.

I might be able to take care of those to merge it, is that cool with you?

jordwalke:
Can't wait to have much faster/better tests. The old bash script we hacked together has lived far passed its prime!

sync:
@davesnx off course feel free to take over. But there is more to it than just fixing merge conflicts. I've been trying to find time to go back to it but hasn't happened yet :-) If you are a maintainer you should be able to push to this branch.

davesnx:
Sure, I don't have permisions to push the branch. Happy to chat what's left on Discord.

PR #2457, opened 2019-10-02
Bump allowed utop version to support OCaml 4.08

glennsl:
This is needed to get oni2 to build on 4.08.1. See https://github.com/onivim/oni2/pull/801.

I have now idea how reason was made to build on 4.08.1 without bumping utop...

### DIFF
--- a/esy.json
+++ b/esy.json
@@ -8,7 +8,7 @@
     "@opam/fix": "*",
     "@opam/ocamlfind": "*",
     "@opam/menhir": " >= 20170418.0.0",
-    "@opam/utop": " >= 1.17.0 < 2.3.0",
+    "@opam/utop": " >= 1.17.0 < 2.5.0",
     "@opam/merlin-extend": " >= 0.4",
     "@opam/result": "*",
     "@opam/ocaml-migrate-parsetree": "*",


esy.json
```diff
@@ -7,7 +7,7 @@
     "ocaml": " >= 4.2.0 < 4.9.0",
     "@opam/ocamlfind": "*",
     "@opam/menhir": " >= 20170418.0.0",
-    "@opam/utop": " >= 1.17.0 < 2.3.0",
+    "@opam/utop": " >= 1.17.0 < 2.5.0",
```
anmonteiro: I would even remove the upper bound constraint

anmonteiro:
> I have now idea how reason was made to build on 4.08.1 without bumping utop...

It probably never was in the esy CI step, given that Esy is allowed to pick any OCaml version between 4.02 and 4.09

anmonteiro:
> This is needed to get oni2 to build on 4.08.1.

BTW OCaml 4.09 is out now, which is basically 4.08.1 + a couple patches.

glennsl:
> BTW OCaml 4.09 is out now, which is basically 4.08.1 + a couple patches.

Yeah, unfortunately merlin, jsoo and I think some other packages haven't updated yet. I think there are internal changes that make it not so straightforward to upgrade for tools that depend on compiler internals?

esy.json
ulrikstrid: I think that upper bound was added because 2.3.1 was broken on Windows

anmonteiro:
@glennsl fair. For Merlin there's a WIP PR here: https://github.com/ocaml/merlin/pull/1031, not sure about JSOO though.

anmonteiro:
>  I think there are internal changes that make it not so straightforward an upgrade for tools that depend on compiler internals?

I think this is mostly only true when the parsetree / typedtree changes between version of OCaml -- in the 4.08.1 -> 4.09.0 upgrade there aren't such changes so it makes tools _more_ compatible.

jordwalke:
Only concerning thing here is that windows utop issue for > 2.3.0.

glennsl:
I'm guessing it's this one? https://github.com/ocaml-community/utop/issues/272 If so, it was fixed back in April.

glennsl:
bump

I don't think the CI failure is related to this one-liner, and this is still blocking us from upgrading to 4.08.

jordwalke:
Azure is passing on master so it might be related to this one liner. I can merge, but I might have to revert if it breaks master.

PR #2459, opened 2019-10-03
Use stages and build matrix

ulrikstrid:


ulrikstrid:
@jordwalke I think this is good to go if the UI works for you. I was having problems seeing any run for some reason

jordwalke:
Any downside to landing this?

ulrikstrid:
This is the more modern approach to azure pipelines and is a nice way of structuring the different parts. This will set us up better for creating releases on the future.
Using matrixes also makes it easier to see what is going to run from just looking at the yaml.
Personally I can't see any downsides, but I love new things and this is in "preview".

sync:
Looking good 👍

PR #2462, opened 2019-10-08
Store refmt as a CircleCI artifact

Schniz:
This commit is for being able to download latest `refmt` on CI without installing and compiling OCaml and the `reason` package, as I mentioned in https://github.com/Schniz/fnm/pull/151

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -167,6 +167,9 @@ jobs:
           <<: *esy_post_build_cache_key
           paths:
             - ~/.esy
+      - store_artifacts:
+          path: _esy/default/build/default/src/refmt/refmt_impl.exe
+          destination: refmt.exe
 
 workflows:
   version: 2


jordwalke:
CircleCI CI is likely going to be deprecated now that we have azure all figured out. I recently made some changes where every commit will publish a reason-cli artifact on azure.

You can check it out here for the latest master commit:
https://dev.azure.com/reasonml/reason/_build/results?buildId=403&view=artifacts

There's a .tgz file in the Release section which can be installed and gives refmt and rtop.

We could create another release that just has refmt so that it would be lighter weight.

Also, I learned we can make platform specific releases so that the total package size will be much less. Instead of bundling them all into one, we can use optionalDependencies: https://stackoverflow.com/questions/15176082/npm-package-json-os-specific-dependency/26069595

Schniz:
Alright! I made a GH action that uses the latest version of `reason` using Azure — https://github.com/Schniz/fnm/pull/152/checks?check_run_id=322770191

I'm not sure that it's good though, because I'm not sure you can look for a specific git tag (if this is how we'll do versioning) and how long will the artifact be stored in Azure.

I think we can also make a smaller artifact, that will just have the .tgz file and then it'll be faster to download. It takes 1m:36s to download that huge >1gb zip file when we only need the .tgz release file

what do you think?

davesnx:
This is not needed anymore, but would be nice to push `refmt` as GH releases instead

PR #2464, opened 2019-10-10
Make external... = "" optional in reason syntax

### LINKED ISSUE #2422: Make ` = ""` optional in reason syntax
```reason
[@bs.obj]
external bsObjMake: (~comment: string=?, ~platform: string, unit) => _ = "";
```
it would be nice to just have
```reason
[@bs.obj]
external bsObjMake: (~comment: string=?, ~platform: string, unit) => _  ;
```

romanschejbal:
Fixes #2422

### DIFF
--- a/formatTest/typeCheckedTests/input/attributes.re
+++ b/formatTest/typeCheckedTests/input/attributes.re
@@ -360,6 +360,10 @@ external readFileSync2 : (
 /* Ensure that attributes on extensions are printed */
 [@test [@attr] [%%extension]];
 
+external debounce : (int, [@bs.meth] unit) => unit;
+
+external debounce : (int, [@bs.meth] unit) => unit = "debounce";
+
 external debounce : (int, [@bs.meth] unit) => unit = "";
 
 external debounce : int => ([@bs.meth] (unit => unit)) = "";
--- a/src/reason-parser/reason_parser.mly
+++ b/src/reason-parser/reason_parser.mly
@@ -1615,7 +1615,12 @@ structure_item:
     | item_attributes
       EXTERNAL as_loc(val_ident) COLON core_type EQUAL primitive_declaration
       { let loc = mklocation $symbolstartpos $endpos in
-        mkstr (Pstr_primitive (Val.mk $3 $5 ~prim:$7 ~attrs:$1 ~loc)) }
+        let prim = if $7 = [""] then [$3.txt] else $7 in
+        mkstr (Pstr_primitive (Val.mk $3 $5 ~prim ~attrs:$1 ~loc)) }
+    | item_attributes
+      EXTERNAL as_loc(val_ident) COLON core_type SEMI
+      { let loc = mklocation $symbolstartpos $endpos in
+        mkstr (Pstr_primitive (Val.mk $3 $5 ~prim:[$3.txt] ~attrs:$1 ~loc)) }
     | type_declarations
       { let (nonrec_flag, tyl) = $1 in mkstr(Pstr_type (nonrec_flag, tyl)) }
     | str_type_extension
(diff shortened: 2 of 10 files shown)


jordwalke:
Great change!

src/reason-parser/reason_parser.mly
```diff
@@ -1705,6 +1705,10 @@ structure_item:
       EXTERNAL as_loc(val_ident) COLON core_type EQUAL primitive_declaration
       { let loc = mklocation $symbolstartpos $endpos in
         mkstr (Pstr_primitive (Val.mk $3 $5 ~prim:$7 ~attrs:$1 ~loc)) }
+    | item_attributes
+      EXTERNAL as_loc(val_ident) COLON core_type SEMI
+      { let loc = mklocation $symbolstartpos $endpos in
+        mkstr (Pstr_primitive (Val.mk $3 $5 ~prim:[""] ~attrs:$1 ~loc)) }
```
anmonteiro: I think instead of `~prim:[""]` we may want to add the name in `val_ident`. Especially because BuckleScript has started to produce warnings about `""` in externals, and people will have a hard time figuring out that we're generating it at the AST level.

romanschejbal: I see... I've just updated the PR, do you mind looking again? :) Thanks for reviewing!

anmonteiro: I think the new change is a good start -- but now the old version will not print to the new one. I think in the printer you still need the `[""]` case in the match statement

romanschejbal: I did not realize that `external map: int = "map";` and `external map: int = "";` are the same 🙈 should be ok now, thank you

anmonteiro:
@romanschejbal I made another pass at this -- it seems that you still need to handle the `.rei` case.

formatTest/typeCheckedTests/input/attributes.re
```diff
@@ -360,6 +360,10 @@ external readFileSync2 : (
 /* Ensure that attributes on extensions are printed */
 [@test [@attr] [%%extension]];
 
+external debounce : (int, [@bs.meth] unit) => unit;
+
+external debounce : (int, [@bs.meth] unit) => unit = "debounce";
+
 external debounce : (int, [@bs.meth] unit) => unit = "";
```
anmonteiro: maybe not `.rei` files, I might have looked too hastily. Why are these still in the old format?

romanschejbal: I thought we want to be backward compatible, so I've kept the old style there for those "input" files.

anmonteiro: but the output still has the empty string (`""`) or am I reading it wrong?

anmonteiro: oh I think I brainfarted... Apologies!

romanschejbal:
Thanks to you for the intro and help! And yes, that makes total sense. I will look into it when I get some time. :-)

romanschejbal:
I think this one is now good to go. It removes those empty strings during parsing phase so we don't get bs warnings. Do you mind taking another look @anmonteiro ?

And sorry if the rebase messed up the code-reviewing.

Thanks

jordwalke:
Thank you for your contribution @romanschejbal !

bobzhang:
> I think this is good to go.
> 
> @romanschejbal thanks again for working on this. I think a good next step could be to parse `external x: ... = "";` into `external x .. = "x"` and avoid having empty strings in the AST at all. I think this would get us rid of the bucklescript warnings.

in bs, bs.obj expect string to be empty, that's why I made such feature request

jordwalke:
So what is the desired behavior? Should it default to empty if there is no string contents supplied?

jordwalke:
I do like the behavior of having `external x : typ;` defaulting to `= "x"`. It is really nice for c bindings, but I agree we cannot make a breaking change without coordinating a major version bump. In future versions of bs, is it possible to relax the requirement that the string content be empty in the AST? Is there a reason why `bs.obj` cannot ignore the string content? (Again, in the future - not right now because of the breaking change).

bobzhang:
See my comments on punning: it is fragile to refactoring (both for C and JS)
So the desired behavior should be empty string, but I am unclear why it broke the CI since we did not use this feature yet

ryyppy:
I think what the desired behavior should be:

```
external x: string => string = "";
```

Should not convert it to 

```
external x: string => string = "x";
```

I think this is what's happening here? If so, I consider this a bug?

romanschejbal:
Yes that's what's happening here. Anyway, I'm happy to create another PR rn that'll preserve the empty string if that's what you'll settle on. :)

bobzhang:
@romanschejbal thanks for looking into it. 

desired behavior
- external f :  'a => 'b ; is a short-cut for empty string
- no punning involved

romanschejbal:
@bobzhang see the new PR, hope it's correct! :)

PR #2468, opened 2019-10-15
Make prepublish script auto-generate opam files

jordwalke:


jordwalke:
Taken from reason-native repo

PR #2473, opened 2019-10-26
Attempt to fix the braces removal in JSX props

### LINKED ISSUE #2471: JSX: refmt removes braces by error
refmt version: Reason 3.5.0 @ 8ee1ff66

```reason
<Component
  prop={
    x->Option.map(x => {let y = x; y ++ y})
  }
/>
```

gets refmtd to:

```reason
<Component
  prop={
    x->Option.map(x =>
      let y = x;
      y ++ y;
    )
  }
/>;
```

which is invalid.

bloodyowl:
This is a very empirical attempt, I'm not exactly sure if it's a safe patch.

I basically just modified the printer bit by bit until I found something that'd fix my test case.

I also tried to change the `dont_preserve_braces` in the `Pexp_apply` case in `formatJSXComponent` but it seemed to be a bit more aggressive. The issue does seem to flow down from there though.

Fixes #2471

### DIFF
--- a/formatTest/unit_tests/expected_output/jsx.re
+++ b/formatTest/unit_tests/expected_output/jsx.re
@@ -658,3 +658,12 @@ let v =
       }}
     </B>
   </A>;
+
+<Component
+  prop={
+    x->Option.map(x => {
+      let y = x;
+      y ++ y;
+    })
+  }
+/>;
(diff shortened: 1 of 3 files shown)


bloodyowl:
Close in favour of #2474

PR #2495, opened 2019-11-19
Update bspack workflow for distributing BS6 compatible artifacts

ryyppy:
We need a way to easily distribute bspacked `reason_api.ml` and `reason_binary.ml` files for BuckleScript v6 and above (4.06), mostly because we want to keep the Reason JS playground bundle up to date with most recent Reason releases.

This PR keeps the original workflow alive (with slight adaptions in the README), but marks it as a legacy build workflow (BuckleScript will not support 4.02 in the future). Right now the 4.02 bspack instructions fail, mostly due to breaking updates to Reason, but it's still good to keep everything around for now.

The new workflow is introduced as `bspacks/reason_bspack406.sh` script, which needs `BSPACK_EXE` to be defined (pointing to a locally built bspack for now). Example usage calls can be found in the script comments.

Also, the 406 script takes the version from the `reason.json` file (since it looks like the source of truth for Reason version numbers). Easier for the user to not make mistakes.

Any feedback would be valuable.

@thangngoc89

### DIFF
--- a/bspacks/reason_bspack406.sh
+++ b/bspacks/reason_bspack406.sh
@@ -129,7 +129,7 @@ $BSPACK_EXE \
 
 
 # This hack is required since the emitted code by bspack somehow adds 
-# 'Migrate_parsetree__Ast...' instead of 'Migrate_parsetree_Ast...'
+# 'Migrate_parsetree_Ast...' instead of 'Migrate_parsetree__Ast...'
 sed -i'.bak' -e 's/Migrate_parsetree__Ast_404/Migrate_parsetree.Ast_404/' build/*.ml
 
 # the `-no-alias-deps` flag is important. Not sure why...
--- a/bspacks/reason_bspack406.sh
+++ b/bspacks/reason_bspack406.sh
@@ -129,7 +129,6 @@ $BSPACK_EXE \
 
 
 # This hack is required since the emitted code by bspack somehow adds 
-# 'Migrate_parsetree_Ast...' instead of 'Migrate_parsetree__Ast...'
 sed -i'.bak' -e 's/Migrate_parsetree__Ast_404/Migrate_parsetree.Ast_404/' build/*.ml
 
 # the `-no-alias-deps` flag is important. Not sure why...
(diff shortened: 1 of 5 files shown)


bspacks/refmtJsApi.ml
```diff
@@ -1,3 +1,9 @@
+(*
+ * Note: This file is currently broken, since Reason removed
+ * Refmt_api.Reason_syntax_util.Error in favor of Reerror's `Printexc.to_string e`
+*)
+open Js_of_ocaml
+
```
bobzhang: what's the rational of opening such module

ryyppy: I added this to get the missing JSOO functions linked used inside this script. This file is not used for any build artifacts used by our playground. I think I will undo this open just for the sake of not messing with the file

ryyppy: dropped the line to be 100% consistent with the original file

bspacks/reason_bspack406.sh
```diff
@@ -0,0 +1,140 @@
...
+  -I "$reasonTargetDir/_build/default/src/reason-parser/vendor/cmdliner/"     \
+  -I "$reasonTargetDir/_build/default/src/refmt/"                             \
+  -I "$reasonTargetDir/_build/default/src/refmttype/"                         \
+  -I "$ocamlMigrateParseTreeTargetDir" \
+  -bs-MD \
+  -o "$REFMT_API.ml"
+
+
+# This hack is required since the emitted code by bspack somehow adds 
+# 'Migrate_parsetree__Ast...' instead of 'Migrate_parsetree_Ast...'
```
thangngoc89: Instead of Migrate__parsetree.Ast...

ryyppy:
@jordwalke I think this is ready to be merged. I looked through @jaredly's old PR regarding his `esy` setup (#2153), but it seems that 1) there's still something broken with the menhir dep 2) it's a more complex change for me and I can't tell how much effort it will take to get this working without shaving a yak.

This solution is definitely working for BS contributors (the target group of this change anyways). Happy to improve things in the future as soon as we figured out what is needed?

jordwalke:
Sounds good! If the bs-contributors are happy with the workflow, then it does the job well! Most important thing is that it's correct. Thanks for the contribution.

jordwalke:
Thanks @ryyppy

Issue #2502, opened 2019-11-27
regressions  when upgrade in bucklescript 7


bobzhang:
- [ ] https://github.com/BuckleScript/bucklescript/issues/3987

- [x] https://github.com/BuckleScript/bucklescript/issues/3969

- [ ] https://github.com/facebook/reason/issues/2493

bobzhang:
@anmonteiro can you reopen it since not all of them are fixed?

anmonteiro:
@bobzhang what's not fixed? As I said in https://github.com/facebook/reason/issues/2499#issuecomment-559883108, the JSX PPX is not maintained in this repo anymore.

bobzhang:
@anmonteiro thanks, I will ping @rickyvetter

Issue #2493, opened 2019-11-18
3.5.0: Strange reformatting of JSX props / labeled args


cknitt:
Original format by Reason 3.4.2 @ 3c6a9ca9 (BuckleScript 5.2.1):

```reason
<ActionButton
  disabled={!validated}
  errorText={msg##errorText}
  onPress={_evt =>
    confirmDepartureReadiness(~operationId={operation.operationId}, ~vehicleId, state)
  }>
```

Reformatted by Reason 3.5.0 @ 6657af4 (BuckleScript 7.0.0):

```reason
<ActionButton
  disabled={!validated}
  errorText={
    msg##errorText;
  }
  onPress={_evt =>
    confirmDepartureReadiness(
      ~operationId={
        operation.operationId;
      },
      ~vehicleId,
      state,
    )
  }>
```

rusty-key:
It's only a problem for the first level, it works fine for children:
```reason
<ActionButton
  disabled={!validated}
  errorText={
    msg##errorText;
  }>
  <InactionText
    color=Primary
    someProp={msg##prop}
    errorText={msg##errorText}
  />
</ActionButton>
```

PR #2503, opened 2019-11-30
Inline braces in JSX

anmonteiro:
This is a fix for the last regression reported by BuckleScript v7 that can be fixed in this repo.

The remaining one is a JSX PPX regression that doesn't live in this repo anymore.

fixes https://github.com/facebook/reason/issues/2493
fixes https://github.com/facebook/reason/issues/2502


jordwalke:
Looks good!

Issue #2456, opened 2019-10-02
Up to date reason npm package


baransu:
The latest npm `reason` package version is `3.3.4` which is over a years old. We want to use it for [reformatting between Reason and OCaml syntax](https://github.com/reason-association/reasonml.org/pull/34) and would be great to have access to up to date `refmt` version. Is there any reason why the package is not updated with every new release?

ryyppy:
Now that I understand the vendoring process I can give more context:
The `bspack.exe` has been removed from recent `bs-platform` releases, which now blocks Reason's `bspacks` scripts to create a single `refmtJS.ml`  bundle and run a build for `refmt.js`.

This apparently happened around 1 year ago I assume.

thangngoc89:
@ryyppy this isn't true. You don't need bspack to produce refmt.js . Probably just lacks of man-power

ryyppy:
This is either solved by merging #2495 (opam based), or finishing work in #2153 (esy based)

jchavarri:
@thangngoc89 You mean refmt.js could be produced using a regular dune + jsoo setup? How would bucklescript compilation work? (side note: I always thought bspack was created originally because BS was a fork of OCaml and so could not be compiled with regular build tooling, but maybe that assumption is false and that's why I'm confused).

cc @ulrikstrid as we've been discussing about bspack stuff lately :)

thangngoc89:
@jchavarri refmt.js has been always produced by jsoo. Adding bspack into that is just noisy and more complexities for no reasons at all. bspack is for vendoring to bucklescript only

jchavarri:
> You mean refmt.js could be produced using a regular dune + jsoo setup?

So the answer to that is yes?

thangngoc89:
Yes. refmt.js SHOULD be produce by a regular dune + jsoo setup

anmonteiro:
We don’t release to NPM anymore

PR #2504, opened 2019-12-02
Use dune instead of a custom bash script to build js_of_ocaml refmt.js

jaredly:
I can't find the issue comment now, but someone mentioned that we should be able to switch this over to use dune without too much trouble, and they were right!


jaredly:
Oh here's the comment https://github.com/facebook/reason/issues/2456#issuecomment-559869881 cc @thangngoc89

esy.json
```diff
@@ -7,6 +7,7 @@
     "ocaml": " >= 4.2.0 < 4.10.0",
     "@opam/fix": "*",
     "@opam/ocamlfind": "*",
+    "@opam/js_of_ocaml": "*",
```
jordwalke: I'd prefer that typical development builds not require a dependency on jsoo. Could we make a new `.json` file that extends this one, and adds jsoo? Or at the very least, make this a devDependency so that when we make releases it doesn't include jsoo prebuilt binaries needlessly.

jordwalke:
Cool, just one comment. You feel pretty good about this change, otherwise?

package.json
```diff
@@ -21,6 +21,6 @@
 		"bs-platform": "1.9.3"
 	},
 	"scripts": {
-		"prepublishOnly": "./bspacks/downloadSomeDependencies.sh && ./bspacks/reason_bspack.sh"
+		"prepublishOnly": "esy dune build js/refmt.bc.js && ./js/closurefy.sh"
```
thangngoc89: This should be built with release profile

jordwalke: So that would be:
```
esy dune build -p reason js/refmt.bc.js
```
correct?
Why does it need release profile?

thangngoc89: If you’re planning to distribute this for consumption then use release profile. There are many optimizations only enabled on release profile.
 You’re the expert on jsoo by the way.

jaredly: jsoo fails with a stack overflow error in release mode 🤷

jaredly:
@jordwalke yup, I think this is ready to go!

jordwalke:
Thank you!

PR #2517, opened 2019-12-12
Fix OCaml identifier mapper for Pexp_field

### LINKED ISSUE #2513: Escaping "method" still does not work correctly
With the latest BuckleScript master (BuckleScript 7.0.2-dev.1 @ b692f6b / Reason 3.5.1 @ 0b7f4e9), 

```reason
type request = {method: string};

let x = {method: "GET"};

let y = x.method;
```

gives me

```
  This expression has type request
  The field method does not belong to type request
Hint: Did you mean method_?```

### PR DESCRIPTION
**Pull request by anmonteiro**
fixes #2513

### DIFF
--- a/formatTest/unit_tests/expected_output/ocaml_identifiers.re
+++ b/formatTest/unit_tests/expected_output/ocaml_identifiers.re
@@ -89,4 +89,4 @@ let x =
     y;
   });
 
-let y = x.method_;
+let y = x.method;
--- a/formatTest/unit_tests/expected_output/ocaml_identifiers.re
+++ b/formatTest/unit_tests/expected_output/ocaml_identifiers.re
@@ -88,3 +88,5 @@ let x =
     ();
     y;
   });
+
+let y = x.method_;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by cristianoc** [comment]
@anmonteiro `Pexp_setfield` missing.

Also, could we have the complete test:
```
type request = {method: string};

let x = {method: "GET"};

let y = x.method;
```

Then a test with mutable field update.

Plus a pattern matching test, to make sure that pattern matching is also renamed?

**Comment by cristianoc** [comment]
Since we're on the subject, there's no sub in `typs` below:
```
    | Ptyp_constr (lid, typs) ->
      Ptyp_constr (map_lident f lid, typs)
```

https://github.com/facebook/reason/issues/2519

anmonteiro:
I think that’s by design. AST mapper maps over the Core types we don’t need to do it ourselves

cristianoc:
@anmonteiro ah that mapper is written differently from the others. Looks like the issue is fixed in master already.

cristianoc:
@anmonteiro  continuing here: https://github.com/facebook/reason/pull/2520.

PR #2521, opened 2019-12-13
Bump version to 3.5.2

ryyppy:
I verified if everything works with the following:

```sh
make clean
make pre_release
make build

_build/default/src/refmt/refmt_impl.exe --version
Reason 3.5.2 @ 2553bdc
```


ryyppy:
This is wrong, we are currently at version 3.5.3, so the next version must be 3.5.4. Will create another PR

PR #2529, opened 2019-12-31
Add support for OCaml 4.10

kit-ty-kate:
Currently reason does not support OCaml 4.10 and the first beta is around the corner. This PR adds support for OCaml 4.10

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -118,6 +118,14 @@ jobs:
       - OCAML_VERSION: "4.09.x"
       - NPM_CONFIG_PREFIX: "~/.npm-global"
     <<: *common_steps
+  4.10.0:
+    docker:
+      - image: ocurrent/opam:debian-9-ocaml-4.10
+    environment:
+      - TERM: dumb
+      - OCAML_VERSION: "4.10.x"
+      - NPM_CONFIG_PREFIX: "~/.npm-global"
+    <<: *common_steps
   esy_build:
     docker:
       - image: ocaml/opam2:debian-9-ocaml-4.06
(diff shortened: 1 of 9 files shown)


anmonteiro:
Thanks, @kit-ty-kate!

I predict this is gonna need some 4.10 specific expected test outputs as in https://github.com/facebook/reason/blob/master/formatTest/errorTests/expected_output/comments1.re.4.09.0

Should probaly be enough to copy them from the 4.09 ones.

jordwalke:
Can we just alias the 4.10 expected test outputs to the 4.09 ones since we don't expect any changes?

kit-ty-kate:
@jordwalke was that what you were asking?

The CI image used is a different one because the `ocaml/opam2` images haven't been updated for a month and I feel like testing beta version requires at least a recent enough version of the compiler. cc @avsm. `ocurrent/opam` are supposed to be the successor of `ocaml/opam2` anyway, so why not use them. At the very least to show that everything works, I can remove it before this PR is merged if it's an issue.

jordwalke:
Thanks for thinking about `ocurrent/opam:`. I just don't know anything about it. If you revert that part, I can merge this, or alternatively, if you could point me to somewhere that suggests that is the right/supported new opam image? I appreciate this pull request!

anmonteiro:
hey @kit-ty-kate I think it's fine to use the ocurrent docker image, but you missed adding the pipeline for 4.10.0 [here](https://github.com/facebook/reason/blob/master/.circleci/config.yml#L181)

kit-ty-kate:
For information: the final release of OCaml 4.10.0 is expected next week

kit-ty-kate:
OCaml 4.10.0 has now been released.

Issue #2523, opened 2019-12-20
Regression: escaping of "method" breaks interop with bs.obj


cknitt:
The just released BS 7.0.2-dev.1 with the latest refmt compiles

```reason
type request;

[@bs.obj] external request: (~method: string) => request = "";

Js.log2("request", request(~method="GET"));
```

to

```js
console.log("request", {
      method_: "GET"
    });
```

I.e. it gives me an object with a field `method_`. In previous versions, I used to get an object with a field `method` (as expected).

cknitt:
(See #2507 and #2513 for previous issues with escaping "method".)

cknitt:
Also,

```reason
let f = x => x##method;
```

gives

```js
function f(x) {
  return x.method_;
}
```

cknitt:
Same for React components, as they are using `bs.obj` internally:

```re
module SomeComponent = {
  [@react.component]
  let make = (~method) => React.string(method);
};

<SomeComponent method="GET" />;
```

gives

```js
React.createElement(Teest$SomeComponent, {
      method_: "GET"
    });
```

cknitt:
Maybe `method` should just be forbidden and not escaped? Then the user is forced to use

```reason
[@bs.obj] external request: (~_method: string) => request = "";
```

which works fine.

anmonteiro:
I don't think this is a bug. We did switch the representation, but it's "correct" now. Is the BuckleScript contract that the generated JS names can be relied upon?

If we revert this change then there are bugs interacting between Reason and OCaml, and IMO those need to remain fixed.

cknitt:
I don't know if there is any "official" BuckleScript contract for that, but as a user of BuckleScript defining an external method in Reason syntax

```reason
[@bs.obj] external request: (~a: string, ~b: string, ~c: string) => t = "";
```

I definitely expect the resulting JS object to have the fields `a`, `b` and `c`, not "randomly" `a`, `b` and `c_`.

If the compiler told me that I am not allowed to use `c` because it is a reserved word, that would be fine with me, because then I'd **know** that something is wrong and could explicitly escape it myself by using `_c` (which BuckleScript changes to `c` in the JS output).

cknitt:
For example,

```reason
[@bs.obj] external request: (~type: string) => t = "";
```

gives me a compile error:

```
Error: type is a reserved keyword, it cannot be used as an identifier.
Try `type_` or `_type` instead
```

I then change it to 

```reason
[...]

PR #2530, opened 2020-01-09
Add flag rename_labels to toggle renaming of labels in function defin…

cristianoc:
…ition/application and records.

The flag `rename_labels` can  be used to turn on/off renaming of labels for function definition/application (i.e. the labels of named arguments) and record creation/access (i.e. the labels of records).

Issues:
- https://github.com/facebook/reason/issues/2523
- https://github.com/reasonml/reason-react/pull/505

Relevant diffs:
- https://github.com/facebook/reason/commit/ec2c797f9ea82517e63c785d9cabf5a2e677babc
- https://github.com/facebook/reason/commit/91a51913f08523dbe6d784d69af1ef0152c0bfea
- https://github.com/facebook/reason/commit/d0bd920124027fd95290af6ee5e79b9da6887ebb

### DIFF
--- a/formatTest/unit_tests/expected_output/ocaml_identifiers.re
+++ b/formatTest/unit_tests/expected_output/ocaml_identifiers.re
@@ -12,20 +12,20 @@ module V = {
 module R = {
   type r = {mutable method: int};
 
-  let foo = {method: 4};
+  let foo = {method_: 4};
 
-  let x = foo.method;
+  let x = foo.method_;
 
-  let () = foo.method = 42;
+  let () = foo.method_ = 42;
 
   let y =
     switch (foo) {
-    | {method} => method
+    | {method_: method} => method
     };
 
   let z =
     switch (foo) {
-    | {method: 12} => 21
+    | {method_: 12} => 21
     };
 };
 
@@ -69,7 +69,8 @@ module O = {
 
 /* Function parameter labels */
 module L = {
-  let f = (~method) => ignore(method);
+  let f = (~method_ as method) =>
+    ignore(method);
 };
 
 /* Module types */
@@ -90,14 +91,14 @@ type method = string;
 [@other_attr: method]
 type foo = {method};
 
-let f = (~method) => Js.log(method);
+let f = (~method_ as method) => Js.log(method);
 
-let x = f(~method="GET");
+let x = f(~method_="GET");
 
 type marshalFields = {. "switch_": string};
 
 let testMarshalFields: marshalFields = {
-  "switch_": "switch",
+  "switch": "switch",
 };
 
 /* Not an identifier test, but this is testing OCaml -> RE */
--- a/formatTest/unit_tests/expected_output/ocaml_identifiers.re
+++ b/formatTest/unit_tests/expected_output/ocaml_identifiers.re
@@ -10,7 +10,7 @@ module V = {
 
 /* Record fields */
 module R = {
-  type r = {mutable method: int};
+  type r = {mutable method_: int};
 
   let foo = {method_: 4};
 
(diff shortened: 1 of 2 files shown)


bobzhang:
> labels for function definition/application (i.e. the labels of named arguments) and record creation/access

So they are mangled or not? I wish we have some consistency (instead of make it configurable)

formatTest/unit_tests/expected_output/ocaml_identifiers.re
```diff
@@ -10,22 +10,22 @@ module V = {
 
 /* Record fields */
 module R = {
-  type r = {mutable method: int};
+  type r = {mutable method_: int};
```
jordwalke: Asked offline. I didn't think that you wanted to revert the record field mangling. Was that intentional?

src/reason-parser/reason_syntax_util.cppo.ml
```diff
@@ -24,6 +24,9 @@ open Ast_mapper
 open Parsetree
 open Longident
 
+(* Rename labels in function definition/application and records *)
+let rename_labels = ref false
```
jordwalke: We're going to flip this to true for non-bs usage pretty soon. Is that going to be easy to coordinate with BS in the next next update? Would it be easier to just flip this to true, and let BS flip it to false from the outside? Don't need to block on that decision, but just curious if there's a default that can be set here that makes it easier later.

formatTest/unit_tests/expected_output/ocaml_identifiers.re
cristianoc: With records being used in place of objects, a difference in mangling would come up pretty soon.
To be on the safe side, I've kept refmt from bs 6.2.1 as a reference.
There are still a couple of difference to take into account, if one wants to have something that behaves like a revert. E.g. mangling of the name in type declarations.

src/reason-parser/reason_syntax_util.cppo.ml
cristianoc: Sure flipping on the bs side can be done. One other question is what to do longer term with tests, if bs and native behave differently.

anmonteiro: This is gonna be my last comment in this discussion, since it seems like my concerns aren't really getting heard: I strongly disagree that BuckleScript and Reason should behave differently in the long term. This is introducing a dangerous precedent for deviating even more from OCaml.

I guess I'm OK with this PR to go in right now to fix the immediate issue of the breaking change we accidentally introduced. But I'm very strongly in favor or moving the ecosystem towards a world where we do output an OCaml compatible AST.

cristianoc: The longer term situation needs to be looked at carefully, and will take more time.
One of the proposals is to have a compatibility mode, where you don't get to use keywords from the other language at all. In that proposal, bs and native would behave identically.
One evolution of the compatibility mode is to also ensure that platform-specific libraries/features are not used. So Js.* can't be used in code shared between Js and native, and you get an error/warning rather than having to think about it yourself. This is kind of the dual approach to trying to unify libraries: let libraries specialize optimally in their domain when it's important, but document the differences.

formatTest/unit_tests/expected_output/ocaml_identifiers.re
jordwalke: On the native side we pretty much want all the mangling we can get for greater compatibility. On the bs side, you can disable whatever you like under a flag. I just wasn't sure if you wanted something that behaved like a perfect revert. If you're fine with it, go ahead and merge.

src/reason-parser/reason_syntax_util.cppo.ml
jordwalke: Let's merge this PR just to unbreak people, but we can continue the discussion on this thread about what to do next.

PR #2537, opened 2020-01-28
Make "external f: 'a => 'b;" shortcut for empty string

romanschejbal:
This is a followup to #2464 based on the discussion in there.

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/attributes.re
+++ b/formatTest/typeCheckedTests/expected_output/attributes.re
@@ -405,9 +405,10 @@ type reactClass;
 
 /* "react-dom" shouldn't spread the attribute over multiple lines */
 [@bs.val] [@bs.module "react-dom"]
-external render: (reactElement, element) => unit;
+external render: (reactElement, element) => unit =
+  "render";
 
-[@bs.module "f"] external f: int => int;
+[@bs.module "f"] external f: int => int = "f";
 
 [@bs.val] [@bs.module "react"] [@bs.splice]
 external createCompositeElementInternalHack:
@@ -460,8 +461,8 @@ external readFileSync2:
 external debounce:
   (int, [@bs.meth] unit) => unit;
 
-external debounce:
-  (int, [@bs.meth] unit) => unit;
+external debounce: (int, [@bs.meth] unit) => unit =
+  "debounce";
 
 external debounce:
   (int, [@bs.meth] unit) => unit;
--- a/formatTest/unit_tests/expected_output/uncurried.re
+++ b/formatTest/unit_tests/expected_output/uncurried.re
@@ -142,7 +142,8 @@ type timerId;
 
 [@bs.val]
 external setTimeout:
-  ((. unit) => unit, int) => timerId;
+  ((. unit) => unit, int) => timerId =
+  "setTimeout";
 
 let id =
   setTimeout((.) => Js.log("hello"), 1000);
--- a/src/reason-parser/reason_parser.mly
+++ b/src/reason-parser/reason_parser.mly
@@ -1615,12 +1615,11 @@ structure_item:
     | item_attributes
       EXTERNAL as_loc(val_ident) COLON core_type EQUAL primitive_declaration
       { let loc = mklocation $symbolstartpos $endpos in
-        let prim = if $7 = [""] then [$3.txt] else $7 in
-        mkstr (Pstr_primitive (Val.mk $3 $5 ~prim ~attrs:$1 ~loc)) }
+        mkstr (Pstr_primitive (Val.mk $3 $5 ~prim:$7 ~attrs:$1 ~loc)) }
     | item_attributes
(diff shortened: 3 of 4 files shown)


formatTest/typeCheckedTests/expected_output/attributes.re
```diff
@@ -405,9 +405,10 @@ type reactClass;
 
 /* "react-dom" shouldn't spread the attribute over multiple lines */
 [@bs.val] [@bs.module "react-dom"]
-external render: (reactElement, element) => unit;
+external render: (reactElement, element) => unit =
+  "render";
```
anmonteiro: this is regressing from #2464 in that the printer shouldn't print the `= "render";` part

romanschejbal: I'm not sure as the input for this particular output looks like this
```
/* "react-dom" shouldn't spread the attribute over multiple lines */
[@bs.val] [@bs.module "react-dom"]
external render : (reactElement, element) => unit = "render";
```

And the tests are passing, too. There's something else failing on the CI though which I don't understand.

anmonteiro: I think we agreed in the last PR that if the ident and the primitive name agree we can drop it on the floor, but I could be misremembering

romanschejbal: Or did I misunderstand the [assignment from Bob](https://github.com/facebook/reason/pull/2464#issuecomment-579232949)? 😄

romanschejbal: I would like to drop it if they are the same too btw. But I don't know all the in & outs of bs.

anmonteiro: Ah I didn’t read the discussion in the other issue.

I disagree with the outcome, but oh well.

bobzhang:
The diff looks good to me.
Note the old PR changed the semantics that's why it broke our CI.
I argued for dropping the punning since it makes refactoring fragile (for both C and JS) but have no strong opinions -- we used to like punning but figured it does not work with editor based refactoring (cc @chenglou)

formatTest/typeCheckedTests/expected_output/attributes.re
ryyppy: I think you all know this, but I still want to write this out for anyone following this discussion. 

> @anmonteiro I think we agreed in the last PR that if the ident and the primitive name agree we can drop it on the floor, but I could be misremembering

With this behavior, it would be a breaking change for BuckleScript.
The recommendation is to always explicitly write the string value, even if the identifier matches the value itself.

```
[@bs.val] [@bs.module "react-dom"]
external render : (reactElement, element) => unit = "render";
```

Previously, you could leave it out blank as well, but this will now yield a warning that this is discouraged. As @bobzhang  said, the reason is to prevent binding errors due to quick renaming of the identifier (with the punning, you wouldn't even know what's going on, especially if you are newcomer and never used the string representation for some reason).

Why does `[@bs.obj]` require a `= ""` then? This seems to be documented [here](https://bucklescript.github.io/docs/en/object-2#function):

```reason
[@bs.obj] external route: (
  ~_method:string,
  ~path:string,
  ~action:(list(string) => unit),
  ~options:Js.t({..})=?,
  unit
) => _ = "";
```

> Note: the = "" part at the end is just a dummy placeholder, due to syntactic limitations. It serves no purpose currently.

So i guess the whole goal of this general external syntax feature is to get rid of this confusing `[bs.obj]` syntax?

romanschejbal:
Thanks for the clarification @ryyppy 👌 

Can someone point me in the direction of how can I make the CI happy? Menhir version seems to be incompatible?

cristianoc:
CI breaks for unrelated reasons (also happens on a change-nothing diff which used to build OK).

jordwalke:
@anmonteiro I think you and I agree on what would be ideal here punning is nice because if you omit a string, there's really only one thing you could ever want it to be - punned. The only exception is if you have to be using the current version of bs's ppx (which many are). Since breaking people isn't an option, we need to find a way to get to the ideal state without breaking people. Do you have an alternative suggestion to get to the ideal state of punning without breaking anyone?

jordwalke:
@bobzhang What would be involved in making the next release of bs gracefully ignore the contents of the string when using bs.obj? It seems like that would be a non-breaking BS change, that allows us to add this feature back. Without bs being graceful in allowing that, it is currently eating up a nice syntactic real estate for non-bs use cases (needlessly?)

TheSpyder:
With this update, isn't the change list for 3.6.0 incorrect?

> `external myFn: (string) => unit` is now equivalent to `external myFn: (string) => "myFn"`

(putting aside the typo, it should read `=> unit = "myFn"`)

I saw that and expected to be able to remove the `= "myFn"` from my code but I still get BS warning 105. This PR explains why.

PR #2541, opened 2020-01-31
Better handling of longidents in JSX

Drup:
The treatement of longidents in JSX is currently .... curious. 

- `<foo a=b>children</foo>` turns into `foo(~a=b,~children,())`
- `<Foo a=b>children</foo>` turns into `Foo.createElement(~a=b,~children,())`
- `<Foo.Bar a=b>children</foo>` turns into `Foo.Bar.createElement(~a=b,~children,())`
- `<Foo.bar a=b>children</foo>` either refuses to parses or turns into `Foo.bar.createElement(~a=b,~children,())` (depending on the context).

`Foo.bar.createElement` is not a valid longident (there is no way to create such longidents for *values* (there are for types, due to inline records, but that's out of scope). This will give a type error `Unknown *module* Foo.bar`, which is nonsense.

I propose to do the expected thing, which is to turn `<Foo.bar a=b>children</foo>` into `Foo.bar(~a=b,~children,())`. I fixed the printers to recover the syntactic sugar and added appropriate tests.

That new form is not handled by either versions of reasonReact's ppx vendored for tests. I did not try to fix that.

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/jsx.re
+++ b/formatTest/typeCheckedTests/expected_output/jsx.re
@@ -370,12 +370,10 @@ let asd =
   [@foo] <One test=true foo=2> "a" "b" </One>;
 let asd2 =
   [@foo]
-  [@JSX]
-  One.createElementobvioustypo(
-    ~test=false,
-    ~children=["a", "b"],
-    (),
-  );
+  <One.createElementobvioustypo test=false>
+    "a"
+    "b"
+  </One.createElementobvioustypo>;
 
 let span =
     (~test: bool, ~foo: int, ~children, ()) => 1;
--- a/formatTest/unit_tests/expected_output/jsx.re
+++ b/formatTest/unit_tests/expected_output/jsx.re
@@ -721,3 +721,7 @@ let v =
     four={msg##errorText}
   />
 </ActionButton>;
+
+<Foo.bar />;
+
+<Foo.Bar.baz arg="hello" />;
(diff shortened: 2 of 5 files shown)


Drup:
And if you are curious about why I came to fix this: It would make my job easier for https://github.com/ocsigen/tyxml/pull/254 :)

Drup:
> The cause for the weirdness seems like the result of #1927

Ah, well, I'm surprised this was ever merged. Functor applications are not valid in value paths either, and for rather good reasons that @jordwalke mention at the end. 

But that's not today's topic! :)

Drup:
Ping ? :)

jordwalke:
thanks for the fix @Drup. Just to confirm, this does not change the behavior/meaning of any previously valid parse?

Drup:
As far as I can tell no, and all the previous tests pass.

jfrolich:
Is this going to be merged?

anmonteiro:
Turns out this breaks the React JSX PPX, which expects the invalid `Foo.bar.createElement` part. So I'm gonna go ahead and revert that part separately.

Drup:
> Turns out this breaks the React JSX PPX, which expects the invalid `Foo.bar.createElement` part. So I'm gonna go ahead and revert that part separately.

Wait, how do you create such elements ? The JSX can only use them, not create them, and the rest of the grammar disallow creating them.

anmonteiro:
The jsx_element function in Reason Parser seems to create them. There's an example in #2650 of code that broke.

Drup:
Can you detail/discuss this a tiny bit more before reverting ? Because I'm quite convinced that if this breaks the React PPX, maybe *that* should be fixed.

anmonteiro:
For sure. I might have rushed before opening the PR.

I'll try to fix this in the PPX itself. Seems like this is only a problem in my fork of BuckleScript because upstream is never gonna upgrade Reason.

PR #2558, opened 2020-03-31
Lift uncurried label restriction

anmonteiro:
Sounds like this is supported in the next version of BuckleScript: https://bucklescript.github.io/blog/2020/03/26/generalize-uncurry


anmonteiro:
The esy + macOS run seems stuck, merging this.

PR #2563, opened 2020-04-14
Fix rtop build on Windows 

prometheansacrifice:
Fixes the failing macos builds


ulrikstrid:
You can use ubuntu-latest and windows-latest as well

prometheansacrifice:
Fixes CI build on MacOS too. As, Ulrik suggested, I think we should use the latest vmImages

PR #2571, opened 2020-05-06
Attempt fixing prebuilt rtop

prometheansacrifice:
On Windows, the prebuilt rtop seems to fail with `Invalid_argument “String.sub / Bytes.sub”`. We traced it to this [line](https://github.com/esy/esy/blob/master/bin/NpmReleaseCommand.re#L274) and found that indeed on Windows, a `INPUT_ARGUMENTS` is empty (and unused AFAIK)

Until, esy is fixed, we populate it with a dummy value to overcome the issue


jordwalke:
Off topic: I was wondering why the releases are 1Gig. I found that we are including a tgz of the release itself inside the release which is 500MB! So that's half the size.

```
Mode                LastWriteTime         Length Name
----                -------------         ------ ----
d-----         5/7/2020   4:17 AM                bin
d-----         5/7/2020   4:17 AM                platform-darwin
d-----         5/7/2020   4:18 AM                platform-linux
d-----         5/7/2020   4:18 AM                platform-windows-x64
-a----         5/7/2020   4:17 AM            399 package.json
-a----         5/7/2020   4:17 AM           4693 postinstall.js
-a----         5/7/2020   4:17 AM           2861 README.md
-a----         5/7/2020   4:17 AM      603190348 reason-cli-3.6.0.tgz
```

PR #2581, opened 2020-05-28
Fix windows heuristic and ci

jordwalke:
This fixes the reading of heuristics file for converting from ocaml to Reason.
It also fixes the CI import on windows to read from correct cache location.


jordwalke:
Merging to rebuild on master branch.

PR #2582, opened 2020-05-28
Support OCaml 4.11

anmonteiro:
This is an early PR that will hopefully allow us to support OCaml 4.11 when it's out (it's in alpha right now).

This is not expected to build, as it's currently dependent on https://github.com/let-def/merlin-extend/pull/13 being merged.

### DIFF
--- a/esy.json
+++ b/esy.json
@@ -9,7 +9,7 @@
     "@opam/ocamlfind": "*",
     "@opam/menhir": " >= 20170418.0.0",
     "@opam/utop": " >= 1.17.0 < 2.5.0",
-    "@opam/merlin-extend": " >= 0.4",
+    "@opam/merlin-extend": " >= 0.6",
     "@opam/result": "*",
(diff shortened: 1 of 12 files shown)


hhugo:
Note that there is a new syntax for string litteral in 4.11

kit-ty-kate:
For information, OCaml 4.11 is going to be released next week

jordwalke:
@hhugo Where can I read about this string literal syntax?

anmonteiro:
This is the PR for quoted extensions: https://github.com/ocaml/ocaml/pull/8820.

I didn't add support for it because it might warrant its own syntax discussion, so I'll leave that to people who care about those discussions. It doesn't add new AST nodes, so in the worst case we'll print them as strings today.

I'm gonna go ahead and merge this, based on the justification above.

jordwalke:
> so in the worst case we'll print them as strings today.

I suspect it will just print them as quoted strings with extension points surrounding them.

PR #2590, opened 2020-06-15
Test case reproing bad behavior reformatting multiline externals.

jordwalke:
Summary:

Test Plan:

Reviewers:

CC:

### DIFF
--- a/formatTest/unit_tests/expected_output/externals.re
+++ b/formatTest/unit_tests/expected_output/externals.re
@@ -1,0 +1,7 @@
+/**
+ * Tests external formatting.
+ */
+external foo: type_ = "%caml_something_or_other";
+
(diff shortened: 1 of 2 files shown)


formatTest/unit_tests/expected_output/externals.re
```diff
@@ -0,0 +1,7 @@
+/**
+ * Tests external formatting.
+ */
+external foo: type_ = "%caml_something_or_other";
+
+external multilineStringExtern: int => int =
+  "\n Did you know you can put whatver you want inside\n of an extern? Good luck with the linker though!\n";
```
jordwalke, line 7: Will send subsequent diff to fix this so that it keeps its line breaks.

PR #2599, opened 2020-07-06
Reason Syntax v4: String Template Literals

jordwalke:
This feature implements a planned v4 feature, but in a way that is totally non-breaking and incremental.

**String template literals (multiline-form)**:
 I was able to make the multiline form of string templates completely non-breaking. It would be a minor version bump, not major. This doesn't implement the _final_ v4 single line version of string templates, which may be a breaking change.

```reason
let x = `
  Hello this is a multiline string literal and
  you can inject an expression ${like ++ this()}
`;
```
This does not conflict with polymorphic variants, because polymorphic variants do not have a newline break after the first backtick.

```reason
let poyVariant = `HelloThere;
let multiLineString = `
  hello!
`;
```
String templates automatically extract the indentation, and normalize it so that you don't have to align text to the zero column (as JS requires you do).

Bonus feature:
Non-multiline version of string templates in a non-breaking change.

Okay, this diff _actually_ _does_ implement single line string templates as well. It is not the final v4 form, but it currently works. It turns out that you can use any initial white space after the opening backtick and before the closing backtick - such as a space, and it will be treated the same.

```reason
let singleLine = ` Here is a template ${stringToo()}! `;
```
In the first breaking change (v4), the leading/final space will not be required, and I would not recommend heavily advertising the single-line version in the mean time, but it was easy to support and did not present any syntactic conflict so I included it.


jordwalke:
This feature allows directly embedding let expressions inside of interpolations such as:

```reason
let commentInLetSequence2 = `
  Okay, the following can accept not just an expression
  but also a "let sequence":
  ${
    let x = 200 + 49;
    string_of_int(x);
  }
  Now wasn't that cool?
`;
```

src/reason-parser/reason_template.ml
```diff
@@ -0,0 +1,97 @@
...
+open Migrate_parsetree
+open Ast_408
+
+module Parse = struct
+  (* Normalizes the last line:
+   *
+   *     `
+   *       abc
+   *       123
+   *       xyz<spc><spc>`
```
cironunes, line 19: Is this a typo? It seems like there's one `<spc>` more?

jordwalke: No, the first and last white space character is ignored in this non breaking change to Reason Syntax.
In the first breaking change where we change the syntax of polymorphic variants, then we will be able to pay attention to the first and last white space character (that is not a newline). The upgrade process will automatically fix any newlines at the end.

cironunes: Nice. Thanks for the explanation.

jordwalke:
Force pushed on top of the version angle braces and the feature that records a version inside of the file [this PR](https://github.com/facebook/reason/pull/2605)

gaku-sei:
I'm not sure that's the proper place to ask, but is there any plan to have the template strings working with unicode characters? As per the doc file [here](https://github.com/facebook/reason/pull/2599/files#diff-dc3d3959c851d8966ae38defc7f32be0R132), it seems the template strings are currently converted to "simple" strings, using the double quotes.

We could have a modifier for this maybe, to make it opt-in, and to use the parsing rules, something like:

```reason
let simple = `Created: ${string_of_int(42)}`;
// "Created: " ++ string_of_int(42)}

let unicode = u`Créée: ${string_of_int(42)}`;
// {j|Créée: |j} ++ string_of_int(42)}
```

PR #2601, opened 2020-07-06
Floating doc comments /** */ should become ocaml.text for odoc

jordwalke:
This allows Reason to be used with odoc such that floating doc comments become the main odoc documentation header for a module.

```reason
/**
 * This is a floating doc comment because it has a semicolon. That means it doesn't attach to the binding below it.
 */;

let x = "hello";
```
This is missing a test case, which is a little hard to test since it prints the same as it parses, without guarantee that it actually becomes ocaml.text in the AST.

### DIFF
--- a/src/reason-parser/reason_attributes.ml
+++ b/src/reason-parser/reason_attributes.ml
@@ -33,7 +33,7 @@ let rec partitionAttributes ?(partDoc=false) ?(allowUncurry=true) attrs : attrib
   | ({ attr_name = {txt="ocaml.text"}; _} as doc)::atTl when partDoc = true ->
     let partition = partitionAttributes ~partDoc ~allowUncurry atTl in
     {partition with docAttrs=doc::partition.docAttrs}
-  | ({ attr_name = {txt="ocaml.doc"}; _} as doc)::atTl when partDoc = true ->
+  | ({ attr_name = {txt="ocaml.doc" | "ocaml.text"}; _} as doc)::atTl when partDoc = true ->
     let partition = partitionAttributes ~partDoc ~allowUncurry atTl in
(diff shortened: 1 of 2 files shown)


src/reason-parser/reason_attributes.ml
```diff
@@ -33,7 +33,7 @@ let rec partitionAttributes ?(partDoc=false) ?(allowUncurry=true) attrs : attrib
   | ({ attr_name = {txt="ocaml.text"}; _} as doc)::atTl when partDoc = true ->
     let partition = partitionAttributes ~partDoc ~allowUncurry atTl in
     {partition with docAttrs=doc::partition.docAttrs}
-  | ({ attr_name = {txt="ocaml.doc"}; _} as doc)::atTl when partDoc = true ->
+  | ({ attr_name = {txt="ocaml.doc" | "ocaml.text"}; _} as doc)::atTl when partDoc = true ->
```
yawaramin, line 36: Could be `| ... when partDoc -> ...`

yawaramin, line 36: Hmm, another thought ... will this account for people using the shorthand form `[@text]`? I'm a bit lazy...

jordwalke: No, it wouldn't. But with Reason, they probably would have no reason to ever write that explicitly because the parser has a nice `/** comment */;` syntax, right? It wouldn't be hard to cover it but it would end up converting `[@text "lkdsajf"]` into `/** comment */` syntax.

yawaramin: Ah, you're right, of course this conversion is going in the other direction. 🤦 myself

PR #2602, opened 2020-07-07
Markdown to ocamldoc conversion for documentation comments

jfrolich:


jfrolich:
I put the markdown in the second "slot" of the attribute. And the reason printer should now print that back instead of the ocamldoc.

How does the test suite of `refmt` work? 
And is this change to the AST ok? I mean there could be other tools that expect the attribute to have single item list?

PR #2605, opened 2020-07-17
Reason V4 Feature [Allow multiple versions of Resaon]

jordwalke:
Summary:This allows multiple versions of Reason in a single project by
inferring and recording the version of syntax used into the file in an
attribute. The attribute allows us to switch the parser and lexer on the
fly. This attribute is not the only way we can infer the version, and we
can allow project level configuration, but this is the approach that is
guaranteed to work with any build system or tooling.

Test Plan:

Reviewers:

CC:


EduardoRFS:
That could also be integrated in dune so that you don't need to write the header later, maybe `refmt` can accept an additional argument and or env variable?

jfrolich:
I think this is quite nifty, but how many times do we expect the syntax to have a breaking change. My concerns are:

- The compiler would need to be more complex incorporating all these previous syntaxes (maintenance burden)
- There's already quite a few bugs in the compiler (printing a parsetree can result in invalid syntax), making it more complex won't make it more solid

Wouldn't it be better to come up with a good syntax proposal that might take a bit more time, and then have a proper migration script?

Then it's also possible to simplify the compiler perhaps incorporating some of the code of the Bucklescript syntax that helps to have better errors and faster parsing/printing.

This partly of comes from my experience with `graphql-ppx` where keeping the compiler compatible adds a lot of complexity and maintenance burden.

jordwalke:
> That could also be integrated in dune so that you don't need to write the header later, maybe refmt can accept an additional argument and or env variable?

Yes, an env var is the easiest way to plumb it through all the tooling, and dune should already support environment variables for build. An build env var could also be set in an esy package.json.

jordwalke:
> There's already quite a few bugs in the compiler (printing a parsetree can result in invalid syntax), making it more complex won't make it more solid

@jfrolich Do you have an example in mind? I've been trying to fix all the ones I've come across so if you have a common one in mind I can take a look at it.

jordwalke:
> The compiler would need to be more complex incorporating all these previous syntaxes (maintenance burden)

That's definitely a valid concern, but either way, we will still want to be able to support two different syntaxes so that we can continually upgrade people without breaking the ecosystem. In the worst case if it gets too burdensome, we could literally just freeze the old parser/printer implementation, fork new ones, and switch depending on a prepassed lex.
Most of the changes for Reason v4 are not _hugely_ different, and most of the changes are actually backwards compatible - such as type "generics" syntax `t<x, y>`.

Lupus:
This actually resembles what dune does. I can specify any dune language version which is `<=` the version of dune that I have, and to make sure my project builds I can depend on dune `>= 1.11` and specify dune language version of `1.11`, and everything will work on any subsequent version of dune that I have installed. It's likely more expensive to maintain, but I see no other way for a tool like reason to evolve rapidly without breaking everything. dune evolves rapidly yet maintains very good compatibility.

Lupus:
Docker is doing the same thing btw, see "Overriding default frontends" in https://docs.docker.com/develop/develop-images/build_enhancements/, special comment at the start of dockerfile can change the syntax flavour being used:

```
The new syntax features in Dockerfile are available if you override the default frontend. To override the default frontend, set the first line of the Dockerfile as a comment with a specific frontend image:

# syntax = <frontend image>, e.g. # syntax = docker/dockerfile:1.0-experimental
```

jfrolich:
> @jfrolich Do you have an example in mind? I've been trying to fix all the ones I've come across so if you have a common one in mind I can take a look at it.

With writing the ppx I often take snapshots of AST and use Refmt to make readable snapshots. I often need to edit them to make them work (like adding some braces in some places). I'll try to post some bug reports.

jordwalke:
@jfrolich Yes, please do report those as issues especially if those same ASTs that format into invalid syntax could be constructed manually in a `.re` file.
(Also: The ocaml syntax printing is often broken, but that's not something related to Reason - Reason just uses the stock ocaml printer. You may not have been  referring to this, but thought I'd mention it.).

jfrolich:
> (Also: The ocaml syntax printing is often broken, but that's not something related to Reason - Reason just uses the stock ocaml printer. You may not have been referring to this, but thought I'd mention it.).

Yeah I am not entirely sure if the AST is possible to construct using plain re. It shouldn't be related to OCaml printer as I just use refmt to parse to binary and then to a AST transform and then print it back. Any idea how to best report this, not sure if a binary AST is very informative.

jfrolich:
I see the tests are now huge bash scripts. Is there any appetite to convert them to rely with snapshot testing? I wrote pretty fast concurrent tests using simple `Unix` processes and pipes for `graphql_ppx` and think I can repurpose a lot of that if the PR is welcome.

jordwalke:
@jfrolich Yes! @davesnx (on discord) is looking into reviving the existing PR that used rely for snapshot testing. It can't come soon enough! These bash scripts are driving me nuts.

jfrolich:
> @jfrolich Yes! @davesnx (on discord) is looking into reviving the existing PR that used rely for snapshot testing. It can't come soon enough! These bash scripts are driving me nuts.

Great! I just saw this PR.

jordwalke:
Couple of other thoughts I had:

1. I've found that sometimes people don't want to upgrade a huge codebase at once because it could cause conflicts with what other people are working on (in a large org) so per file opt in is preferred in those cases. (Even if/when we control it via an env var, some people will want the per file control, just as a transition).
2. With a more concise syntax for attributes, it would look a little nicer:
```reason
@reason.version.3.8;
```
3. We can also optionally encode the version in the first floating doc block comment (if present).
```reason
/**
 * My regular doc comment
 * @reason.version.3.8
 */;
```

To encourage people to use doc comments more we could parse the triple star comments as floating doc comments.

```reason
/***
 * This is a floating doc comment
 * @reason.version.3.8
 */
```

jfrolich:
I like the floating doc comment because it entices people to write proper module documentation. I also like the triple star for floating comments! The only thing is that it might be easier to parse (for humans and compilers) if the reason version is always on the first line of the comment.

PR #2613, opened 2020-07-23
fix object type inherit when using modules

EduardoRFS:
The syntax for object inheritance on types should be any core_type valid, as in OCaml, that is important because of valid types like `Module.t(param)`

```reason
type x = {. ...Module.t(param)}
```

OCaml also does that, as it's shown at https://github.com/ocaml/ocaml/blob/trunk/parsing/parser.mly#L3399


jordwalke:
Thank you @EduardoRFS !

PR #2614, opened 2020-07-24
Hash tags

jordwalke:
Reason v4 - Parse Hashtags for polymorphic variants.

Implements parsing for "hashtags" polymorphic variant constructors.
Since Reason Syntax still supports object syntax, we needed to rearrange
some syntactic real estate to make this work.

```reason
let red = #FF000;

let isRed = color => switch(color) {
  | #FF0000 => true
  | _ => false
};

let callAMethod = someObject::methodName(isRed, "testing red");

let templateLiteral = `
  String template literals are still using backticks.
  String template literals are still using backticks.
`;

```


flash-gordon:
I guess polyvars for colors isn't particularly a good fit, `#00FF00` won't work, will it? 😄

formatTest/typeCheckedTests/input/oo_3_dot_8.re
```diff
@@ -0,0 +1,435 @@
+/* Copyright (c) 2015-present, Facebook, Inc. All rights reserved. */
+
+[@reason.version 3.8];
+
+class virtual stack('a) (init) = {
```
yawaramin, line 5: Does it lead to ambiguity to make this `stack<'a>` too?

src/reason-parser/reason_template.ml
```diff
@@ -0,0 +1,97 @@
+(**
```
yawaramin, line 1: Looks like this comment is a duplicate of the one in the interface, I understand why you might want to have it here though, this can probably be not a doc comment

EduardoRFS:
@jordwalke @flash-gordon maybe we can accept variants starting with poly variants? It's going to be really useful on a BS context.

jordwalke:
@flash-gordon
> I guess polyvars for colors isn't particularly a good fit, #00FF00 won't work, will it? 😄

Why do you have to ruin my demo :D j/k

But yes, you are correct. I suppose there's nothing preventing us from actually forcing those identifiers to be valid in the parse tree. This diff doesn't implement that feature and I wasn't planning on implementing it but in theory it's possible. (The problem is making it interoperate with OCaml).

src/reason-parser/reason_template.ml
jordwalke: This diff is actually tiny, but on top of a rebase of two other PRs unmerged. The one you are discussing is actually [here](https://github.com/facebook/reason/pull/2599). Would you mind commenting on the specific PR if you have feedback?

I will likely be aggressively rebasing and force pushing to all of these PRs to keep them clean and comprehensible, where each PR's *top* commit is the real change for that PR with the underlying commits on the PR being other PRs.

jordwalke:
@EduardoRFS  what do you mean exactly?

EduardoRFS:
@jordwalke sorry not even I know what I intended on that comment, but the point is, why not allow poly variants starting with numbers? The AST clearly supports it, and probably the compiler also supports it(I would be surprised otherwise).

It's same kinda of change as we did on `let.alpha`

jordwalke:
@EduardoRFS Yeah that's a good point. We very well could parse those identifiers. The main downside is that then we would be able to define APIs that OCaml cannot call. That is probably something to be minimized - or at the _very_ least, there should be a large benefit to being able to define those kinds of APIs, and I think that if colors could be expressed practically, that would be one such benefit, but I cannot imagine how to make the types comprehensible. We may prefer to go another route for type safe hex colors. There may be enough syntactic real estate for postfix hashtag `00FF00#` but one challenge is that on some platforms it would require a string and on others it would require an integer.

EduardoRFS:
Oh that's a good point, you will not be able to check it on the type system unless you make a variant with all colors, which is a terrible idea and no one should do it.

Et7f3:
In ocaml we can have `123P` and it is parsed correctly but refused at type checking step so it need to be processed by a ppx. Suffix are one letter from h to z (because a-f are used for hexadecimal). We could have then `00FFFFs` printed as `0x00FFFFs` or `("00FFFF" [@color])` in OCaml and prefix t for tuple. Also not related but about suffix OCaml parser forbid 10pct or 10pt because suffix are 2 letters so we have to choose carefully our suffix.

EduardoRFS:
https://github.com/ocaml/ocaml/pull/9429 Possible problem

PR #2615, opened 2020-07-26
fix parsing of "module type of" on functors return

### LINKED ISSUE #2608: module type of on functor return type fails
Currently we cannot do `(module type of {})` on functor returns, this is completly okay and is even the generated code from ocaml -> reason, so it's probably a parser bug. Tried to track it, didn't find the problem

Tested on master

```reason
module X = (()) : (module type of {}) => {};
```

EduardoRFS:
As there is no ambiguity and `(module type of X)` seems to fit on our definition of `simple_module_type` it seems that it should be there.

Also this is probably safe and there is only two usage of `simple_module_type`, at `module_type` itself and at at https://github.com/facebook/reason/blob/master/src/reason-parser/reason_parser.mly#L1574

This PR closes #2608

### DIFF
--- a/formatTest/unit_tests/expected_output/modules.re
+++ b/formatTest/unit_tests/expected_output/modules.re
@@ -639,3 +639,8 @@ module type Event = (module type of {
 });
 
 include (Version2: (module type of Version2));
+
+/* https://github.com/facebook/reason/issues/2608 */
+module Functor =
+       (())
+       : (module type of {}) => {};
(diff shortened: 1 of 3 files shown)


jordwalke:
You should add this one to the type checked tests because it will catch issues in the parsing of it.

EduardoRFS:
@jordwalke I think that's not needed, there was just a problem on the expected output. I added the comment and forgot to update the tests.

PR #2621, opened 2020-08-12
ocaml-migrate-parsetree 2.0

EduardoRFS:
OMP 2.0 was launched and we got some breaking changes, so if you try it now it's not working with esy, also we probably need a release, maybe as 3.6.1

```json
{
  "dependencies": {
    "@opam/dune": "2.6.2",
    "ocaml": "4.10.0",
    "@esy-ocaml/reason": "3.6.0"
  }
}
```

@jordwalke @anmonteiro

### DIFF
--- a/reason.json
+++ b/reason.json
@@ -14,7 +14,7 @@
     "@opam/menhir": " >= 20170418.0.0",
     "@opam/merlin-extend": " >= 0.3",
     "@opam/result": "*",
-    "@opam/ocaml-migrate-parsetree": "*",
+    "@opam/ocaml-migrate-parsetree": "< 2.0.0",
     "@opam/dune": "*"
   },
   "devDependencies": {


anmonteiro:
Why are there breaking changes? https://github.com/ocaml/opam-repository/pull/16999 fixed this upstream.

anmonteiro:
Is it perhaps a problem in the releases to NPM? That's a world I don't know anything about.

EduardoRFS:
@anmonteiro OMP 2.0 is something completely different as you can see here

https://discuss.ocaml.org/t/ppx-omp-2-0-0-and-next-steps/6231/2

So we need to keep the version in control and it's broken on our npm release, so we need a 3.6.1 to address that, but even if we're using a resolution that will break on esy as esy will choose the most recent version

anmonteiro:
I'm aware it's different, I was mentioning that the upper bound was added in OPAM. If it's needed for NPM, I'll defer to Jordan.

smorimoto:
I have the same opinion as @EduardoRFS, and I suspect that we have a problem with this.

EduardoRFS:
Closed in favor of #2622

PR #2622, opened 2020-08-16
Put upper bound of 2.0.0 for Ocaml migrate parsetree.

jordwalke:
This is a more complete version of [this PR](https://github.com/facebook/reason/pull/2621)

Questions:
I think that this won't be enough to fix package installs because we might need
to deprecated the old version of Reason packages, or at least edit the esy opam
overrides repo to make sure earlier versions of Reason don't get solved
assuming it is compatible with omp 2.0.0


smorimoto:
Looks good ☺️

PR #2624, opened 2020-08-19
Fix letop binary printing

anmonteiro:
this was happening after the downgrade so it effectively only worked if
you were printing to the console?


anmonteiro:
I think this is fairly low-risk and can be reverted easily so I'm gonna go ahead and merge it.

PR #2628, opened 2020-09-21
Bumps deps

Et7f3:
This PR allow to compile it with 4.11. It just need to be released on npm.

Relevant diff: https://github.com/facebook/reason/pull/2628/files/4c3336f0ab875d38eebef67dd24f61df3ddf295b

Ok so ubuntu 16.04 can't have glibc 2.27 I have bumped ubuntu image and it worked :white_check_mark:

ci failure: maybe try to used old ubuntu with a esy compiled with a glibc < 2.27.
@jordwalke PR is ready to review and can be released.


Et7f3:
I think it is just debian that is broken: I have tried old dune (< 2.0.0) and get error.

azure-pipelines.yml
```diff
@@ -35,7 +35,7 @@ stages:
     - template: .ci/build-platform.yml
       parameters:
         platform: Linux
-        vmImage: ubuntu-16.04
+        vmImage: ubuntu-18.04
```
anmonteiro, line 38: I'm personally not opposed to bumping this here, but if Esy is gonna keep distributing prebuilt binaries then it should probably make sure they're built on an older Linux (cc: @ManasJayanth)

ManasJayanth: I'm not sure I understand the issue here - Dune 1.x doesn't build on older Ubuntu? I cant understand why.

anmonteiro: The issue seems to have been that Esy didn't run with a libc < 2.27

ManasJayanth: Got it. This could be reported on esy's issue tracker. Thank you.

anmonteiro:
superseded by https://github.com/facebook/reason/pull/2639

PR #2635, opened 2020-11-08
Add support for OCaml 4.12

kit-ty-kate:


kit-ty-kate:
Successfully used to compile `naboris.0.1.2` on OCaml 4.12

anmonteiro (approved):
Thanks, Kate! I went ahead and fixed a few minor things.

This is ready to merge and all the test runs pass in opam. If someone who uses esy would like to fix the esy failures, that'd be great, but those aren't related to this PR.

PR #2647, opened 2021-02-28
Fix Circle-CI

kit-ty-kate:
The ocurrent and opam2 images are not updated anymore. `ocaml/opam` should be used instead.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,7 +66,7 @@ version: 2
 jobs:
   4.03.0:
     docker:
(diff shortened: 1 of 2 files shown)


kit-ty-kate:
i do not have access to the cycleci log, could someone tell me what's wrong?

anmonteiro:
I pushed a squashed commit of this PR to https://github.com/reasonml/reason/pull/2646 's branch. 

@kit-ty-kate thank you, as always!

PR #2650, opened 2021-03-25
Revert "Better handling of longidents in JSX (#2541)"

anmonteiro:
This reverts commit 35896458c960e582965e5f977705d17618578d67.

The PR breaks the following example in the JSX PPX.

```reason
module Internal = {
  [@react.component]
  let header = () => <div />;
};

[@react.component]
let make = () => <div> <Internal.header /> </div>;
```


anmonteiro:
Alright, I rushed to this one, this can be fixed in the React JSX PPX. cc @Drup

PR #2666, opened 2022-03-24
Windows long paths issue: rename Reason_migrate_parsetree to Reason_omp

ManasJayanth:
Dune creates paths as long as `_build/.sandbox/b1f521e428a6b318fe988137841394cd/default/src/vendored-omp/src/.reason_migrate_parsetree.objs/byte/reason_migrate_parsetree__Migrate_parsetree_compiler_functions.cmi` which hit Windows' max_path. Enabling long paths in the registry isn't enough as the underlying win32 API that accept long paths are not called.

This PR renames the vendored ocaml  migrate parsetree's namepsace to make the paths shorter.


anmonteiro:
Thanks

hhugo:
> Enabling long paths in the registry isn't enough as the underlying win32 API that accept long paths are not called.

Has this been reported upstream ?

ManasJayanth:
@hhugo I'm not entirely sure, but I think this is a stdlib (more correctly, Unix module) issue. It relies on C stdlib routines, which I don't think support long paths.

PR #2674, opened 2022-10-29
Fix extensions printing

davesnx:
When printing code with refmt: Extensions are printed with and without brackets `[]` in some cases, creating a few problems: 

- Different expressions are treated differently when formatting (that might give the impression to be different expressions when they aren't)
- Creates unbalance code where used inside branches

This PR tries to unify them to always add `[]` while respecting the inline let attributes added https://github.com/reasonml/reason/pull/1703.


dune
```diff
@@ -1 +1 @@
-(ignored_subdirs (node_modules js))
+(data_only_dirs node_modules js)
```
anmonteiro, line 1: this should be `(dirs :standard \ node_modules js)`.

`data_only_dirs` is a different concept.

formatTest/typeCheckedTests/expected_output/attributes.re
```diff
@@ -237,7 +237,10 @@ class boxB ('a) (init: 'a) =
 
 /* To be able to put an attribute on just the return in that case, use
  * parens. */
-[@onBoxC x; y]
+[@onBoxC
```
anmonteiro, line 240: this one is a little awkward

formatTest/typeCheckedTests/expected_output/attributes.rei
```diff
@@ -78,7 +78,5 @@ type ttttt = {
   x: int,
 };
 
-module Foo: {
-  [@someattr]
-  let foo: int => int;
-};
+module Foo: {[@someattr]
```
anmonteiro, line 81: this needs to be fixed

formatTest/unit_tests/expected_output/extensions.re
```diff
@@ -1,13 +1,12 @@
 /* Extension sugar */
 
-%extend
```
anmonteiro, line 3: if this was valid syntax, shouldn't we preserve its printing?

anmonteiro: I guess this is what the PR is fixing, ignore my comment.

anmonteiro (commented):
looks good, there's just the module case to fix

formatTest/typeCheckedTests/expected_output/attributes.re
davesnx: At first, I found it weird but later made sense. Any structure_item inside an attribute should be treated like a structure_item without an attribute...

I can roll it back thought.

formatTest/typeCheckedTests/expected_output/attributes.rei
davesnx: Fixed in e17fe1b

anmonteiro: very nice, thanks

formatTest/typeCheckedTests/expected_output/attributes.re
anmonteiro: that does make sense. happy to keep it as is

anmonteiro (approved):
LGTM. I think we're ready to land this

PR #2678, opened 2022-12-04
Update reactjs_jsx_ppx_v2.ml

lambda0xE:
The code has been refactored to improve its structure and indentation, making it more readable. Additionally, a match operator with handling of all possible cases has been used instead of multiple nested if conditions. This simplifies reading the code and increases its readability.


davesnx:
This ppx isn't being used anywhere and it's quite old. I will close this PR, but feel free to open this PR into the right place: either [Melange repo](https://github.com/melange-re/melange) or https://github.com/rescript-lang/rescript-compiler

PR #2679, opened 2022-12-06
WIP Prettier Printing of function calls DO NOT MERGE

jordwalke:
This demonstrates how to achieve prettier printing of function calls. Sending a PR just for discussion.
Instead of:

```reason
let x =
  reallyLongFunction(
    arg,
    arg,
  );
```

It will print:

```reason
let x = reallyLongFunction(
  arg,
  arg,
);
```

In its current state it also applies the same convention to other constructs which should _not_ be wrapped in that manner - it's just a matter of going through and telling the printer which style to use in each case.
I included some test output changes which demonstrate both some improvements and some regressions in printing.

### DIFF
--- a/formatTest/typeCheckedTests/expected_output/reasonComments.re
+++ b/formatTest/typeCheckedTests/expected_output/reasonComments.re
@@ -148,46 +148,41 @@ type t2 = (int, int); /* End of line on (int, int) */
 
 type t3 = (int, int); /* End of line on (int, int) */
 
-type variant =
-  | X(int, int) /* End of line on X */
-  | Y(int, int); /* End of line on Y */ /* Comment on entire type def for variant */
+type variant = | X(int, int) /* End of line on X */
+               | Y(int, int); /* End of line on Y */ /* Comment on entire type def for variant */
 
 /* Before let */
-let res =
-  /* Before switch */
-  switch (X(2, 3)) {
-  /* Above X line */
-  | X(_) => "result of X" /* End of arrow and X line */
-  /* Above Y line */
-  | Y(_) => "result of Y" /* End of arrow and Y line */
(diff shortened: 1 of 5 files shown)


formatTest/typeCheckedTests/expected_output/reasonComments.re
```diff
@@ -148,46 +148,41 @@ type t2 = (int, int); /* End of line on (int, int) */
 
 type t3 = (int, int); /* End of line on (int, int) */
 
-type variant =
-  | X(int, int) /* End of line on X */
-  | Y(int, int); /* End of line on Y */ /* Comment on entire type def for variant */
+type variant = | X(int, int) /* End of line on X */
```
jordwalke, line 151: This is a regression obviously.

formatTest/typeCheckedTests/expected_output/reasonComments.re
```diff
@@ -209,86 +204,80 @@ and y2 = {
...
-    2 + tmp;
-  | Some({
-      fieldOne: n, /* end of line */
-      fieldA: a /* end of line */
-    }) =>
-    let tmp = n;
-    n + tmp;
-  | None => 20
-  };
+let result = switch (None) {
```
jordwalke, line 230: This is the right idea.

formatTest/typeCheckedTests/expected_output/reasonComments.re
```diff
@@ -209,86 +204,80 @@ and y2 = {
...
+
+type optionalTuple = | OptTup(
+                         option(
+                           (
+                             int, /* First int */
+                             int /* Second int */
+                           ),
+                         ),
+                       );
+
```
jordwalke, line 274: Right idea below

Issue #2682, opened 2023-01-31
Empty Functor adds unit on declaration and empty module on instantiation


davesnx:
```reason
module Lola = () => {
  let a = 33;
};

module L = Lola();
```

Formats
```reason
module Lola = (()) => {
  let a = 33;
};

module L =
  Lola({});
```

anmonteiro:
this was fixed by #2683

PR #2683, opened 2023-02-23
Improve functor printing.

SanderSpies:
Reduces the amount of parentheses around functor usage.

Related to: #2682.

### DIFF
--- a/src/reason-parser/reason_pprint_ast.ml
+++ b/src/reason-parser/reason_pprint_ast.ml
@@ -7542,7 +7542,7 @@ let printer = object(self:'self)
         formatPrecedence (self#module_type letPattern mt)
     | Pmod_structure s ->
         let wrap = if hug then
-          if List.length s = 0 then 
+          if s = [] then 
             ("(", ")") 
           else 
             ("({", "})")


anmonteiro:
Thanks, looks good. Can you add an entry to `HISTORY.md`?

SanderSpies:
@anmonteiro I added a history entry. I hope 3.9 makes sense as a next version.

SanderSpies:
OCaml 4.03 CI  is having problems with installing nodejs and npm.

src/reason-parser/reason_pprint_ast.ml
```diff
@@ -7550,15 +7555,15 @@ let printer = object(self:'self)
             s
         in
         makeList
-          ~break:Layout.Always_rec
+          ~break:(if List.length s = 0 then Layout.IfNeed else Layout.Always_rec)
```
davesnx, line 7558: I have seen this logic in more than a few places, maybe we could try to do this check inside makeList and see the snapshots?

SanderSpies: Apologies for the late reply. I've added this to the changes of this PR.

src/reason-parser/reason_pprint_ast.ml
```diff
@@ -7541,7 +7541,12 @@ let printer = object(self:'self)
         in
         formatPrecedence (self#module_type letPattern mt)
     | Pmod_structure s ->
-        let wrap = if hug then ("({", "})") else ("{", "}") in
+        let wrap = if hug then
+          if List.length s = 0 then
```
anmonteiro, line 7545: `List.length` is O(n)

```suggestion
          if s = [] then 
```

PR #2686, opened 2023-03-01
Improve printing of anonymous function return type.

SanderSpies:
fixes https://github.com/reasonml/reason/issues/2589.

### DIFF
--- a/src/reason-parser/reason_pprint_ast.ml
+++ b/src/reason-parser/reason_pprint_ast.ml
@@ -44,6 +44,8 @@
  *
  *)
 
+let () = Printexc.record_backtrace true
+
 (* TODO more fine-grained precedence pretty-printing *)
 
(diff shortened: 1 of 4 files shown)


formatTest/unit_tests/input/pexpFun.re
```diff
@@ -43,21 +43,15 @@ Mod.Update(
 );
 
 Mod.Update(
-  [@foo] [@bar]
```
SanderSpies, line 46: TODO: change this back. Probably an accidental "format on save".

src/reason-parser/reason_pprint_ast.ml
```diff
@@ -44,6 +44,8 @@
  *
  *)
 
+let () = Printexc.record_backtrace true
```
SanderSpies, line 47: TODO: remove.

PR #2687, opened 2023-03-02
[refmt] Add env variable to configure --print-width

davesnx:
The workaround is to have an env variable to set the wrapping width for printing. It should unblock https://github.com/reasonml/reason/issues/2567 and delay the discussion about the need for a config file.

This unblocks editor integrations to configure it however they want it.

```
$ echo "let initialState = uiStateFromValidated(~ownership=RemoteData.NotAsked, ~limits=initialLimits, SiteAuditSettings.default);" | REFMT_PRINT_WIDTH=80 esy x refmt
let initialState =
  uiStateFromValidated(
    ~ownership=RemoteData.NotAsked,
    ~limits=initialLimits,
    SiteAuditSettings.default,
  );

$ echo "let initialState = uiStateFromValidated(~ownership=RemoteData.NotAsked, ~limits=initialLimits, SiteAuditSettings.default);" | REFMT_PRINT_WIDTH=120 esy x refmt
let initialState =
  uiStateFromValidated(~ownership=RemoteData.NotAsked, ~limits=initialLimits, SiteAuditSettings.default);
```

---

This PR also adds cram tests, which tests refmt end to end, and are a little nicer than our test.sh: 

- cram tests are fast
- cram makes it easy to generate snapshots (something we had to manually do) and diff snapshots for us with meaningful error messages
- Much easier and less error prone to add new tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,7 @@ test-once-installed: clean-tests
 	./miscTests/rtopIntegrationTest.sh
 	./miscTests/backportSyntaxTests.sh
 	cd formatTest; ./test.sh
+	dune runtest
 
 .PHONY: coverage
 coverage:
--- a/dune-project
+++ b/dune-project
@@ -1,5 +1,5 @@
-(lang dune 2.3)
-
+(lang dune 2.7)
 (name reason)
 
 (using menhir 2.0)
+(cram enable)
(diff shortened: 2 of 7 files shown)


SanderSpies:
LGTM.

PR #2688, opened 2023-03-02
Add cram tests

davesnx:


davesnx:
Closing since it will be added into

PR #2690, opened 2023-03-03
Fix ppat_constraint case

SanderSpies:
Fixes issue [2636](https://github.com/reasonml/reason/issues/2636) where 

```
let branch_with_variant_and_annotation =
  fun
  | (VariantType1: test_foo) => true
  | VariantType2 => false;
```

is printed as

```
let branch_with_variant_and_annotation =
  fun
  | VariantType1: test_foo => true
  | VariantType2 => false;
```

and fails to compile.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,7 @@
+## 3.9
+
+- Fix missing patterns around contraint pattern (a pattern with a type annotation).
+
 ## 3.8.2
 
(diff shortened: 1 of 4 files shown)


jordwalke:
Thank you. Curious how this was introduced. I think the test failure is unrelated?

anmonteiro:
Yup, CI failure is unrelated. Fix looks good to me. @davesnx please add these tests to your cram PR.

jordwalke:
Thank you

PR #2693, opened 2023-03-06
Fix top level extensions

davesnx:
While migrating to Reason 3.8.2 (and above) found a bug when printing extensions. It might be hidden a long time ago since most extensions aren't top-level.


refmt_test/extensions.t/run.t
```diff
@@ -12,7 +12,7 @@ Format extensions
   
   type a = [%extend int];
   
-  let%extend x = "hi";
+  [%%extend let x = "hi"];
```
davesnx, line 15: This is expected since it's a top_level extension with a let binding. I can fix this particular case or leave as it is now not entirely sure if it's worth it.

src/reason-parser/reason_pprint_ast.ml
```diff
@@ -7576,29 +7576,40 @@ let printer = object(self:'self)
...
     | [] -> atom ""
-    | first::_ as structureItems ->
+    | first :: _ as structureItems ->
       let last = match (List.rev structureItems) with | last::_ -> last | [] -> assert false in
       let loc_start = first.pstr_loc.loc_start in
       let loc_end = last.pstr_loc.loc_end in
       let items =
         groupAndPrint
-          ~xf:self#structure_item
+          ~xf:self#top_level_structure_item
```
SanderSpies, line 7600: This looks fishy.

davesnx: Probably a bad name, but the logic is to check for a specific combination of AST such as: `structure -> structure_item -> Pexp_extension`

formatTest/unit_tests/input/extensions.re
```diff
@@ -348,7 +348,7 @@ let () = {
   something_else();
 };
 
-[%bs.raw x => x];
```
anmonteiro, line 351: don't we also accept this form? I think it should keep being tested.

src/reason-parser/reason_pprint_ast.ml
```diff
@@ -7576,29 +7576,42 @@ let printer = object(self:'self)
     | Pmod_constraint _
     | Pmod_structure _ -> self#simple_module_expr x
 
-
   method structure structureItems =
+    (* We don't have any way to know if an extension is placed at the top level by the parsetree
```
anmonteiro, line 7580: I thought we supported extensions at the toplevel with `%`, and that `%%bs.raw` was specifically a bucklescript thing.

formatTest/unit_tests/input/extensions.re
davesnx: Pushed back the test case, we do support them.

anmonteiro: It's still removed. did you forget to push?

davesnx: I added all of the cases in the cram tests. Let me add them in the `formatTest` as well

src/reason-parser/reason_pprint_ast.ml
davesnx: It is a BuckleScript thing, but we support both cases. Would be a nice idea to unify all of these cases with Melange. `%bs.raw` -> `bs.raw_js_expr` and `%%bs.raw` -> `bs.raw_js` or something like this.

anmonteiro: I don't get what needs to be unified?

anmonteiro (approved):
This needs a changelog entry in HISTORY.md. Then please "squash and merge" it.

SanderSpies (approved):

src/reason-parser/reason_pprint_ast.ml
davesnx: We could avoid having `%` and `%%` at the same time and support only `%`

PR #2698, opened 2023-03-13
Add Github Actions to migrate CI from CircleCI and Azure DevOps

davesnx:
## Introduction

This PR adds GitHub workflow files that have been tested in a separate repository https://github.com/davesnx/reason-ci-gh-actions/actions to ensure correctness.

What it consist of the CI:

- opam-ci
- esy-ci

Both running against 3 different OS (windows, ubuntu and mac) and running with 4 different OCaml versions (4.06, 4.10, 4.12, 4.14).

The list of OCaml versions has been changed just because we might want to stop supporting < 4.06 versions for Reason. There is no technical reason to not run 4.02 in GH actions.

## Why

- Running one CI is preferred to running 2 separate ones.
- GH actions support windows (that wasn't true 5 years ago, so that why Azure was used)
- GH cache is superior from CircleCI cache and Azure cache.
- In ocaml we can use ocaml/setup-ocaml and run it in Windows
- The config files are easier to understand IMHO
- The performance is superior

## Benchmark 
The slowest builds right are in Azure: esy+windows: 27minutes opam+windows: 45min, the fastest are esy+mac: 9m 27sec.
The builds on CircleCI are with esy: 5m 39s (with cache)

Stats for Github Actions
  - **esy+ubuntu**: between `1m 28s` and `1m 52s`
  - **esy+windows**: between `3m 50s` and `7m`
  - **opam+ubuntu**: between `4m 29s` and `6m 57s`
  - **opam+windows**: between `5m 19s` and `8m 58s`

It's much faster since it takes advantage of action/cache from GH. Stores _export and ~/.esy folders in case of esy and ~/.opam in case of opam.
In esy+windows most of the time spent is installing esy each time where isn't possible to install globally a dependency with the cache enabled, the rest is as fast as the rest of esy pipelines. (If we want to remove this minutes, I can probably create an action that uses-esy).

Inspiration was taken from https://github.com/davesnx/styled-ppx and https://github.com/grain-lang/grain

## Doubts

The release step from Azure is somewhat mysterious for me, and the overall process of releasing a new Reason version seems a little broken. I would appreciate a little bit of guidance here since would be nice to move it to GH actions as well.

- What should we do with `js`? as far as I know playground and sketch.sh (the only use-cases I can think of) they build reason with jsoo and live without this. Maybe this should be replaced with a section on the documentation about "Run refmt in the browser", instead of releasing an npm package that contains refmt compiled with jsoo. It complicates the repo with a minimum benefit
- What should we do with `bspacks` folder?
- npm workflow -> we create this on each CI run but we don't push it to npmjs registry?
- Why do we have a package.json with bs-platform 1.9.3 when on publish we re-generate our own package.json? Is that a mistake or an artifact from BuckleScript?
- opam workflow -> `scripts/opam-release.sh` runs manually and then open the PR to opam registry 
- Github Releases -> Running check-version.sh? Manually? 

## Next steps

We will run all CIs for a few weeks, and when everything seems to work fine we will remove the CircleCI, Azure DevOps, and `formatTest/` when all tests are migrated.

Theoretically, the release process would be more clear and we won't have problems with `refmt --version`, GH releases, npm packages, opam packages being completely different.


anmonteiro (approved):
This doesn't look to be testing the error / type check tests. So we can't disable Circle / Azure yet.

In the meantime, I think this can go in

davesnx:
Merging this, future work is to migrate all tests to cram test, disable Circle/Azure and polish the release process.

PR #2701, opened 2023-03-21
Fix version on refmt

davesnx:
This PR is a step close to solve the issue with the release process being broken.
Fixing the version on running `refmt --version`

This would always print the version from `dune-project` which would need to be in sync with `reason.json`, `package.json`, `esy.json` which is not ideal.

- Using build_info from dune to display version (defined in dune-project)
- Using dune rule to get the commit hash and write it to `git_commit.ml`
- Remove pre-release on Makefile
- Set ocaml to 4.14 and ocaml-lsp to 1.15 for development

In the future the publish is going to be automated by CI, only updating the version on the `dune-project` file will propagate to all releases.


esy.json
```diff
@@ -3,9 +3,10 @@
   "notes":
     "This is just the dev package config (also built as globally installable reason-cli). See ./refmt.json ./rtop.json for individual release package configs.",
   "license": "MIT",
-  "version": "3.8.1",
+  "version": "3.8.2",
   "dependencies": {
     "@opam/dune": "3.6.2",
```
anmonteiro, line 8: we should probably be on the same dune version as dune-build-info

reason.json
```diff
@@ -15,7 +15,8 @@
     "@opam/merlin-extend": " >= 0.6",
     "@opam/result": "*",
     "@opam/ppx_derivers": "< 2.0.0",
-    "@opam/dune": "2.9.3"
+    "@opam/dune": "2.9.3",
```
anmonteiro, line 18: let's try to use the same dune version across all `.json` files.

davesnx: Updated in 22e86a686dbd893821f20fea8f04d3d69994e464 + update reason.json/rtop.json

PR #2704, opened 2023-03-24
generate opam files with dune

anmonteiro:


davesnx:
Rebased with master and fixed some conflicts. I touched the default.nix and didn't test it manually, can you double check it works/make sense?

anmonteiro:
Thanks! looks right to me.

PR #2707, opened 2023-04-17
Polyvariants printing similar to variants 

SanderSpies:
```
type align = [ | `Start | `End | `Center];
```
to 
```
type align = [ 
  | `Start
  | `End 
  | `Center
];
```

Solves: https://github.com/reasonml/reason/issues/2706


SanderSpies:
Hmm, looks like something went wrong with git.

Issue #2695, opened 2023-03-08
local open + let bindings


anmonteiro:
```reason
let y = Promise.Ops.(
  let* x = Js.Promise.resolve(42);
  Js.Promise.resolve(x * 2);
);
```

this currently produces a syntax error "unclosed parenthesis"

anmonteiro:
Done in #2716

PR #2716, opened 2023-04-24
Fix 2695: local open + let bindings

SanderSpies:
Solves #2695.

Makes it possible to use:

```
 let y = Promise.Ops.(
  let* x = Js.Promise.resolve(42);
  Js.Promise.resolve(x * 2);
);
```

Printing isn't perfect, but good enough for now.

### DIFF
--- a/src/reason-parser/reason_pprint_ast.ml
+++ b/src/reason-parser/reason_pprint_ast.ml
@@ -4365,21 +4365,16 @@ let printer = object(self:'self)
      * Or ReasonReact.(<> {string("Test")} </>);
      *)
     | _ -> (
-      match parent with 
-      | Some (parent) -> (
-        if has_open_notation_attr parent.pexp_attributes then
-          makeList
-            ~break:IfNeed
-            ~inline:(true, false)
-            ~postSpace:true
-            ~wrap:("(",")")
-            ~sep:(SepFinal (";", ""))
-            (self#letList e)
-        else
-          makeList ~wrap:("(",")") ~break:IfNeed [self#unparseExpr e]
-      )
-      | None -> 
-        makeList ~wrap:("(",")") ~break:IfNeed [self#unparseExpr e]
+      match parent with
+      | Some parent when has_open_notation_attr parent.pexp_attributes ->
+        makeList
+          ~break:IfNeed
+          ~inline:(true, false)
+          ~postSpace:true
+          ~wrap:("(",")")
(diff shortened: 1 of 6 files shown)


anmonteiro (approved):
This looks great, thanks!

src/reason-parser/reason_pprint_ast.ml
```diff
@@ -6442,7 +6442,12 @@ let printer = object(self:'self)
                     (atom (".")))
                   (self#formatNonSequencyExpression e))
           else
-            Some (makeLetSequence (self#letList x))
+            Some
```
anmonteiro, line 6445: this looks exactly like the `then` branch.

I'm surprised this doesn't break `let open Foo; bar;` for example?

Perhaps the `letop` case should be added to `isSeriesOfOpensFollowedByNonSequencyExpression`, then we'd get printing for free?

SanderSpies: Well, not exactly the same. The first branch results into extra braces: 
```
-      let* x = Js.Promise.resolve(42);
-      Js.Promise.resolve(x * 2);
+      {
+        let* x = Js.Promise.resolve(42);
+        Js.Promise.resolve(x * 2);
+      }
```

SanderSpies: I did make a small change to avoid code duplication

anmonteiro: I don't see how this is correct yet. I might be missing something.

Aren't we replacing ALL `open` calls with `Foo.( ...  )`?

SanderSpies: @anmonteiro 

Not exactly, but the solution wasn't perfect (it didn't work for nested local open modules). I've tried to improve on it with help of an extra attribute set by the parser.

src/reason-parser/reason_pprint_ast.ml
```diff
@@ -6433,15 +6458,17 @@ let printer = object(self:'self)
                 | _ -> Some (self#extension e)
           end
         | Pexp_open (me, e) ->
-            if self#isSeriesOfOpensFollowedByNonSequencyExpression x then
+            if self#isSeriesOfOpensFollowedByNonSequencyExpression x then (
```
SanderSpies, line 6461: Todo: make the same as before.

anmonteiro:
Thank you again, and sorry for the delay

PR #2717, opened 2023-04-25
Add formatting diff on typecheck

davesnx:
- Rename typedtree to outcometree (to be correct cc @anmonteiro)
- Added the printed output for the typeChecked tests
- Removed unused scripts/failing commands from Make


anmonteiro:
Thanks!

PR #2721, opened 2023-07-22
recognize `mel.obj` instead of `bs.obj`

anmonteiro:
We're moving from `%bs.obj` to `%mel.obj` in https://github.com/melange-re/melange/pull/663

- This change makes that possible in Reason as well
- BuckleScript is no longer; ReScript is not using Reason
- This change should be safe to ship (and release as part of the melange v2 releases)

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -5,11 +5,11 @@
         "systems": "systems"
       },
       "locked": {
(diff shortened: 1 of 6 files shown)


davesnx:
This should be released under 3.9.1?

PR #2723, opened 2023-09-19
test: reproduce extension bug when nested inside a module

anmonteiro:


anmonteiro:
Pushed a fix for this but I'm not happy with the extra `~indent` argument.

cc @jchavarri

jchavarri (approved):
Nice fix.

PR #2725, opened 2023-10-05
fix: record constraint pattern

anmonteiro:
fixes https://github.com/reasonml/reason/issues/2724


anmonteiro:
Esy CI is mad at something

PR #2731, opened 2023-11-09
Rtop for ocaml5

ManasJayanth:
`rtop` 's compiler constraint can be relaxed as it's compatible with `5.1.0` This makes it available on for such switches


anmonteiro:
Thanks

PR #2739, opened 2024-06-23
fix: macOS CI

anmonteiro:
it's trying to run our tests on arm64, which OCaml < 4.12 doesn't support


anmonteiro:
not sure what's up with esy, but this fixes the macOS CI for opam.

PR #2814, opened 2024-11-25
fix: better printing of `Pexp_extension`

anmonteiro:
fixes https://github.com/reasonml/reason/issues/2677

@davesnx what do you think about this fix for the issue you opened?


davesnx:
That looks correct to me! @anmonteiro

PR #2824, opened 2025-01-24
Fix record field and jsx props punned when with attributes 

pedrobslisboa:
## Description

This PR adds validation for records fields and jsx props to avoid pun them when the have attributes

### DIFF
--- a/src/reason-parser/reason_pprint_ast.ml
+++ b/src/reason-parser/reason_pprint_ast.ml
@@ -2113,8 +2113,10 @@ let createFormatter () =
         true
       | _ -> false
 
-    let isPunnedJsxArg lbl ident =
-      (not (isLongIdentWithDot ident.txt)) && Longident.last_exn ident.txt = lbl
+    let isPunnedJsxArg lbl ident attr =
+      (not (isLongIdentWithDot ident.txt))
+      && Longident.last_exn ident.txt = lbl
+      && attr = []
 
     let is_unit_pattern x =
       match x.ppat_desc with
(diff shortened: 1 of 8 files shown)


src/reason-parser/reason_pprint_ast.ml
```diff
@@ -5260,7 +5265,9 @@ let createFormatter () =
               let value_has_jsx = jsxAttrs != [] in
               let nextAttr =
                 match expression.pexp_desc with
-                | Pexp_ident ident when isPunnedJsxArg lbl ident -> atom lbl
+                | Pexp_ident ident
+                  when isPunnedJsxArg lbl ident expression.pexp_attributes ->
```
anmonteiro, line 5269: same as above

src/reason-parser/reason_pprint_ast.ml
```diff
@@ -7099,9 +7099,11 @@ let createFormatter () =
               match e.pexp_desc, shouldPun, allowPunning with
               (* record value punning. Turns {foo: foo, bar: 1} into {foo, bar: 1} *)
               (* also turns {Foo.bar: bar, baz: 1} into {Foo.bar, baz: 1} *)
+              (* don't turn {bar: [@foo] bar, baz: 1} into {bar, baz: 1} *)
               (* don't turn {bar: Foo.bar, baz: 1} into {bar, baz: 1}, naturally *)
               | Pexp_ident { txt = Lident value; _ }, true, true
-                when Longident.last_exn li.txt = value ->
+                when Longident.last_exn li.txt = value && e.pexp_attributes = []
```
anmonteiro, line 7105: I think we may need to partition attributes here too

anmonteiro (commented):
thanks for solving this one! almost there

pedrobslisboa:
> thanks for solving this one! almost there

Thanks for the review @anmonteiro

PR #2835, opened 2025-02-16
feat: support ppxlib with OCaml 5.2 AST

### LINKED ISSUE #2839: fix pvb_constraint with coercion after 5.2 AST upgrade
this

```re
let x :> [> `A | `B ] = `A;
```

reformats to

```re
let x: :> [>
            | `A
            | `B
          ] = `A;
```

there's an extra `:` in there, we should fix it

anmonteiro:
fixes #2839


test/wrapping-re.t/run.t
```diff
@@ -2174,40 +2176,46 @@ Format wrapping in .re files
...
+      : (
+          (a, b) =>
+          (
+            inputEchoRecord(a),
+            inputEchoRecord(b),
+          )
+        ) =>
+    (input: a, input2: b) => (
+      {inputIs: input},
+      {inputIs: input2},
```
davesnx, line 2209: This is a weird printing

test/variants.t/run.t
```diff
@@ -322,15 +322,14 @@ Format variants
...
-      | App(f, x) => eval(f, eval(x))
-      };
+  let rec eval: type a. term(a) => a = x =>
+    switch (x) {
+    | Int(n) => n
+    /* a = int */
+    | Add => ((x, y) => x + y)
+    /* a = int => int => int */
+    | App(f, x) => eval(f, eval(x))
+    };
```
davesnx, line 332: looks like an improvement, nice!

davesnx (approved):
Hard to review the pprint_ast. Looks good, minor changes

test/wrapping-re.t/run.t
anmonteiro: I think it's a pretty big edge case-- no one really writes code like this.

I'm tempted to be reactive here since everything else looks pretty good.

Please feel free to report weird printing when you upgrade internally and I'd be happy to fix those cases.

PR #2941, opened 2026-08-30
fix(printer): keep package constraints on the lambda body (#2925)

### LINKED ISSUE #2925: `refmt` rewrites valid code into code that doesn't compile
Signature constraint on a first-class-module lambda body is hoisted into a return annotation that re-parses as an arrow type.

Narrowing a first-class module to a smaller signature inside `Option.map` — this program compiles:

```reason
// narrow.re
module type FOO = {
  let foo: string;
};

module type FOO_WITH_X = {
  let foo: string;
  let x: int;
};

let some_foo_with_x: option(module FOO_WITH_X) =
  Some(
    (module
     {
       let foo = "foo";
       let x = 1;
     }),
  );

let some_foo =
  some_foo_with_x
  |> Option.map((module Foo_with_x: FOO_WITH_X) =>
       (module Foo_with_x: FOO)
     );
```

Run `refmt narrow.re` once. The body's signature constraint is hoisted into a return-annotation position on the lambda:

```reason
let some_foo =
  some_foo_with_x
  |> Option.map((module Foo_with_x: FOO_WITH_X): (module FOO) =>
       (module Foo_with_x)
     );
```

That output no longer compiles, because it re-parses with the annotation as an **arrow type**, swallowing the lambda body into a type expression. `refmt --parse re --print ml` on the formatted file:

```ocaml
let some_foo =
  some_foo_with_x |>
    (Option.map (((module
       Foo_with_x) : (module FOO_WITH_X)) : (module FOO) ->
                                              (module Foo_with_x)))
```

```console
$ ocamlc -c narrow_fmt.ml
Error: Unbound module type Foo_with_x
```

So a plain `refmt` / `dune fmt` pass silently converts a compiling program into a non-compiling one. It is also not idempotent — a second pass rewrites the lambda again:

```reason
  |> Option.map(
       ((module Foo_with_x): (module FOO_WITH_X)):
                                                   (module FOO) =>
                                                   (module Foo_with_x),
     );
```

The same wrong reading applies if you write the return annotation by hand: `(module M: SIG): (module SUB) => body` parses as a pattern annotated with an arrow type, rather than as a lambda with a return annotation — unlike the same shape with ordinary (non-module) patterns.

## Versions

Reproduced identically on Reason **3.14.0** and **3.17.3** (macOS/arm64, ocamlc 5.1.1) and **3.18.0** (Linux/x86_64, ocamlc 4.14.2). All snippets above verified verbatim.

## Expected

Printing must preserve semantics: refmt shouldn't hoist the body's package constraint into a position that re-parses as an arrow type. (And ideally, `(pattern): T => body` with a `(module …)` pattern would parse as a lambda with return annotation, consistent with ordinary patterns.)

## Workaround

Unpack in the body instead — typechecks, and is stable under refmt:

```reason
|> Option.map((foo_with_x: (module FOO_WITH_X)) => {
     module Foo_with_x = (val foo_with_x: FOO_WITH_X);
     ((module Foo_with_x): (module FOO));
   })
```

MavenRain:
Fixes #2925.

## Problem

refmt hoists a package constraint on a lambda body into a return annotation. For the program in #2925:

```reason
|> Option.map((module Foo_with_x: FOO_WITH_X) =>
     (module Foo_with_x: FOO)
   );
```

one `refmt` pass produces:

```reason
|> Option.map((module Foo_with_x: FOO_WITH_X): (module FOO) =>
     (module Foo_with_x)
   );
```

That output does not re-parse as a function. After the colon, `(module FOO)` starts an arrow type, so the whole argument re-parses as a constraint expression with the type `(module FOO) => (module Foo_with_x)`. The formatted file no longer compiles, and a second format pass rewrites the line again.

No parenthesization of the return annotation avoids this: `((module FOO))` re-parses the same way. The lexer inserts a speculative `ES6_FUN` token in front of `(module FOO)` because `=>` follows it, and the constraint-expression interpretation wins in the multi-parser whenever the lambda body is itself a valid type, which a package expression always is. So the return-annotation form is not a faithful printing of this AST in expression position, and the printer must not produce it there.

## Fix

Keep a `Ptyp_package` constraint on the body in the two printing paths that serve expression positions:

- `formatPexpFun` (lambdas in argument position and JSX expression containers)
- the trailing-callback path in `formatFunAppl`

The output for the issue example becomes:

```reason
|> Option.map((module Foo_with_x: FOO_WITH_X) =>
     ((module Foo_with_x): (module FOO))
   );
```

which re-parses to the same AST and is stable under repeated formatting.

Let bindings keep the return-annotation form (`let f = (module X: S): (module FOO) => ...`): it re-parses correctly there, so their formatting does not change. Ordinary (non-package) return types are not affected anywhere.

## Testing

New cram test `test/firstClassModuleLambda.t` covers the issue example, a non-trailing callback argument, an ordinary return type, and a let binding. It checks the formatted output, checks that it re-parses with the same meaning via `--parse re --print ml`, and checks idempotency. The `--print ml` output of the formatted issue example compiles with ocamlc. All 77 non-rtop cram tests pass.

### DIFF
--- a/src/reason-parser/reason_pprint_ast.ml
+++ b/src/reason-parser/reason_pprint_ast.ml
@@ -5452,6 +5452,11 @@ let createFormatter () =
              here *)
           let return, optConstr =
             match ret.pexp_desc with
+            (* Keep a package constraint on the body: as a return annotation,
+               `(module Foo)` re-parses as the start of an arrow type, which
+               changes the meaning of the printed code (see #2925). *)
+            | Pexp_constraint (_, { ptyp_desc = Ptyp_package _; _ }) ->
+              ret, None
             | Pexp_constraint (e, ct) -> e, Some (self#non_arrowed_core_type ct)
             | _ -> ret, None
           in
@@ -9753,6 +9758,11 @@ let createFormatter () =
             in
             let retCb, cbArgs =
               match retCb.pexp_desc with
+              (* Keep a package constraint on the body: as a return annotation,
+                 `(module Foo)` re-parses as the start of an arrow type, which
+                 changes the meaning of the printed code (see #2925). *)
+              | Pexp_constraint (_, { ptyp_desc = Ptyp_package _; _ }) ->
+                retCb, cbArgs
               | Pexp_constraint (a, t) ->
                 a, makeList [ cbArgs; atom ": "; self#core_type t ]
               | _ -> retCb, cbArgs
--- a/src/reason-parser/reason_pprint_ast.ml
+++ b/src/reason-parser/reason_pprint_ast.ml
@@ -737,6 +737,12 @@ let is_simple_construct : construct -> bool = function
   | `nil | `tuple | `list _ | `simple _ | `btrue | `bfalse | `cons _ -> true
   | `normal -> false
 
+(* A package constraint must stay on the lambda body: as a return annotation,
+   `(module Foo)` re-parses as the start of an arrow type, which changes the
+   meaning of the printed code (see #2925). *)
+let is_package_typ ct =
+  match ct.ptyp_desc with Ptyp_package _ -> true | _ -> false
+
 let uncurriedTable = Hashtbl.create 42
 
 (* Determines if a list of expressions contains a single unit construct * e.g.
@@ -5452,12 +5458,8 @@ let createFormatter () =
              here *)
           let return, optConstr =
             match ret.pexp_desc with
-            (* Keep a package constraint on the body: as a return annotation,
-               `(module Foo)` re-parses as the start of an arrow type, which
-               changes the meaning of the printed code (see #2925). *)
-            | Pexp_constraint (_, { ptyp_desc = Ptyp_package _; _ }) ->
-              ret, None
-            | Pexp_constraint (e, ct) -> e, Some (self#non_arrowed_core_type ct)
+            | Pexp_constraint (e, ct) when not (is_package_typ ct) ->
+              e, Some (self#non_arrowed_core_type ct)
             | _ -> ret, None
           in
           let returnExpr, leftWrap =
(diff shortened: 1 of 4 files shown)


src/reason-parser/reason_pprint_ast.ml
```diff
@@ -9753,6 +9758,11 @@ let createFormatter () =
             in
             let retCb, cbArgs =
               match retCb.pexp_desc with
+              (* Keep a package constraint on the body: as a return annotation,
+                 `(module Foo)` re-parses as the start of an arrow type, which
+                 changes the meaning of the printed code (see #2925). *)
+              | Pexp_constraint (_, { ptyp_desc = Ptyp_package _; _ }) ->
+                retCb, cbArgs
               | Pexp_constraint (a, t) ->
```
davesnx: instead of adding the _ { ptyp_desc = Ptyp_package _; _ } branch at the top, why don't you ensure the 9766 condition is met on the opposite of Ptyp_package? This way the code remains simpler, and you have a chance to test all other cases

MavenRain: Done.  Both sites now keep the single `Pexp_constraint` arm and guard it with `when not (is_package_typ ...)`.  The package case falls through to the existing `_` arm.  The predicate lives next to the other small helpers at the top of the file.  The full cram suite passes, and a plain constraint such as `(a) => (a: int)` still prints as `(a): int => a`, so the general branch stays covered.

src/reason-parser/reason_pprint_ast.ml
```diff
@@ -737,6 +737,12 @@ let is_simple_construct : construct -> bool = function
   | `nil | `tuple | `list _ | `simple _ | `btrue | `bfalse | `cons _ -> true
   | `normal -> false
 
+(* A package constraint must stay on the lambda body: as a return annotation,
+   `(module Foo)` re-parses as the start of an arrow type, which changes the
+   meaning of the printed code (see #2925). *)
```
davesnx: let's remove this comment as it is a bit unnecesary

